PR #54, opened 2017-02-08
Fix for Spark < 1.6

invadergir:
This one line PR fixes a crash when doing spatial partitioning in Spark versions < 1.6.

The fix is to use JavaRDD.rdd().partitions().length rather than JavaRDD.getNumPartitions() for Spark <1.6 compatibility.  This safely gets the number of partitions for this RDD for all versions of Spark 1.X.  (1.0.0 through 1.6.X).  

Note JavaRDDLike has a partitions() method but it is not available in Spark 1.0.0 so I used this approach.

Request: when releasing a new version for 1.X could you please tag the branch with a tag?  It seems only the mainline releases get tags on master.  Thanks...


jiayuasu:
Hi @invadergir ,

Thanks for finding this! The 1.X test was tested on Spark 1.6 so that I didn't find the bug.

This PR is accepted now.

PR #112, opened 2017-09-06
Make SpatialRDD a generic class parameterized over geometry type

mbasmanova:
Allow creating SpatialRDD with a mix of geometry types by making SpatialRDD a generic class parameterized over geometry type. Restricting objects stored in Spatial RDD to instances of Geometry enhances type safety and reduces the need for dynamic cast.

Updated core/pom.xml to add dependencies to allow 'mvn clean install' to succeed.


jiayuasu:
@mbasmanova Thanks for your contribution!

PR #122, opened 2017-09-19
Combine XMin/XMax/YMin/YMax into a single aggregation

mbasmanova:
Before this change, SpatialRDD.analyze method computed the extent of the whole RDD by computing min-X, min-Y, max-X and max-Y values using 4 aggregations. This PR introduces custom aggregation to compute an extent of an RDD as an Envelope in a single scan of the data.


core/src/main/java/org/datasyslab/geospark/spatialRDD/SpatialRDD.java
```diff
@@ -388,24 +385,52 @@ public void buildIndex(final IndexType indexType,boolean buildIndexOnSpatialPart
...
+			return combine(object.getEnvelopeInternal(), agg);
+		}
+	}
+
+	/**
+	 * Boundary.
+	 *
+	 * @return the envelope
+	 */
+	public Envelope boundary() {
```
jiayuasu: Use Aggregate to get the boundary is a good solution! But @mbasmanova could you please provide a unit test for this boundary aggregate function? Maybe you can use PointRDD or PolygonRDD to test its boundary.

mbasmanova:
@jiayuasu , thanks for review. I added unit tests for BoundaryAggregation class and updated PointRDDTest and PolygonRDDTest to assert correct boundary for SpatialRDD. Would you take another look?

PR #123, opened 2017-09-21
Fix QuadRectange.contains to process envelopes of points correctly

mbasmanova:
QuadRectange.contains used to return false for all rectangles of zero width or height including all rectangles which represent envelopes of points. As a result quad-tree partitioning of RDD of points never created more than 5 partitions making join operations very slow for large datasets. This PR fixes QuadRectange.contains to properly handle zero width or height rectangles.


jiayuasu:
Thanks @mbasmanova I have noticed this bug in other places. Thanks for solving this.

PR #130, opened 2017-10-01
Collect envelopes instead of full geometry objects when sampling

mbasmanova:
Spatial partitioning involves sampling and building partitioning scheme from the samples. Building of partitioning scheme happens in the driver and therefore requires samples to be transferred from the cluster to the driver machine. Each sample contains a geometry object and user data. Both, geometry object and user data can be large. In particular, complex polygons may contain lots of vertexes and holes and therefore be expensive to transfer other the network. Since partitioning building schemes use only bounding boxes of the samples, we can avoid transferring large geometries and user data to the driver by transforming samples into bounding boxes before collecting them to the driver. The PR updates SpatialRDD to do that and updates *Partitioning class to remove no longer needed logic of converting Geometry objects to bounding boxes.

This PR is built on top of two other PRs:

https://github.com/DataSystemsLab/GeoSpark/pull/126
https://github.com/DataSystemsLab/GeoSpark/pull/127

### DIFF
--- a/core/src/main/java/org/datasyslab/geospark/spatialPartitioning/HilbertPartitioning.java
+++ b/core/src/main/java/org/datasyslab/geospark/spatialPartitioning/HilbertPartitioning.java
@@ -6,93 +6,56 @@
  */
 package org.datasyslab.geospark.spatialPartitioning;
 
+import com.vividsolutions.jts.geom.Envelope;
+
 import java.io.Serializable;
 import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


jiayuasu:
@mbasmanova This PR seems to be in conflict. Could you please resolve this?

mbasmanova:
@jiayuasu , I rebased on top of new master. There are no conflicts now.

PR #132, opened 2017-10-05
Fix Circle.covers and Circle.intersects

mbasmanova:
The original implementation of Circle.covers(geometry) used to return true iff all 4 corners of the bounding box of the input geometry were covered by the circle. However, a line or a polygon may be covered by the circle while their bounding box is not. 

Similarly, the implementation of Circle.intersects(geometry) used to return true iff at least one corner of the bounding box of the input geometry was covered by the circle. However, a line cutting through the circle or a polygon covering the circle do intersect the circle, but this logic would report false.

This PR attempts to implement covers and intersects checks correctly.

### DIFF
--- a/core/src/main/java/org/datasyslab/geospark/geometryObjects/Circle.java
+++ b/core/src/main/java/org/datasyslab/geospark/geometryObjects/Circle.java
@@ -12,9 +12,13 @@ import com.vividsolutions.jts.geom.CoordinateSequenceComparator;
 import com.vividsolutions.jts.geom.CoordinateSequenceFilter;
 import com.vividsolutions.jts.geom.Envelope;
 import com.vividsolutions.jts.geom.Geometry;
+import com.vividsolutions.jts.geom.GeometryCollection;
 import com.vividsolutions.jts.geom.GeometryComponentFilter;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/datasyslab/geospark/geometryObjects/Circle.java
```diff
@@ -101,59 +105,160 @@ public void setRadius(Double givenRadius) {
...
+
+        if (other instanceof LineString) {
+            return covers((LineString) other);
+        }
+
+        if (other instanceof Polygon) {
+            return covers(((Polygon) other).getExteriorRing());
+        }
+
+        if (other instanceof GeometryCollection) {
```
jiayuasu: @mbasmanova Hi Masha, since you have added the support of GeometryCollection, can you please also add the support for MultiPoint, MultiLineString, MultiPolygon (see https://en.wikipedia.org/wiki/Well-known_text)? MultiPolygon is popular I think. Thank you!

mbasmanova:
@jiayuasu , since Mutli{Point, LineString, Polygon} classes share GeometryCollection as a base class, no code changes were needed to support Multi* geometries. I just added tests.

PR #137, opened 2017-10-12
Updated examples to use the new SpatialRDD.spatialPartitioning(partitioner) API

mbasmanova:


jiayuasu:
Hi @mbasmanova , please do not remove the user supplied mappers. Those mappers are templates for users.

mbasmanova:
@jiayuasu , sorry, didn't realize that. I'll put them back in.

PR #138, opened 2017-10-13
Make ShapeParser easier to use by allowing it to read directly from java.nio.ByteBuffer

mbasmanova:
In the upcoming PR, I'm using ShapeParser to parse ESRI shapefile representation of a geometry. This PR makes ShapeParser easier to use by allowing it to work directly with java.nio.ByteBuffer. I removed ShapeReader family of classes as they seem to be no longer needed.


jiayuasu:
@zongsizhang Zongsi, can you please verify this PR?

zongsizhang:
@mbasmanova @jiayuasu It's safe to directly use ByteBuffer when ShapeReader actually take use of ByteBuffer. But FSDataInputStream is not guaranteed to work in all file system, so I create ShapeReader family to isolate logics in ShpFileParser.java from FSDataInputStream.
I recommend to keep ShapeReader family and make ShpFileParser.java back to ShapeReader. 
All the other modifications looks great.
Really appreciate your work.

mbasmanova:
@zongsizhang , thanks for taking a look. I don't think I understand your comment. I'm seeing that ShapeParser is used only in PrimitiveShape#getShape. In that case the input is ByteBufferReader. ShpFileParser doesn't appear to use ShapeParser in any way. Am I missing something?

zongsizhang:
@mbasmanova So briefly we handle shapefile in following flow.
 shapefile - 1 > primitiveRDD<bytes> - 2 > ShapeRDD<Geometry>. 
What you are talking about is step 2 and you are right, only ByteBuffer is used and your modifications make sense. But for step 1, FSDataInputStream is not guaranteed to work in all file systems so I create ShapeReader to isolate the logics so that if new problems of file reading comes in, we don't need to modify codes in ShpFileParser.java.

mbasmanova:
@zongsizhang , thanks for explaining. I didn't realize that FSDataInputStream doesn't work properly in all cases and requires custom patching. I'm wondering if it makes sense to create a separate wrapper around FSDataInputStream for the purpose of patching any broken methods. ShapeReader name suggests that this class knows something about shape binary format. However, patching FSDataInputStream seems unrelated to whether it is used to read shapes or any other data.

zongsizhang:
@mbasmanova I agree, I still got a lot to learn about naming rules.Really appreciate your help. I thinks this PR is valuable, it makes codes clean. Are you going to do further modification or checkout changes about FSDataInputStream?

mbasmanova:
Abandoning in favor or https://github.com/DataSystemsLab/GeoSpark/pull/139

mbasmanova:
@zongsizhang , I folded this PR into https://github.com/DataSystemsLab/GeoSpark/pull/139 . I made sure to preserve the logic of patching FSDataInputStream.

PR #143, opened 2017-10-24
Allow user to designate build side of the join.

mbasmanova:
When both sides of the join fit in memory, best join performance is achieved by building an index from the larger side. However, when larger side of the join doesn't fit in memory, the only option is to build an index from the smaller side. This change updates JoinQuery.spatialJoin API to allow the user to choose which side to use when building an index.

Furthermore, the behavior of JoinQuery.SpatialJoinXXX APIs when useIndex = true has been modified as follows:

- if spatialRdd has an index, that index is used (same as before)
- otherwise, if queryRdd has an index, that index is used (threw an exception before),
- otherwise, an ephemeral index on spatialRdd is built on the fly(threw an exception before)


jiayuasu:
Hi @mbasmanova , this change was done by me. It seems I misunderstood your meaning before. In DynamicIndexLookup, you build index on windows not shapes, shapes will be used to stream. This behavior is different from other JoinQuery (with/without index). Am I right?

mbasmanova:
Hi @jiayuasu , I didn't realize you made that change. I'm using point-in-polygon use case as a guide for which side to load into an index. I assume that there are a lot of points and a relatively small number of polygons to join using 'contains' relationship. In this case, it makes sense to index the polygons and stream points. The join API takes two RDDs and assumes that 'contains' relationship is defined as "geometry from the second RDD contains geometry from the first RDD". WindowShapes refers to second RDD (polygons). Shapes refers to first RDD (points). Hence, DynamicIndexLookupJudgement needs to build index from the windowShapes and stream shapes. 

I did notice that join-using-index APIs assume that index is created from the first RDD (points), but I couldn't understand why. 

Overall, I think it makes sense to allow the user to designate build and stream sides. Since 'intersects' relationship is symmetric, the user already can make this choice by choosing which RDD to pass as first and second arguments. However, for 'contains' relationship this flexibility does not exist today.

Hope this makes sense.

jiayuasu:
@mbasmanova Got your point. But, say, we have M windows, N shapes, M<<N, having index will reduce the traverse complexity to O(log).  It is reasonable that M*(logN) << (logM)*N. That's why I build index on point partition which is larger

So, can I conclude that, DynamicIndexLookUp which streaming shapes and index windows, will decrease the join speed, in contrast to indexed SpatialJoinQuery which doesn't follow iteration model but indexes shapes?

mbasmanova:
@jiayuasu , I'm starting to understand where you are coming from. Indeed, you are correct, streaming larger set of points through a smaller index of polygons requires more computation than the other way around. However, to use this mechanism we need to ensure that large side can fit in memory. This means that we have to partition data into more partitions which in turn leads to higher duplication factor as objects start intersecting more and more partitions. Loading smaller partition in memory allows for much larger streaming partitions. Specifically, we can stream *infinitely* many points through a relatively fixed number of polygons.

What do you think about allowing the user to designate build and stream sides for the join?

jiayuasu:
@mbasmanova Hi Masha, I think it would be great if we can allow the user to designate build/stream sides.

We also need to allow this behavior to the other indexed SpatialJoin.

mbasmanova:
@jiayuasu , I'll work on adding this flexibility and update the PR when done.

mbasmanova:
@jiayuasu , I modified JoinQuery.spatialJoin API to allow user to designate build side of the join when building index dynamically as part of the join. I haven't modified the join API for the case when user builds index prior to invoking the join because I have some questions.

1. What is the use case for building an index prior to invoking the join? Would it be simpler to always build index dynamically?

1. Currently, SpatialJoinQuery API requires spatialRdd to have an index if useIndex = true. Would it make sense to relax this requirement to allow index on queryRdd? E.g. if useIndex = true and spatialRdd has an index, then use it. Otherwise, if queryRdd has an index, use that. If neither spatialRdd nor queryRdd have indexes, throw an exception.

jiayuasu:
@mbasmanova Thanks for updating this!

For question 1:
An common case is that:
I build the index on the left RDD, then I cache the left RDD. I may use a bunch of different right RDDs to join with this indexed left RDD, vice versa. The dynamic index building doesn't allow this.

In addition, in our future plan, I am thinking to persist the index on the disk because building index is time-consuming. The user just wants to build indexes once and use them multiple times. See Issue https://github.com/DataSystemsLab/GeoSpark/issues/35 This is another reason why I want to separate the buildIndex function.

For question 2:

Your idea is great. Thus, my thinking of the new spatial Join procedure is 

Step (1) Let the user to do spatial partitioning and indexing first. He can build the index on either left or right RDD (spatialPartitionedRDD). We don't control this and we don't change the old buildIndex API either.
Step (2A) The user calls SpatialJoinQuery with "useIndex=true" but without saying left/right Side. Our spatial join query will search on left and right spatialPartitioned RDD for an index. If they both have, pick the left index and use the right side as the streaming side. If no indexes exist, we build an index dynamically (with a warning) on the left instead of throwing an Exception.

Step (2B) The user calls SpatialJoinQuery with "useIndex=true" and "indexSide". Our spatial join query will search the specified indexSide for an index and use the other side for streaming. If no index exist on the specified indexSide, we build an index for him (with a warning).

In short, if the user builds an index in advance, we leverage that and don't dynamically build index again. If the user doesn't build the index in advance but still wants an index, we dynamically build for him with a logger.warning or logger.info. No exceptions any more. We need to add one more SpatialJoinQuery API which allows a parameter "indexSide".

What do you think of this?

mbasmanova:
@jiayuasu , I like your proposal and updated this PR accordingly. I didn't include any new APIs here thinking that it might be cleaner to do that in a separate PR.

PR #146, opened 2017-10-26
Fix java.lang.StackOverflowError in StandardQuadTree.insert

mbasmanova:
StandardQuadTree.insert could get into an infinite split cycle if there was a lot of data in the corner of the overall extent. This PR puts a cap on the depth of the tree equal to the number of desired partitions to avoid the problem.

java.lang.StackOverflowError
  at org.datasyslab.geospark.spatialPartitioning.quadtree.StandardQuadTree.findRegion(StandardQuadTree.java:62)
  at org.datasyslab.geospark.spatialPartitioning.quadtree.StandardQuadTree.insert(StandardQuadTree.java:153)
  at org.datasyslab.geospark.spatialPartitioning.quadtree.StandardQuadTree.insert(StandardQuadTree.java:159)
  at org.datasyslab.geospark.spatialPartitioning.quadtree.StandardQuadTree.insert(StandardQuadTree.java:169)
  at org.datasyslab.geospark.spatialPartitioning.quadtree.StandardQuadTree.insert(StandardQuadTree.java:159)

### DIFF
--- a/core/src/main/java/org/datasyslab/geospark/spatialPartitioning/QuadtreePartitioning.java
+++ b/core/src/main/java/org/datasyslab/geospark/spatialPartitioning/QuadtreePartitioning.java
@@ -33,8 +33,11 @@ public class QuadtreePartitioning implements Serializable {
 
     public QuadtreePartitioning(List<Envelope> samples, Envelope boundary, final int partitions, int minTreeLevel)
             throws Exception {
+        // Make sure the tree doesn't get too deep in case of data skew
+        int maxLevel = partitions;
+        int maxItemsPerNode = samples.size() / partitions;
         partitionTree = new StandardQuadTree(new QuadRectangle(boundary), 0,
-            samples.size() / partitions, 100000);
+            maxItemsPerNode, maxLevel);
         if (minTreeLevel > 0) {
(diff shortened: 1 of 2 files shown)


jiayuasu:
@mbasmanova Hi Masha, thanks for finding this issue. Max tree level should be a meaningful number rather than a magic number.

Correct me if I am wrong:
Since we use all leaf nodes as the partition boundaries, so the minimum MaxTreeLevel must be log<sub>4</sub>Partitions. But the max MaxTreeLevel shouldn't be too large otherwise it will lead to infinite loops.

Setting the max level to "partition" is acceptable. But is there a better number? If no, let's set it to "partition".

mbasmanova:
@jiayuasu , you are right. I'm thinking the same. I can't think of a better number though. Actually, I think we should switch to K-D-B tree [1] to achieve more even partitioning.

[1] https://en.wikipedia.org/wiki/K-D-B-tree

jiayuasu:
@mbasmanova We may make use of the KD tree in JTS. It is robust enough. I can do some changes on it to return hierarchy boundaries like QuadTree. But let me merge this PR for now.

PR #192, opened 2018-02-20
Support UTF-8 charset in ShapefileReader

jiayuasu:
This PR contains the bugfix to solve the unreadable character in shapefile reader. See Issue https://github.com/DataSystemsLab/GeoSpark/issues/190

If you are working with Arabic, Chinese or Korean character sets in ESRI shapefile, you need to force GeoSpark shapefile reader to use UTF-8 charset.

Add the following line before GeoSpark ShapefileReader:
```
System.setProperty("geospark.global.charset","utf8")
```


pedromorfeu:
Does it need to be set in each executor?

yijialee:
now you should write like this:
```
System.setProperty("sedona.global.charset","utf8")
```

PR #220, opened 2018-04-19
DataFrameFactory added

tociek:
No

Created new utility object DataFrameFactory. Currently it has only one method called geometryDfFromShapeFile which is reading shape file and returns dataframe directly (no need for the user to go through rdd and Adapter steps). Additionally, if folder containing shapefile does contain .dbf file, column names are read from it and returned DF contains original column names
In the future we could also add methods which instead of returning rddshape column, return Geometry or shape class field.
## How was this patch tested?
Tested on two different shape files with different number of columns. Tests succeeded.

Yes, SQL section of the tutorial was updated.

### DIFF
--- a/docs/tutorial/sql.md
+++ b/docs/tutorial/sql.md
@@ -78,6 +78,23 @@ The output will be like this:
 |POLYGON ((-104.56...| 35|011|00933054|35011|    De Baca|      De Baca County| 06| H1|G4020|null| null|null|   A|6015539696|29159492|+34.3592729|-104.3686961|
 |POLYGON ((-96.910...| 31|109|00835876|31109|  Lancaster|    Lancaster County| 06| H1|G4020| 339|30700|null|   A|2169240202|22877180|+40.7835474|-096.6886584|
 ```
+## Load data from shape file directly into DataFrame
+
+Use the following code to create DataFrame from your shape file. If the folder provided in path parameter contains metadata .dbf file, column names will be adjusted accordingly
+```Scala
+import org.datasyslab.geosparksql.utils.DataFrameFactory
+val df = DataFrameFactory.geometryDfFromShapeFile(spark,"/data/nh_polygons")
+df.show
++--------------------+---------+
+|            rddshape|    nh_id|
++--------------------+---------+
+|POLYGON ((-31.265...|267741819|
+|POLYGON ((-31.265...|267741820|
+|POLYGON ((-31.255...|267759817|
+|POLYGON ((-31.255...|267759818|
+|POLYGON ((-31.255...|267759819|
+```
+
 ## Create a Geometry type column
 
 All geometrical operations in GeoSparkSQL are on Geometry type objects. Therefore, before any kind of queries, you need to create a Geometry type column on a DataFrame.
--- a/sql/src/main/scala/org/datasyslab/geosparksql/utils/DataFrameFactory.scala
+++ b/sql/src/main/scala/org/datasyslab/geosparksql/utils/DataFrameFactory.scala
@@ -1,5 +1,27 @@
 package org.datasyslab.geosparksql.utils
+import com.vividsolutions.jts.geom.Geometry
 
-class GeosparkDataframe {
+import org.apache.hadoop.fs.{FileSystem, Path}
+import org.datasyslab.geospark.formatMapper.shapefileParser.ShapefileReader
+import org.datasyslab.geospark.formatMapper.shapefileParser.parseUtils.dbf.DbfParseUtil
+import org.datasyslab.geospark.spatialRDD.SpatialRDD
+import scala.collection.JavaConversions._
 
+object DataFrameFactory {
+  def geometryDfFromShapeFile(spark: org.apache.spark.sql.SparkSession, path: String): org.apache.spark.sql.DataFrame ={
+    val srdd = new SpatialRDD[Geometry]
+    srdd.rawSpatialRDD = ShapefileReader.readToGeometryRDD(spark.sparkContext, path)
+    val df = Adapter.toDf(srdd, spark)
+    val inputPath = new Path(path)
+    val fileSys = FileSystem.get(spark.sparkContext.hadoopConfiguration)
+    val dbfs = fileSys.listStatus(inputPath).filter(i => i.getPath.getName.toLowerCase.contains("dbf"))
+    if(dbfs.length>0) {
+      val inputStream = fileSys.open(dbfs(0).getPath)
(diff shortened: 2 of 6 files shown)


sql/src/main/scala/org/datasyslab/geosparksql/utils/DataFrameFactory.scala
```diff
@@ -0,0 +1,27 @@
+package org.datasyslab.geosparksql.utils
+import com.vividsolutions.jts.geom.Geometry
+
+import org.apache.hadoop.fs.{FileSystem, Path}
+import org.datasyslab.geospark.formatMapper.shapefileParser.ShapefileReader
+import org.datasyslab.geospark.formatMapper.shapefileParser.parseUtils.dbf.DbfParseUtil
+import org.datasyslab.geospark.spatialRDD.SpatialRDD
+import scala.collection.JavaConversions._
+
```
jiayuasu: Can you please do the following change 
(1) Create a function called readFieldName()in ShapefileReader.java (a small function similar [readBoundBox](https://github.com/DataSystemsLab/GeoSpark/blob/master/core/src/main/java/org/datasyslab/geospark/formatMapper/shapefileParser/ShapefileReader.java#L126)). This function returns a list/array of strings which are the field name.
(2) Call ShapefileReader.readFieldName in geometryDfFromShapeFile instead of having the following
```
+    val inputPath = new Path(path)
+    val fileSys = FileSystem.get(spark.sparkContext.hadoopConfiguration)
+    val dbfs = fileSys.listStatus(inputPath).filter(i => i.getPath.getName.toLowerCase.contains("dbf"))
+    if(dbfs.length>0) {
+      val inputStream = fileSys.open(dbfs(0).getPath)
+      val dbfParser = new DbfParseUtil
+      dbfParser.parseFileHead(inputStream)
+      val x = dbfParser.getFieldDescriptors
+      val op = "rddshape" +: x.map(x => x.getFieldName).toSeq
```
(3) Add the necessary function call of DataFrameFactory in constructorTestScala it("Read shapefile to DataFrame"). This will run unit test on DataFrameFactory function.

docs/tutorial/sql.md
```diff
@@ -78,6 +78,23 @@ The output will be like this:
 |POLYGON ((-104.56...| 35|011|00933054|35011|    De Baca|      De Baca County| 06| H1|G4020|null| null|null|   A|6015539696|29159492|+34.3592729|-104.3686961|
 |POLYGON ((-96.910...| 31|109|00835876|31109|  Lancaster|    Lancaster County| 06| H1|G4020| 339|30700|null|   A|2169240202|22877180|+40.7835474|-096.6886584|
 ```
+## Load data from shape file directly into DataFrame
```
jiayuasu: Can you please remove the change on docs for now? I am thinking to have a more general API to load DataFrame fields names for different inputs  such as shapefile and GeoJSON.

tociek:
Changes made.
I have also added analyze() to DF creation and updated method name (Df to DF) to be inline with other methoids in Spark (ie .toDF)

tociek:
For the test, it would be good to have shape file with more columns than just the shape, so that test could compare if df column names match what was expected, but didn't have small shape file to put that into project resources.

sql/src/main/scala/org/datasyslab/geosparksql/utils/DataFrameFactory.scala
```diff
@@ -0,0 +1,19 @@
...
+import org.datasyslab.geospark.formatMapper.shapefileParser.ShapefileReader
+import org.datasyslab.geospark.spatialRDD.SpatialRDD
+import scala.collection.JavaConversions._
+
+object DataFrameFactory {
+  def geometryDFFromShapeFile(spark: org.apache.spark.sql.SparkSession, folderPath: String): org.apache.spark.sql.DataFrame ={
+    val srdd = new SpatialRDD[Geometry]
+    srdd.rawSpatialRDD = ShapefileReader.readToGeometryRDD(spark.sparkContext, folderPath)
+    srdd.analyze()
```
jiayuasu: Hi @tociek , can you please remove srdd.analyze()? I called srdd.analyze() in the unit test just for test purpose. The analyze here doesn't affect anything in DataFrame.

sql/src/test/scala/org/datasyslab/geosparksql/constructorTestScala.scala
```diff
@@ -107,11 +107,7 @@ class constructorTestScala extends FunSpec with BeforeAndAfterAll {
     }
 
     it("Read shapefile to DataFrame") {
-      var spatialRDD = new SpatialRDD[Geometry]
-      spatialRDD.rawSpatialRDD = ShapefileReader.readToGeometryRDD(sparkSession.sparkContext, shapefileInputLocation)
-      spatialRDD.analyze()
-      var shapfileDf = Adapter.toDf(spatialRDD, sparkSession)
-      shapfileDf.show()
+      DataFrameFactory.geometryDFFromShapeFile(sparkSession,shapefileInputLocation).show()
```
jiayuasu: @tociek Can you change the value of shapefilelocation to "shapefiles/dbf" instead of "shapefiles/polygon"? That shapefile has a dbf file which can be used to demonstrate the new feature.

tociek: done

sql/src/main/scala/org/datasyslab/geosparksql/utils/DataFrameFactory.scala
tociek: Done. Could you explain what analyze() does and why it is useless for DF?

tociek:
@jiayuasu , thanks for merging :smile: 
How can I use the newest version with spark 2.1, which I have on my cluster? In the newest version, registration of SQL function fails and ideally I would like to avoid creating workaround for that (unless there is no other way :smile: )

jiayuasu:
@tociek Run "mvn clean install -DskipTests" on this branch: https://github.com/DataSystemsLab/GeoSpark/tree/spark-2.1
You will get the jar for Spark 2.1

tociek:
Cool. Thanks!

jiayuasu:
@tociek I just released the 1.2.0-SNAPSHOT versions. You can use the hosted jar instead of compiling the jar by yourself.  Read this to learn how to use the SNAPSHOT version: http://datasystemslab.github.io/GeoSpark/download/GeoSpark-All-Modules-Maven-Central-Coordinates/#snapshot-versions

PR #232, opened 2018-05-13
spatio-temporal 3DRtree Octree

hucaser:
No
Add spatio temporal index
## How was this patch tested?
All tested
No

### DIFF
--- a/core/src/main/java/org/datasyslab/geospark/SpatioTemporalObjects/Cube.java
+++ b/core/src/main/java/org/datasyslab/geospark/SpatioTemporalObjects/Cube.java
@@ -1,0 +1,176 @@
+package org.datasyslab.geospark.SpatioTemporalObjects;
+
+import com.vividsolutions.jts.geom.Envelope;
+
+/**
+ * The Class Cube.
+ */
+public class Cube implements java.io.Serializable{
+
+    public Envelope envelope;
+
+    public double minz;
+
+    public double maxz;
+
+
+    /**
+     * Instantiates a new Cube.
+     *
+     * @param envelope the spatial envelop
+     * @param beginTime the temporal begin time
+     * @param endTime the temporal end time
+     */
+    public Cube(Envelope envelope, double beginTime, double endTime) {
+        this.envelope = envelope;
+        this.minz = beginTime;
+        this.maxz = endTime;
+    }
+
+    /**
+     * Instantiates a new Cube.
+     *
+     * @param minx the minx
+     * @param max the max
+     * @param miny the minx
+     * @param maxy the maxy
+     * @param miniz the minz
+     * @param maxz the maxz
+     */
+    public Cube(double minx, double maxx, double miny, double maxy, double minz, double maxz) {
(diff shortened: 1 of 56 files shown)


jiayuasu:
@hucaser 
First of all, thanks a lot for committing this PR in such a short time period.

I generated this summary after reading your code. Please address the issues mentioned below and ping us (@jiayuasu @Sarwat) when you are done. Please feel free to contact me if you need some help to improve your code.

# Summary

You extend SpatialRDD to create a SpatialTemporalRDD which can accomodate point objects with an additional time attribute.

### SpatialTemporal RDD

1. For data-partitioning, you implemented four partitioning methods: 
	* Uniform grids on time
	* Uniform grids on spatial coordinates
	* OC-Tree on time and coordinates (implemented by extending GeoSpark Quad-tree code)
	* 3D-RTree on time and coordinates (implemented by directly calling another open source project)
2. For local-index, you directly used STR-Tree and Quad-Tree in GeoSpark. This means the local index querying only works with spatial coordinates, not spatial-temporal attributes


### SpatialTemporal Query Operator

You didn't implement any spatial-temporal query operators. In other words, your code currently only works with spatial range queries, not spatial-temporal queries.


### Unit tests (for code correctness)

Your correctness test cases tested the correctness of four data partitioning methods you wrote.  

### Unit tests (for performance)
Your performance tests don't have test data. I am not able to confirm the performance results in your slides. The following observations are based on my understanding of your code.

1. Your performance test of the data partitioning step runs in parallel. This test is thorough.
2. Your performance test on local range query step is not in parallel. You only tested it on one RDD partition.
3. Your performance test on local range query step only works with spatial range queries, not spatial-temporal query.


# Pros

Here are the pros I found in the code:

1. You understands what is spatial-temporal data and why we need to do data partitioning on both spatial coordinates and time attributes. You also know what are the challenges of implementing spatialtemporal support in GeoSpark.
2. The code is compatible with GeoSpark. From your code, we can know you read GeoSpark's source code a lot and understand the system structure well.
3. Your code on spatial-temporal data partitioning is good.
4. Your code contains some performance tests and correctness checks. They look good.


# Cons

Here are the cons I found in the code:

1. Doesn't support spatial-temporal queries.
2. The code quality is not high. Some copy-paste of GeoSpark source code and other open source projects. This leads to many redundant code and potential bugs/conflicts. It is not a good idea to merge this code into GeoSpark for now. You need to clean your code a lot.
3. The performance test only tested some small datasets. I guess its size is less than 1GB.
4. Only supports point objects.

hucaser:
@jiayuasu Thanks for giving me so much advice.

I directly use STR-Tree and Quad-Tree as local-index.  I try to use Octree and 3DRtree in local index, but there are some error when spark SerDe objects.

I didn't implement a spatio-temporal query, I may add it to the code next time. I query the object use RDD.filter(new QueryFilter(queryCube)), I don't know whether it is correct. But the QueryFilter is use the spatio-temporal information.

The test data has hundreds of MB, it is a little big to push to github, so I didn't commit it. I will clean the redundant code next time.

PR #241, opened 2018-06-12
Added ST_IsValid function

devlucasc:
Added ST_IsValid function to check if geometry is valid or not. This function return true if geometry is valid or false it else. No documentation added yet.


jiayuasu:
@devlucasc This is a great patch! The only thing that needs to be updated is the doc😁

Issue #288, opened 2018-11-04
ST_isValid is not working properly - falls when the geometry is null


avshRepo:
## Expected behavior
validate that the geometry is not valid

## Actual behavior
exception is thrown when a geometry is null

the exception that shown is: Exception in thread "main" org.apache.spark.SparkException: Job aborted due to stage failure: Task 3 in stage 2.0 failed 1 times, most recent failure: Lost task 3.0 in stage 2.0 (TID 5, localhost, executor driver): java.lang.RuntimeException: java.lang.NullPointerException

## Steps to reproduce the problem
give file where one of the geometries is null 

## Settings

GeoSpark version = 1.2.0

Apache Spark version = 2.3.0

JRE version = 1.8

API type = Scala

PR #295, opened 2018-11-21
Check for null in ST_IsValid function #288

netanel246:
Return null if the geometry is null in [ST_IsValid](https://github.com/DataSystemsLab/GeoSpark/blob/de05c0eed36237fd828c24375376403c86457022/sql/src/main/scala/org/apache/spark/sql/geosparksql/expressions/Functions.scala#L235) function
Define nullable = true
Add a test to cover it

#288 
Add check for null geometry, and return null when it happens.
This behavior defined by PostGIS for the [ST_IsValid](https://postgis.net/docs/ST_IsValid.html) function docs
## How was this patch tested?
I added a test that query ST_IsValid with null geometry
There are no documentation updates


jiayuasu:
@netanel246 Merged. Thank you!

Issue #297, opened 2018-11-22
Add option to skip invalid geometries in result RDD of GeoJsonReader.readToGeometryRDD


AntonPeniaziev:
## Expected behavior
When creating a spatial rdd from geoJson file, there should
an option to pass a flag invalidGeomsEnabled with default value True.
## Actual behavior
An rdd will always conatin invalid geometries if there are some invalid geometries in geoJson file
## Steps to reproduce the problem
Create an rdd from geoJson file with some invalid geometry. There will be crash on any attempt to evaluate 
## Settings

GeoSpark version = 1.2.0

Apache Spark version =  2.3.0

JRE version = 1.8

API type = Scala and Java

jiayuasu:
@AntonPeniaziev This is good. Please go ahead and implement it. 

If the user wants to make the GeoJSON reader silent, we will do try-catch. If we catch an exception when loading a record in GeoJSON RDD, we should skip this geometry instead of printing null geometry.

Issue #291, opened 2018-11-06
ST_IsValid Failed to process all geometries if there are one Invalid Geometry


netanel246:
## Expected behavior
Write invalid geometry to log, and continue processing the valid geometries
## Actual behavior
Throws invalid Geometry exception and not proceeding to the next geometries
## Steps to reproduce the problem
Using this code, we can read invalid geometry a, and then a valid geometry, b.
I would expect to write to log the invalid geometry and continue processing the valid one,
and not get an exception because of the invalid geometry a.
```
var testtable=sparkSession.sql(
        "SELECT ST_IsValid(ST_GeomFromWKT('POLYGON((0 0, 10 0, 0 0))')) AS a, " +
        "ST_IsValid(ST_GeomFromWKT('POLYGON ((30 10, 40 40, 20 40, 10 20, 30 10))')) as b"
      )
```

## Settings

GeoSpark version = 1.1.3, 1.2.0

Apache Spark version = 2.3.0

JRE version = 1.8

API type = Scala With Java

jiayuasu:
@netanel246 I don't think this is an issue. ST_GeomFromWKT should throw exceptions for wrong geometries which don't satisfy the meaning of a polygon (e.g., only 2 points, or, wrong WKT syntax). ST_Invalid is to find the geometries that are "topology invalid" but in correct geometries syntax.

The users should make sure all input WKT strings are correct in terms of grammar.

Please read PostGIS explanation: https://postgis.net/docs/ST_IsValid.html

netanel246:
OK. Thanks for your replay.
I have a problem that when I am using Generic Spatial RDD with some of the
geometries are invalid, I got an exception and the process terminate. I
expected that if there are some invalid geometries, they will be written to
log, and the process will continue to work on the valid ones.

I have dirty data, which can contain invalid geometries, and I always fail
on those invalid geometries and therefore can't process that data.

I found that I can handle this if I will add try...catch block to swallow
the exception and write to log the error, and returning null instead of
geometry.
Do you think it's a valid bug?

netanel246:
OK. But how can I handle Generic SpatialRDD which contains a mix of valid and invalid geometries?
In the current version, I got an exception, and I expected to process the other valid geometries, and basically swallow the invalid geometries and write the error to log.

jiayuasu:
[...]

PR #298, opened 2018-11-24
issue #297: added an overload for GeoJsonReader.readToGeometryRDD wit…

AntonPeniaziev:
…h additional parameter to skip invalid geometries in result RDD

#297 
Add an option to create an RDD from geoJson, skipping invalid geometries
## How was this patch tested?
Added a test with both valid and invalid geometries.
No


AntonPeniaziev:
Hi @jiayuasu  ,

1. Regarding CI fail, I have a technical question: I've added a json file for test into core\target\test-classes that is git ignored, so I've pushed my file with --force. I indeed can see the file pushed successfully https://github.com/AntonPeniaziev/GeoSpark/commit/c483fd505b6e8e66e70db42427f7a2e0c38d9b2a 

However, the CI still can't see it that is I've got an error GeoJsonReaderTest.onceExecutedBeforeAll:56 NullPointer .

Can you please elaborate on how should I add new test sources to ignored directory?

2.  The objective is not to validate the invalid geometries rather give to user an option to build and RDD from geoJson that will contain only **topologically** valid geometries. Note that this cannot be done with ST_Isvalid expression, since it receives single geometry, when GeoJsonReader.readToGeometryRDD receives a path to file. Theoretically there is an option to make a filter,  somehow with lambda in java when using
val isvalidop = new IsValidOp(geometry)
    isvalidop.isValid

but I haven't succeed to do so due to Java version issues.  I think public API should be more friendly in that way.

3. Another topic I haven't covere here is huge geoJsons with **invalid syntax** geometry.
I've seen @netanel246 's issue #291 relates it. 
We actually have a use case with a big GeoJson (ten's of millions rows) that contains few geometries with invlid syntax, for example - a polygon that is not closed or a polygon with zero area, like
POLYGON((a, b, a))
Of course we are not talking about total syntax mess, but there are GEO processing tools in the market that can create those kind of **syntactically** invalid geometries that I've mentioned above and we need to deal with them. 

I've managed to solve it with some try-catch warappers around each readGeometry.  Do you think it's worse another issue opened or I can just edit a title a bit?

Regards,

Anton

jiayuasu:
@AntonPeniaziev 
Got your point.

Since GeoSpark has a bunch of different format readers: GeoJSON, Shapefile, WKT, WKB, HDF...

So I recommend that, don't do validation for GeoJSON.

Instead, please create a new function in SpatialRDD:
https://github.com/DataSystemsLab/GeoSpark/blob/master/core/src/main/java/org/datasyslab/geospark/spatialRDD/SpatialRDD.java#L118

called RemoveInvalidGeometry(). It checks rawSpatialRDD and filters out all geometries whose isvalidop.isValid is false.

This provides a very generic API for all input formats. So the user first loads all geometries to rawSpatialRDD and then call "RemoveInvalidGeometry" to remove invalid geometries.

Most importantly, this will help all other users.

Regarding your Point3, we probably need to have a separate PR to discuss.

Thank you again for your effort!

Jia

AntonPeniaziev:
@jiayuasu ,
Regarding RemoveInvalidGeometry function
I've looked at SpatialRdd code and I don't see any proper way to implement it since after filtering invalid geometries index becomes obsolete and calculating new one could sometimes be against users intentions. Like it was mentioned in Geospark paper there are cases when absence of index improved the performance.

The only aternative I can think about is implemeting <Format>Reader for each type like GeoJsonReader,
and use the functionality I've already implemented in FormatMapper (use allowInvalidGeometries flag)

Tell me please what do you think is better approach.

Regards,

Anton

jiayuasu:
@AntonPeniaziev OK.  Understand, it makes sense. Let me first accept this PR and then figure out how to do the same check for other formats.

In your another PR for Issue https://github.com/DataSystemsLab/GeoSpark/issues/299, please make sure you differentiate invalidTopologyGeom and invalidSyntaxGeom in the overload API, otherwise the user may feel confused.

Issue #242, opened 2018-06-13
join optimization


geoHeil:
A common tip to improve the spatial join for postGIS is to use:

```
SELECT 
  a.id, b.id,
  CASE 
    WHEN ST_Contains(a.geom, b.geom)
    THEN b.geom
    ELSE ST_Intersection(a.geom, b.geom)
  END AS geom
FROM a, b
WHERE ST_Intersects(a.geom, b.geom);
```
instead of a naive intersection.
This could also be helpful for geostark / maybe directly integrated into the join?

jiayuasu:
@geoHeil Thanks. Will figure out a way to include this trick in spatial join.

A more detailed description of PostGIS ST_Intersection trick is here: https://postgis.net/2014/03/14/tip_intersection_faster/

avshRepo:
@jiayuasu hi you close this issue becasue i have already enhanced it in the following pull request #301

jiayuasu:
@avshRepo Hi Avsh, please note that this optimization is for join and needs override Spark logical plan "rules" to perform query rewrite. What you did is for the comparison between two objects.

PR #301, opened 2018-11-29
enhance: enhanced ST_Intersection according to the PostGis Tip. solve…

avshRepo:
… #242

yes, i implemented the proposal change

i changed ST_Intersection according to PostGis tip

## How was this patch tested?
this patch was tested with the previous ST_Intersection test

no

### DIFF
--- a/sql/src/main/scala/org/apache/spark/sql/geosparksql/expressions/Functions.scala
+++ b/sql/src/main/scala/org/apache/spark/sql/geosparksql/expressions/Functions.scala
@@ -206,6 +206,7 @@ case class ST_Transform(inputExpressions: Seq[Expression])
   override def children: Seq[Expression] = inputExpressions
 }
 
+
 /**
   * Return the intersection shape of two geometries. The return type is a geometry
   *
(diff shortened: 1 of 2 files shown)


sql/src/main/scala/org/apache/spark/sql/geosparksql/expressions/Functions.scala
```diff
@@ -215,13 +216,27 @@ case class ST_Intersection(inputExpressions: Seq[Expression])
...
+
+    if (isIntersects && isLeftContainsRight) {
+      return new GenericArrayData(rightgeometry)
+    }
+
+    if (isIntersects && isRightContainsLeft) {
+      return new GenericArrayData(leftgeometry)
+    }
+
+    return new GenericArrayData(GeometrySerializer.serialize(leftgeometry.intersection(rightgeometry)))
```
jiayuasu: Thank you very much for implementing this!

But I think you forgot one case:

If (!isIntersects) return empty geometry

sql/src/main/scala/org/apache/spark/sql/geosparksql/expressions/Functions.scala
```diff
@@ -215,13 +216,27 @@ case class ST_Intersection(inputExpressions: Seq[Expression])
...
+    val leftgeometry = GeometrySerializer.deserialize(inputExpressions(0).eval(inputRow).asInstanceOf[ArrayData])
+    val rightgeometry = GeometrySerializer.deserialize(inputExpressions(1).eval(inputRow).asInstanceOf[ArrayData])
+
+    val isIntersects = leftgeometry.intersects(rightgeometry)
+    val isLeftContainsRight = leftgeometry.contains(rightgeometry)
+    val isRightContainsLeft = rightgeometry.contains(leftgeometry)
 
+
+    if (isIntersects && isLeftContainsRight) {
+      return new GenericArrayData(rightgeometry)
```
jiayuasu: This probably should be new GenericArrayData(GeometrySerializer.serialize(rightgeometry)).

I think the current unit test doesn't cover this case. Please add some test cases for it.

Thank you very much!

sql/src/main/scala/org/apache/spark/sql/geosparksql/expressions/Functions.scala
jiayuasu: Yes, I think you can do that. Just make sure you add one test case for this scenario.

sql/src/main/scala/org/apache/spark/sql/geosparksql/expressions/Functions.scala
avshRepo: hi, i created a new pull request with the changes you asked:)

avshRepo:
@jiayuasu I added the changes that you asked

PR #303, opened 2018-11-30
This was a part of CSE-598 Topics in Data Systems Project.

abhishekthorat02:


harryprince:
@abhishekthorat02 could we use `st_difference` right now?

PR #313, opened 2018-12-19
Add ST_IsDisjoint

avshRepo:
no

added ST_IsDisJoint

## How was this patch tested?
i added tests

yes, i updated


jiayuasu:
@avshRepo Thank you for the contribution! ST_IsDisjoint is supposed to be an outer join. Currently, all GeoSpark join in JoinQueryDetector is designed for inner join. Therefore, you should develop an outer join algorithm, instead of using an existing GeoSpark join.

Please correct me if I am wrong. Please see PR #255.

PR #331, opened 2019-02-21
BugFix in Adapter.toDF when creating a DataFrame from a ShapeFile

Serhuela:
No.

When using Adapter.toDF in order to create a DataFrame from a shapeFile, a bug can occur. If the *.dbf has at least one row with the last field empty, the bug will occur.
The problem is the default behaviour of scala's split function; it trims trailing empty fields.

## How was this patch tested?
A new unitary test called "Read shapefileWithMissing -> DataFrame" has been created on file "adapterTestScala.scala". A specific ShapeFile has been created to spot the bug.

Not sure. Maybe not.


jiayuasu:
@Serhuela Thanks for your contribution!

PR #411, opened 2020-01-23
shading transitive geotools to private geospark

geoHeil:
https://github.com/DataSystemsLab/GeoSpark/issues/410

shade transitive dependencies to not pollute classpath
upgrade maven shade plugin as the old one had bugs

## How was this patch tested?
run the included test suite

none required

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -175,7 +175,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
-                <version>2.1</version>
+                <version>3.2.1</version>
                 <executions>
                     <execution>
                         <phase>package</phase>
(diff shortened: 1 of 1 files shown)


geoHeil:
locally it works fine , Travis has a StackOverflow error - potentially due to memory constraints

jiayuasu:
@geoHeil It looks like the new PR has a much higher mem requirement. I re-built the PR again. It still failed.  I may need to investigate a little bit.

geoHeil:
Please see the comment in https://github.com/DataSystemsLab/GeoSpark/issues/410 I think there is more to it than just shading.

jiayuasu:
@geoHeil I have a quick question. Will the "exclude" function in your case solve the problem? Usually, I use "exclude" many conflict jars in dependencies, and add the jar I want. As long as the API remain stable across versions, this should not be a problem. What do you think?

geoHeil:
I think a different approach is better @jiayuasu . There are two aspects we should take care of (https://github.com/DataSystemsLab/GeoSpark/issues/410):

- keeping it consistent for existing users. Adding everything (as you currently do) to make it easy to get started
- creating a separate module in maven which is meant for library integration into different projects or potentially complex workflows. Here, no shading (=no fat jar) should be applied at all.

PR #421, opened 2020-02-10
Fixed null value in geojson properties

netanel246:
Instead of throw exception, return "null" string
Added test

yes. [this one](https://github.com/DataSystemsLab/GeoSpark/issues/420
)
Fixed null value in geojson properties
## How was this patch tested?
added test
no docs


jiayuasu:
@netanel246 Thank you for your contribution!

PR #441, opened 2020-05-14
In method Iterator of class KnnJudgement,when adding element to Array…

yangshuxuan:
…,should check if the element is null


jiayuasu:
@yangshuxuan Could you explain why the element can be null? An RDD is not supposed to be null. Can you add a test case for the case you described?

PR #451, opened 2020-06-08
modify CRSTransform  to ignore datum shift（增加一个函数，可以忽略Datum Shift）

scially:
1. The original `CRSTransform`  considering the  datum  shift between different coordinate systems, but for some coordinate transformations with low accuracy , sometimes this parameter does not need to be considered, so we add method `public boolean CRSTransform(String sourceEpsgCRSCode, String targetEpsgCRSCode, boolean lenient)`
2. Because `SpatialRdd` uses `String sourceEpsgCRSCode` to save the coordinate information, so if the user wants to consider the difference of `Datum Shift`, he must define `CoordinateReferenceSystem` to set Datum, but it needs to be changed here `SaptialRdd` (such as replacing `sourceEpsgCRSCode` with `CoordinateReferenceSystem`, or not considering Datum Shift directly),  I need to discuss with you.
3. In the test method  `onceExecutedBeforeAll`  of  `ShapefileReaderTest` class, `hdfsURI` is `'localhost'`, but in some environments, Hadoop will prompt that it should be `127.0.0.1`, although the effect is the same,  switching to `127.0.0.1` is certainly no problem.
---
1. 原来GeoSpark里的`CRSTransform`默认是考虑了Datum Shfit，这会导致在有些坐标转换的时候出错（比如EPSG:4326, EPSG:4490），GeoTools会抛出异常，所以我建议增加一个`CRSTransform`，第三个参数是boolean，运行用户忽略这个差异（因为对于精度不高的坐标转换，不用考虑这个差异的）。
2. 因为`SpatialRdd`是用了 `String sourceEpsgCRSCode`来保存的坐标信息，所以用户如果要考虑`Datum Shift`差异，他应该是自己定义坐标参考系`CoordinateReferenceSystem`来设置Datum，但是这里就需要改动`SaptialRdd`了（比如用`CoordinateReferenceSystem`替换`sourceEpsgCRSCode`，或者就干脆直接不考虑Datum Shift），我就先不动，需要在和于嘉博士你讨论下。
3. 在`ShapefileReaderTest`测试类中`onceExecutedBeforeAll`的测试方法里，`hdfsURI`是`'localhost'`，但是在一些测试环境里，Hadoop会提示应该是`127.0.0.1`，虽然效果是一样的，但是换成`127.0.0.1`肯定是没有问题的。
1. `ShapefileReaderTest`, method: `onceExecutedBeforeAll`
2. `SpatialRDD`, method: `CRSTransform`
## How was this patch tested?
1. Existing test cases are enough
--- 
1. 现有用例满足测试要求
yes, it to be updated for  `CRSTransform` .


jiayuasu:
@scially I am not sure whether we should take into account datum shift. Currently, the SQL interface also just disable the exception.

PR #460, opened 2020-06-20
Add ST_NumGeometries

imdany:
[Implement complete SQL/MM and PostGIS functions in GeoSparkSQL](https://github.com/DataSystemsLab/GeoSpark/issues/200)

Added ST_NumGeometries. 

Returns the number of Geometries. If geometry is a GEOMETRYCOLLECTION (or MULTI*) return the number of geometries, for single geometries will return 1.

## How was this patch tested?

Unit Testing and Jar tested in Databricks environment

Yes.


sql/src/test/scala/org/datasyslab/geosparksql/functionTestScala.scala
```diff
@@ -285,5 +285,30 @@ class functionTestScala extends TestBaseScala {
...
+      val m2 = "MULTILINESTRING ((-29 -27, -30 -29.7, -36 -31, -45 -33), (-45.2 -33.2, -46 -32))"
+      val p1 = "POLYGON ((8 25, 28 22, 15 11, 33 3, 56 30, 47 44, 35 36, 43 19, 24 39, 8 25))"
+      val p2 = "MULTIPOLYGON(((0 0, 3 0, 3 3, 0 3, 0 0)), ((3 0, 6 0, 6 3, 3 3, 3 0)))"
+      val g1 = "GEOMETRYCOLLECTION(MULTIPOINT(-2 3 , -2 2), LINESTRING(5 5 ,10 10), POLYGON((-7 4.2,-7.1 5,-7.1 4.3,-7 4.2)))"
+
+
+      val testData = Seq((m1), (m2), (p1), (p2), (g1)).toDF("Geometry")
+
+      // Call transformation
+      var testDF = testData.selectExpr("Geometry", "ST_NumGeometries(ST_GeomFromText(Geometry)) as geom")
```
Imbruced, line 303: var is not necessary, use val instead

sql/src/test/scala/org/datasyslab/geosparksql/functionTestScala.scala
```diff
@@ -285,5 +285,30 @@ class functionTestScala extends TestBaseScala {
...
+      testDF.show(false)
+
+      val result = testDF.take(5)
```
Imbruced, line 306: You used spark implicits, with scala matchers assert can be used like 
```scala
testData.selectExpr("ST_NumGeometries(ST_GeomFromText(Geometry)) as geom").as[Int].collect().toList should contain theSameElementsAs List(1, 2, 1, 2, 3)
```
you have example here https://github.com/Imbruced/GeoSpark/blob/additional-sql-functions/sql/src/test/scala/org/datasyslab/geosparksql/functionTestScala.scala#L683

Imbruced:
@jiayuasu can you look on this ?

sql/src/test/scala/org/datasyslab/geosparksql/functionTestScala.scala
imdany: Thanks for your comments, @Imbruced, I like the GivenWhenThen approach, so I have follow the same approach here than in the link that you sent me.

I'll do the same in the other PR.

jiayuasu:
@imdany Please fix this as well. Thank you very much!

jiayuasu:
@imdany Unfortunately, the git conflict appears again. Could you please fix it?

PR #470, opened 2020-07-30
Fix wrong examples in the RDD tutorial

sekikn:

No

I found some examples in the "Spatial RDD application" tutorial don't work or are inconsistent with their context. This PR fixes them.

## How was this patch tested?

Ran `mkdocs serve` locally, and confirmed that copy-pasted snippets from the generated page worked manually.

Yes (it's a fix for the documentation itself)

### DIFF
--- a/docs/tutorial/rdd.md
+++ b/docs/tutorial/rdd.md
@@ -78,9 +78,9 @@ This file has 11 columns and corresponding offsets (Column IDs) are 0 - 10. Colu
 	
 Use the following code to create a PolygonRDD.
 ```Scala
-val polygonRDDInputLocation = "/Download/checkin.csv"
+val polygonRDDInputLocation = "/Download/checkinshape.csv"
 val polygonRDDStartOffset = 0 // The coordinates start from Column 0
-val polygonRDDEndOffset = 8 // The coordinates end at Column 8
+val polygonRDDEndOffset = 9 // The coordinates end at Column 9
 val polygonRDDSplitter = FileDataSplitter.CSV
 val carryOtherAttributes = true // Carry Column 10 (hotel, gas, bar...)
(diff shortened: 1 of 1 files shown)


docs/tutorial/rdd.md
```diff
@@ -181,7 +181,7 @@ We use [checkin.csv CSV file](#pointrdd-from-csvtsv) as the example. You can cre
 
 1. Load data in GeoSparkSQL.
 ```Scala
-var df = sparkSession.read.format("csv").option("delimiter", "\t").option("header", "false").load(csvPointInputLocation)
+var df = sparkSession.read.format("csv").option("header", "false").load(csvPointInputLocation)
```
jiayuasu, line 184: Can you keep the delimiter option here?

sekikn: Thank you for the review! Just before this example, the text says "We use checkin.csv CSV file as the example", so I thought this example should be coordinated with it.
How about keeping the delimiter option in the example, and modifying the sentence like "We use checkin.csv CSV file with tab delimiters as the example"?

jiayuasu: @sergiimk Sorry, I misunderstood the example. I think what you did is correct. I will merge this PR.

jiayuasu:
@sekikn Since you are already an Apache member, do you still need to submit a CLA to Apache saying "notifying project is Apache Sedona (incubating)"?

sekikn:
> do you still need to submit a CLA to Apache saying "notifying project is Apache Sedona (incubating)"?

No, I've already submitted my ICLA when I became a committer of another project, so I don't think I have to submit it again.

> We are about to import this repo into Apache infra in the weeks to come.

Great! I'm looking forward to it, and would like to keep contributing :)

PR #475, opened 2020-08-03
Fix Python compare_versions function

sekikn:

No

The compare_versions function in python/geospark/core/jvm/config.py seems to have a bug.
I think this function is supposed to return True if the left side is equal or higher than the right, but it's broken in some cases.

```
In [1]: from geospark.core.jvm.config import compare_versions                                                                 

In [2]: compare_versions('1.2.3', '1.0.0')                                                                                    
Out[2]: True

In [3]: compare_versions('1.0.0', '1.3.2')                                                                                    
Out[3]: False

In [4]: compare_versions('1.3.2', '1.2.3')                                                                                    
Out[4]: False <- ??
```

This PR fixes the above bug.

## How was this patch tested?

I locally ran `cd python; pytest tests/utils/test_geo_spark_meta.py` and confirmed it succeeded.

No

### DIFF
--- a/python/geospark/core/jvm/config.py
+++ b/python/geospark/core/jvm/config.py
@@ -11,7 +11,7 @@ FORMAT = '%(asctime) %(message)s'
 logging.basicConfig(format=FORMAT)
 
 
-def compare_versions(version_a: str, version_b: str) -> bool:
+def is_greater_or_equal_version(version_a: str, version_b: str) -> bool:
     if all([version_b, version_a]):
         version_numbers = version_a.split("."), version_b.split(".")
         if any([version[0] == "" for version in version_numbers]):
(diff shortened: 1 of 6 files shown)


Imbruced:
Thank you for finding the bug. I have no idea why your build is failing so I manually retried it.

sekikn:
@Imbruced Thank you for the review!

> I have no idea why your build is failing so I manually retried it.

It looks like there are some reasons. I think I can submit another PR for fixing them in a few days.

* For now, building GeoSpark generates geospark-1.3.2-SNAPSHOT.jar and geospark-sql_3.0-1.3.2-SNAPSHOT.jar by default, but the current regular expressions in .travis.yml don't match these filenames.

* Spark 2.4.4 is downloaded in .travis.yml, but it may not be compatible with 3.0.0.

* GeoTools were changed as 'provided' dependency due to the license incompatibility, so we have to download and deploy them manually to run full tests.

sekikn:
Thanks for the comment, that's a good improvement for clarity!
I've just updated my PR.

PR #485, opened 2020-10-11
Jtsplus python fix

Imbruced:
#484 
Fix failing python tests

## How was this patch tested?
-
No


Imbruced:
@jiayuasu That's PR to resolve an issue with Python tests. But one thing worries me. Can we add to POM, definition of fat jar for Python purposes ? Currently to make it work I have to add additional jars, which may not be user friendly and in addition it causes continuous CI script modification when we change dependencies versions.

jiayuasu (commented):
Hi @Imbruced ,

1. I notice there are some binary files in the test/resource folder, such as ".part-0000.crc"
Are these binary files necessary to be included?

2. The reason why we remove the geotools dependencies is that the license of these jar is not compatible with Apache Software Foundation. So they cannot be included in the source code and release.

In order to fix it, in out new release, we need to add the list of required GeoTools jars in our document website and ask users to download them.

Imbruced:
@jiayuasu 
1. Files are used by few tests.
2. I can try to create additional jar file which will be published on PyPI with wheel files (consists of those missing dependencies).

jiayuasu:
@Imbruced That sounds good!

Let's do it in a separate PR. The next PR in Sedona will be the name change for all paths, then you can finalize the wheel file there. We will have to generate our first Apache release on Spark 3.0 ASAP. Many users are waiting for it.

jiayuasu (approved):

PR #487, opened 2020-11-02
[SEDONA-3] Sedona release 3.0 python adapter [WIP]

Imbruced:
It is related to existing PR [Sedona-3]

Package name change

## How was this patch tested?
No additional tests were required

yes


Imbruced:
@jiayuasu Only changes in documentation left to finish.

Imbruced:
What do you think about changing artifacts names like below (I included them within this PR)
core -> sedona-core
sql -> sedona-sql
viz -> sedona-viz

Imbruced:
@jiayuasu  Can you restart my build ? I dont have such privileges, and I think Travis build get stuck.

jiayuasu:
@Imbruced Unfortunately,  I don't have the privilege as well. You can still push new content to your branch.

If you are done and have nothing to commit any more, I will review the PR later tonight and merge it to the 3.0 release branch to trigger the build. Then I will merge it back to the master branch only when it passes all test cases.

Imbruced:
@jiayuasu It's ready to review.

jiayuasu (approved):
Hi bro, thank you for your effort. Overall this looks good to me. I also agree with the names, sedona-core, sedona-sql, sedona-python-adapter, and sedona-viz.

However, I notice that you put sedona python adapter in a separate module. Could you please explain what jars need to be imported if the user wants to use Python API?

I assume it should be
```
sedona-core
sedona-sql
sedona-python-adapter
```
And the following combination will not work, right?
```
sedona-core
sedona-python-adapter
```

Imbruced:
@jiayuasu I dicided to add python-adapter for simplicity reasons, only sedona-python-adapter is required to get all functionality. As you can see within travis script. 

```yml
- find python-adapter/target/ -iregex "python-adapter\/target\/sedona-python-adapter-[0-9]\.[0-9]\.[0-9]-incubator\(-SNAPSHOT\)?\.jar" -exec cp {} $SPARK_HOME/jars \;
- (cd python;pipenv run pytest tests)
```
Assuming that user has Spark installed and properly set up Spark environment, he needs to add only one jar file. He can do that using SparkSession config argument or by putting the jar file within $SPARK_HOME/jars location.

I removed confusing upload_jars function (and also dependency, **find_spark** which can cause issues with other libraries versions). sedona-python-adapter.jar file will be included as additional files on PyPI and I think it should not be uploaded on maven.

jiayuasu:
@Imbruced I think the reason why travis-ci took a lot of time on "job received" is that we are now under Apache organization on Travis-ci. This organization has hundreds of builds every moment so the long waiting time is expected. You can take a look at here ("average queue time"): https://travis-ci.org/github/apache?tab=insights

Let me ask Felix whether we can improve the integration speed.

PR #496, opened 2020-12-12
[SEDONA-11] Add faster Python conversion from spatial rdd to df.

Imbruced:
SEDONA-3

- Faster converting SpatialRDD, RDD[Geometry, Geometry] to DataFrame (all string types, Python API) 
- Faster converting SpatialRDD, RDD[Geometry, Geometry] to geometry DataFrame (geometrytype and string type for user data, Python API) 
- toGeometryDf in scala Adapter for SpatialRDD and RDD[Geometry, Geometry] (GeometryUDT type for geometries, String type for whole userData)
- Fix issue with ```toDf(spatialPairRDD: JavaPairRDD[Geometry, Geometry], leftFieldnames: List[String], rightFieldNames: List[String], sparkSession: SparkSession)``` method within scala Adapter

## How was this patch tested?
- Essential test for ```toDf(spatialPairRDD: JavaPairRDD[Geometry, Geometry], leftFieldnames: List[String], rightFieldNames: List[String], sparkSession: SparkSession)```
- Tests toGeometryDf method Scala API
- Tests for python API 

Essential documentation updates will be on other PR [SEDONA-5]


sql/src/main/scala/org/apache/sedona/sql/utils/Adapter.scala
```diff
@@ -136,19 +136,60 @@ object Adapter {
...
+    }
+    var fieldArray = new Array[StructField](2)
+    fieldArray(0) = StructField("_c0", GeometryUDT)
+    fieldArray(1) = StructField("_c1", StringType)
+
+    val schema = StructType(fieldArray)
+    sparkSession.createDataFrame(rowRdd, schema)
+  }
+
+  def toGeometryDf(spatialPairRDD: JavaPairRDD[Geometry, Geometry], sparkSession: SparkSession): DataFrame = {
```
jiayuasu, line 155: I think you made a good improvement. We should directly convert the geometry type in RDD back to Geometry type in DataFrame. However, I believe we should put this improvement you made in the toGeometryDf to the old toDf function, instead of maintaining two separate functions. Would you mind grant me the write access to your fork and I can merge the two functions together?

jiayuasu (changes requested):

Imbruced:
@jiayuasu I granted you access to my fork.

jiayuasu:
@Imbruced I am still working on it. It will be done in one day.

Another feature I would like to add to the adapter is that we should also allow the user to carry the geometry column name. Currently, it is hard-coded as "geometry", "leftgeometry" and "rightgeometry". Since we are moving fast to the first release. I will put this in the second release. And I think this will change the public API so help from you will also be needed in the future.

jiayuasu:
@Imbruced I have merged toGeometryDf and toDf together. So the final API is toDf but this time it will convert a geometry object directly to a geom type column. See my commit: https://github.com/apache/incubator-sedona/pull/496/commits/4c284674a4fe01f0d7e5f392c581776446d3b620

I also change the signature of that function from scala List to scala seq for the ease of use. But I believe this small will not affect any downstream projects or API.

Could you please fix the Python API accordingly?

Imbruced:
@jiayuasu I added required changes. Please give me one additional day to make sure everything is fine.

Imbruced:
I think it is ready to merge.

jiayuasu (approved):

PR #507, opened 2021-02-04
[SEDONA-15] Add ST_MinimumBoundingRadius and ST_MinimumBoundingCircle functions

michaelmerg:
https://issues.apache.org/jira/browse/SEDONA-15

Add the new geometry processing functions:
- ST_MinimumBoundingRadius
- ST_MinimumBoundingCircle

## How was this patch tested?
See unit tests

Yes

### DIFF
--- a/docs/api/sql/GeoSparkSQL-Function.md
+++ b/docs/api/sql/GeoSparkSQL-Function.md
@@ -584,3 +584,31 @@ FROM df
 Input: `POINT (1 2)`
 
 Output: `POINT (2 1)`
+
+
+## ST_MinimumBoundingRadius
+
+Introduction: Returns a struct containing the center point and radius of the smallest circle that contains a geometry.
+
+Format: `ST_MinimumBoundingRadius(geom: geometry)`
+
+Since: `v1.?.?`
+
+Spark SQL example:
+```SQL
+SELECT ST_MinimumBoundingRadius(ST_GeomFromText('POLYGON((1 1,0 0, -1 1, 1 1))'))
+```
+
+
+## ST_MinimumBoundingCircle
+
+Introduction: Returns the smallest circle polygon that contains a geometry.
+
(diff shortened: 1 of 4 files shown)


docs/api/sql/GeoSparkSQL-Function.md
```diff
@@ -583,4 +583,32 @@ FROM df
...
+Output: `POINT (2 1)`
+
+
+## ST_MinimumBoundingRadius
+
+Introduction: Returns a struct containing the center point and radius of the smallest circle that contains a geometry.
+
+Format: `ST_MinimumBoundingRadius(geom: geometry)`
+
+Since: `v1.?.?`
```
jiayuasu, line 595: Please put Since: `1.0.1` here.

jiayuasu (changes requested):
Thanks for your contribution! Can you please add some test cases for the two new functions?

You can add your test cases here: https://github.com/apache/incubator-sedona/blob/master/sql/src/test/scala/org/apache/sedona/sql/functionTestScala.scala#L852

docs/api/sql/GeoSparkSQL-Function.md
michaelmerg: Updated

michaelmerg:
Sorry, forgot to commit the test cases. 
I'm not 100% happy with the `ST_MinimumBoundingCircle` test case because of the odd numbers when comparing the strings. Do you have any suggestions to improve of make it more robust?

jiayuasu:
@michaelmerg To make it robust, since ST_ functions internally call JTS functions, so probably you can call JTS function to validate the result in the test, instead of hard-coding the test result? What do you think?

michaelmerg:
@jiayuasu Yes, that's a good idea. I updated the test case.
Is there anything else that I need to do before we can merge this PR?

jiayuasu (approved):

jiayuasu:
@michaelmerg Thank you very much for the contribution!

PR #508, opened 2021-02-09
[SEDONA-14] Fix JSON representation of GeometryUDT schema

michaelmerg:
SEDONA-14

Fix the JSON representation of the GeometryUDT schema so that it can be parsed by `DataType.fromJson(...)` which is required to write a data frame in Parquet format.

## How was this patch tested?
Unit test added

No documentation update required

### DIFF
--- a/sql/src/test/scala/org/apache/spark/sql/sedona_sql/UDT/GeometryUdtTestScala.scala
+++ b/sql/src/test/scala/org/apache/spark/sql/sedona_sql/UDT/GeometryUdtTestScala.scala
@@ -16,17 +16,18 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+package org.apache.spark.sql.sedona_sql.UDT
 
-package org.apache.sedona.sql
-
+import org.apache.sedona.sql.TestBaseScala
 import org.apache.spark.sql.DataFrame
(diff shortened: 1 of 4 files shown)


sql/src/test/scala/org/apache/spark/sql/sedona_sql/UDT/GeometryUdtTestScala.scala
```diff
@@ -0,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
```
jiayuasu, line 2: Can you please move this test file to the path "org/apache/sedona/sql"?

jiayuasu (changes requested):

sql/src/test/scala/org/apache/spark/sql/sedona_sql/UDT/GeometryUdtTestScala.scala
michaelmerg: Sure, I thought to put it in the same package as the GeometryUDT class.

jiayuasu (approved):

jiayuasu:
@michaelmerg Thanks for your contribution!

PR #509, opened 2021-02-14
[SEDONA-16] Use a GeoTools Maven Central wrapper to fix notebooks in Python Binder

jiayuasu:

https://issues.apache.org/jira/browse/SEDONA-16

1. Use org.datasyslab:geotools-wrapper:geotools-24.0 in Jupyter Notebook
2. Move Jupyter Notebook to a binder folder to enable interactive live notebooks on binder

## How was this patch tested?

Tested in notebooks

Some are updated. Others to be fixed in a separate doc commit.


jiayuasu:
@Imbruced @netanel246 

Folks, as I tested in Jupyter notebook, if the users want to use GeoTools lib for ShapefileReader and CRS transformation, it will definitely fail because GeoTools's jai-core issue, unless the user compiles the source code.

Therefore, I moved GeoTools to Maven Central using a wrapper: https://github.com/jiayuasu/geotools-wrapper

PR #510, opened 2021-02-23
[SEDONA-17] Use WKB serde instead of the ShapeSerde

netanel246:
Removed unused ShapeSerde.java

[SEDONA-17](https://issues.apache.org/jira/browse/SEDONA-17)
Use WKB serde instead of the ShapeSerde in the core module (RDD API)
Removed unused ShapeSerde.java
## How was this patch tested?
tests are already exists
Nope.

Serde result comparison:
https://gist.github.com/netanel246/f85777761ebfc0a5ddef54170ea62f11

You can see that the SerDe time and size of the objects are better with the Shape Serde.
Do you know of any other Geo format that is binary, and possibly can give better results?


jiayuasu:
@netanel246 While waiting for your serde performance comparison, I wonder whether the serializer in SQL module could re-use the same serializer as Sedona core. If I remember correctly, in the past, they share the same serializer. This way, we don't have to maintain two serializers in Sedona.

netanel246:
@jiayuasu I think I can pull the WKB reader/writer from the core to another file, and use it in the SQL module instead of the [GeometrySerializer](https://github.com/apache/incubator-sedona/blob/2e60fc07b0eae78ccae3876d970e677fc9319c40/sql/src/main/scala/org/apache/sedona/sql/utils/GeometrySerializer.scala).

But I saw that there are also an indexSerde in both of the modules. I need to check if I can use it also from the SQL module

netanel246:
@jiayuasu 
I checked about the IndexSerializer and I think that we can leave it on the SQL module.
There is no overlap between the modules regards to the IndexSerializer.

So basically you can review the code now unless we have a better SerDe option instead of the WKB format.
What do you think?

core/src/main/java/org/apache/sedona/core/formatMapper/shapefileParser/parseUtils/shp/ShapeSerde.java
```diff
@@ -1,264 +0,0 @@
...
-
-        if (geometry instanceof MultiLineString) {
-            return serialize((MultiLineString) geometry);
-        }
-
-        if (geometry instanceof Polygon) {
-            return serialize((Polygon) geometry);
-        }
-
-        if (geometry instanceof MultiPolygon) {
```
jiayuasu, line 84: A few issues with the our old ShapeSerde serializer:

1. Does not serializer/deserialize GeometryCollection. A GeometryCollection object should be decomposed to its internal objects and serialize
2. Does not serialize/deserialize Multi-Multi object. It currently allows one level of nested objects

jiayuasu:
@netanel246 Thank you for your hard work!

I am not sure whether we should use WKB serializer now. It looks like the existing Serde is around 4 - 5 times faster than WKB SerDe in terms of serialization and 2 times faster in terms of deserialization. I believe this will make a huge difference for big data processing.

Two directions to move forward. Please let me know what you think.
1. Fix the issues in the existing Serde. See my comment above.
2. Leave the WKB serializer as an optional serializer and use the old Serde as the default SerDe for both Core and SQL. The user should only use it if they use geometries that are currently not supported by the old Serde.

The user should be able to choose the serializer as follows
```
.config("spark.serializer", classOf[KryoSerializer].getName) // org.apache.spark.serializer.KryoSerializer
.config("spark.kryo.registrator", classOf[SedonaKryoRegistrator].getName)
```

```
.config("spark.serializer", classOf[KryoSerializer].getName) // org.apache.spark.serializer.KryoSerializer
.config("spark.kryo.registrator", classOf[SedonaWKBKryoRegistrator].getName)
```

core/src/main/java/org/apache/sedona/core/formatMapper/shapefileParser/parseUtils/shp/ShapeSerde.java
netanel246: 1. I think that the old serde handles GeometryCollection. we also have a test for that - [GeometrySerdeTest](https://github.com/apache/incubator-sedona/blob/4800676929d3d07db2f7e6421a643ffd562f90a3/core/src/test/java/org/apache/sedona/core/formatMapper/shapefileParser/shapes/GeometrySerdeTest.java#L55)

This is handled at the write method of the [GeometrySerde](https://github.com/apache/incubator-sedona/blob/2e60fc07b0eae78ccae3876d970e677fc9319c40/core/src/main/java/org/apache/sedona/core/geometryObjects/GeometrySerde.java#L72)

2. What do you mean by the multi-multi object? For example, a multi polygon can have a nested polygons hierarchy? do you have some examples for that?

jiayuasu: @netanel246 I remember there was a GitHub issue that you have a multipolygon inside a multipolygon. But I couldn't find it. https://docs.google.com/spreadsheets/d/e/2PACX-1vQ4G8zwdfVhvoJAMWmBAGWNDTtE32Oh0dk95BF5HL996k0YCRDyU0cXry1xwuTGcQ4sixB0NTc4K4-0/pubhtml

1.Anyway, would you please make WKB serializer as an optional serializer? So if the old serializer does not support some very special geometries, the user could use the WKB serializer.
2. Would you please also make the SQL module serializer also allows both the old Serde and WKB serializer?

Thanks,
Jia

netanel246: I saw [here](https://github.com/geotools/geotools/blob/main/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/shp/ShapefileReader.java) that geotools already implemented Shapefile reader and writer.

I saw the source code and it seems that they already handle holes in polygons (the bug that cause us to move to WKB: [ST_GeomFromWkt and ST_GeomFromText converts multipolygon to polygon](https://docs.google.com/spreadsheets/d/e/2PACX-1vQ4G8zwdfVhvoJAMWmBAGWNDTtE32Oh0dk95BF5HL996k0YCRDyU0cXry1xwuTGcQ4sixB0NTc4K4-0/pubhtml))

I wonder if we can use that as a SerDe instead of creating two serdes that might be confusing for users.
What do you think?

jiayuasu: @netanel246 I wouldn't recommend GeoTools Serde due to their license issue. According to ASF policy, components that require Category X license has to be an optional feature. Serializer is the fundamental component in Sedona. If we use their library, I am afraid that we won't be able to graduate from the incubator.

netanel246: OK.
So I will add support for both of the serdes.

netanel246:
@jiayuasu I am opening a new branch for the optional WKB serde because this one is not relevant.
I will close it.

PR #511, opened 2021-03-01
[SEDONA-19] Update JoinParams to not always set useIndex to false

Kimahriman:
https://issues.apache.org/jira/browse/SEDONA-19

Fixes a bug where useIndex was always false when setting the index type and index build side. 

## How was this patch tested?
Existing tests still pass, but not sure how you would check if the index was used or not.

Updates the code to reflect what the documentation already says.


jiayuasu (approved):
Thanks for your contribution!

PR #514, opened 2021-03-11
[SEDONA-24] Add LineString interpolation functions

rendeer:
[[SEDONA-24]](https://issues.apache.org/jira/browse/SEDONA-24)

add geometry processing functions:

- ST_LineSubString
- ST_LineInterpolatePoint

## How was this patch tested?
scala test extended

yes

### DIFF
--- a/sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala
+++ b/sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala
@@ -734,7 +734,7 @@ case class ST_MinimumBoundingCircle(inputExpressions: Seq[Expression])
  *
  * @param inputExpressions
  */
-case class ST_LineSubString(inputExpressions: Seq[Expression])
+case class ST_LineSubstring(inputExpressions: Seq[Expression])
   extends Expression with CodegenFallback {
 
   override def nullable: Boolean = true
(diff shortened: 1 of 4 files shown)


sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala
```diff
@@ -727,6 +728,88 @@ case class ST_MinimumBoundingCircle(inputExpressions: Seq[Expression])
 }
 
 
+/**
+ * Return a linestring being a substring of the input one starting and ending at the given fractions of total 2d length.
+ * Second and third arguments are Double values between 0 and 1. This only works with LINESTRINGs.
+ *
+ * @param inputExpressions
+ */
+case class ST_LineSubString(inputExpressions: Seq[Expression])
```
jiayuasu, line 737: Thank you for your contribution! Not sure whether or not Spark function name is case sensitive or not. So would you please change the name to 'ST_LineSubstring' to make it consistent with PostGIS? https://postgis.net/docs/ST_LineSubstring.html

rendeer: Yes. You are right. It would be better to be consistent with PostGIS.

PR #515, opened 2021-03-19
[SEDONA-26] Add broadcast join support

Kimahriman:
https://issues.apache.org/jira/browse/SEDONA-26

Adds SQL broadcast join support. Broadcast hints are detected in the SQL Join detector and will execute a new `BroadcastIndexJoin` plan (that uses a `SpatialIndex` child plan, mostly for nice display purposes in the query plan).

## How was this patch tested?
New UT, still need to add more

Yes

Still have some work to do on this, but wanted to get some initial feedback on the overall approach. I opted to keep things purely in SQL land instead of trying to add more parameters to spatialJoin or add another new core package that SQL calls out to. ~Currently works for Spark 3, and it compiles in Spark 2 but the new test fails because broadcasts aren't currently detected in Spark 2.~ Figured out the Spark 2 support.

Also, I tried to better maintain the concept of left and right sides of the join, to more easily implement join types other than inner in the future. Rather than refer to the shapes as left and right where left.contains(right), I wanted to use different naming scheme. I went with window.contains(object), but very open to other ideas on how to separate the join sides versus <what>.contains(<what>).

TODOs:
- [x] Either figure out how to detect broadcast hints in Spark 2, or update the tests to only run the broadcast test on Spark 3
- [x] Add support for distance joins
- [x] Add this to the documentation somewhere
- [x] Probably add more unit tests

### DIFF
--- a/sql/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/SpatialIndexExec.scala
+++ b/sql/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/SpatialIndexExec.scala
@@ -26,7 +26,8 @@ import org.apache.spark.internal.Logging
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.{Attribute, BindReferences, Expression, UnsafeRow}
-import org.apache.spark.sql.execution.{SparkPlan, UnaryExecNode}
+import org.apache.spark.sql.execution.SparkPlan
+import org.apache.spark.sql.execution.exchange.Exchange
 
 
 
@@ -34,7 +35,7 @@ case class SpatialIndexExec(child: SparkPlan,
                             shape: Expression,
                             indexType: IndexType,
                             radius: Option[Expression] = None)
-  extends UnaryExecNode
+  extends Exchange
     with TraitJoinQueryBase
     with Logging {
 
--- a/sql/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/SpatialIndexExec.scala
+++ b/sql/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/SpatialIndexExec.scala
@@ -32,7 +32,8 @@ import org.apache.spark.sql.execution.{SparkPlan, UnaryExecNode}
 
 case class SpatialIndexExec(child: SparkPlan,
                             shape: Expression,
-                            indexType: IndexType)
+                            indexType: IndexType,
+                            radius: Option[Expression] = None)
   extends UnaryExecNode
     with TraitJoinQueryBase
     with Logging {
@@ -47,9 +48,12 @@ case class SpatialIndexExec(child: SparkPlan,
   override protected[sql] def doExecuteBroadcast[T](): Broadcast[T] = {
     val boundShape = BindReferences.bindReference(shape, child.output)
 
-    val resultRaw = child.execute().asInstanceOf[RDD[UnsafeRow]]
+    val resultRaw = child.execute().asInstanceOf[RDD[UnsafeRow]].coalesce(1)
 
-    val spatialRDD = toSpatialRdd(resultRaw.coalesce(1), boundShape)
+    val spatialRDD = radius match {
+      case Some(radiusExpression) => toCircleRDD(resultRaw, boundShape, BindReferences.bindReference(radiusExpression, child.output))
+      case None => toSpatialRDD(resultRaw, boundShape)
+    }
 
     spatialRDD.buildIndex(indexType, false)
     sparkContext.broadcast(spatialRDD.indexedRawRDD.take(1).asScala.head).asInstanceOf[Broadcast[T]]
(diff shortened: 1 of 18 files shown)


sql/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/SpatialIndexExec.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  override def output: Seq[Attribute] = child.output
+  
+  override protected def doExecute(): RDD[InternalRow] = {
+    throw new UnsupportedOperationException(
+      "SpatialIndex does not support the execute() code path.")
+  }
+
+  override protected[sql] def doExecuteBroadcast[T](): Broadcast[T] = {
+    val boundShape = BindReferences.bindReference(shape, child.output)
```
jiayuasu, line 48: It looks like, in the broadcast join, the spatial index is built on the broadcasted RDD (namely A).

Is it a better choice to build local spatial index on each partition of the distributed RDD (namely B)?

Since I assume the broadcasted RDD A is usually very small (a couple hundred records) while each partition of B might be larger?

jiayuasu:
@Kimahriman Thank you very much for your contribution, Adam! Indeed this is lots of work. Please refer to my comment above. But, that is a very minor issue anyway.

We are looking forward to the final version of this PR.

It looks like you also figured out that I use // spark2 anchor // spark3 anchor together with spark-version-converter.py (executed by Maven Pom.xml) to switch code between Spark 2 and 3. This is great!

If making it work with Spark 2 is not possible, please let me know.

Thank you again for the great job!

Jia

Kimahriman:
> It looks like you also figured out that I use // spark2 anchor // spark3 anchor together with spark-version-converter.py (executed by Maven Pom.xml) to switch code between Spark 2 and 3. This is great!
> 
> If making it work with Spark 2 is not possible, please let me know.
> 
> Thank you again for the great job!
> 
> Jia
Hah yeah I thought that was an interesting way to support both. Luckily a majority of it seems to be fine for both versions, it's just the detecting of the join hints that I haven't looked into much yet for Spark 2. Any idea how much longer you plan to support Spark 2?

sql/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/SpatialIndexExec.scala
Kimahriman: Yeah I guess that's why I thought it made sense to build the index on the broadcast sided. For one, you only have to do it once and can reuse the same index on all partitions. But I guess that main question is (that I have no idea the answer to): is it faster to build an index on N points or run N points through an existing index? I assumed the second one, but haven't actually tried the first.

Kimahriman:
I added a few snippets to the documentation. Let me know if you want any changes/any more info somewhere else. 

I also added extra condition support because I actually needed it for my use case, let me know what you think.

sql/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/SpatialIndexExec.scala
Kimahriman: Also, if you index the map side I believe that would require loading all geometries per partition into memory at once, limiting the scalability and requiring a certain amount of executor memory based on your data. But indexing the broadcast side let's the map side be fully lazy and iterative, optimizing memory usage.

jiayuasu:
@Kimahriman Thank you very much! This looks good to me.

@Imbruced Pawel, will the new broadcast join also work for Python API?

Kimahriman:
It will for any SQL/dataframe python. It's what I primarily use, usually in the same syntax as I wrote the tests

```python
import pyspark.sql.functions as F
pointdf.alias('pointdf').join(F.broadcast(polygondf).alias('polygondf'), on=F.expr('ST_Contains(polygondf.shape, pointdf.shape'))
```

Imbruced:
@jiayuasu That should work for sure.
@Kimahriman Great work, You will write Python tests or should I ? I have some spare time during this week and can accomplish that until Saturday. WDYT ?

Kimahriman:
Fine with me, I don't think I've tried running the python tests yet

Kimahriman:
Made one change so it still compiles on Spark 3.1.1. Also tried adding a basic test to python but couldn't get it to pass locally. I couldn't tell where it was getting sedona jars from, so it may just have been using the public jars? The only way to really test without looking at the plan is to check the partitioning of the joined dataframe.

Imbruced:
@Kimahriman To properly test locally you have to build source code and copy python-adapter jar to your spark location.

jiayuasu:
@Kimahriman Here is the instruction about how to run Sedona python test cases

Kimahriman:
I think I'm finally done making any changes, so let me know what you think. I figured anything beyond a basic Python test would mostly be testing that PySpark SQL works correctly.

jiayuasu (approved):

jiayuasu:
@Kimahriman This looks fantastic! Thank you very much for your contribution!

PR #516, opened 2021-03-29
[SEDONA-28] shape&wkb ser de

netanel246:
[SEDONA-28](https://issues.apache.org/jira/browse/SEDONA-28)
Added WKB serializer as an optional serializer and use the old Serde as the default SerDe for both Core and SQL. The user should only use it if they use geometries that are currently not supported by the old Serde.
The user should be able to choose the serializer as follows

```
.config("spark.serializer", classOf[KryoSerializer].getName) // org.apache.spark.serializer.KryoSerializer
.config("spark.kryo.registrator", classOf[SedonaKryoRegistrator].getName)
.config("spark.serializer", classOf[KryoSerializer].getName) // org.apache.spark.serializer.KryoSerializer
.config("spark.kryo.registrator", classOf[SedonaWKBKryoRegistrator].getName)
```

## How was this patch tested?
Using the existing tests. I Will add more tests soon
Not yet. Will be added soon.


netanel246:
@jiayuasu , I will be adding more commits soon to eliminate the code repetition of the ShapeGeometrySerde and the WKBGeometrySerde. 

You can start reviewing the other code if you want to.

jiayuasu:
Awesome. Once you finish the alternative WKB writer, we can publish the Sedona 1.0.1 release. There are a few relatively critical patches to be released in 1.0.1.

jiayuasu:
@netanel246 Hi Netanel, any update on this PR? If you don't have time to finish this in 1.0.1, we can put in in 1.1.0 and I will then work on 1.0.1 release because we have a few important patches to release.

netanel246:
@jiayuasu Hi Jia,
Unfortunately, I don't have much free time now, since there are more things to do here than I originally thought.
This can probably be pushed to 1.1.0, and I'll try to work on it for that release.

netanel246:
@jiayuasu ,
Did you encounter that exception using the Github Action? [Full error message](https://github.com/apache/incubator-sedona/pull/516/checks?check_run_id=2343806608#step:8:59)

`2021-04-14 14:20:33,533 WARN  [main] util.Utils (Logging.scala:logWarning(66)) - Service 'sparkDriver' could not bind on a random free port. You may check whether configuring an appropriate binding address.
2021-04-14 14:20:33,539 ERROR [main] spark.SparkContext (Logging.scala:logError(91)) - Error initializing SparkContext.
java.net.BindException: Cannot assign requested address: Service 'sparkDriver' failed after 16 retries (on a random free port)! Consider explicitly setting the appropriate binding address for the service 'sparkDriver' (for example spark.driver.bindAddress for SparkDriver) to the correct binding address.`

In my env, all the tests are passing. Only on the CI, I've got this exception which means that spark context cannot be created using this binding address.

I saw [here](https://stackoverflow.com/questions/52133731/how-to-solve-cant-assign-requested-address-service-sparkdriver-failed-after) that to solve that we need to add this ENV: `SPARK_LOCAL_IP="127.0.0.1"`

Any idea will be appreciated.
Thanks.

jiayuasu:
@netanel246 This is because you have multiple test cases create SparkSession which all try to take over the port (4040, I guess). To fix that, I suggest you use the initialize function in TestBase. Here is an example: https://github.com/apache/incubator-sedona/blob/master/core/src/test/java/org/apache/sedona/core/formatMapper/TestReadInvalidSyntaxGeometriesTest.java#L42

Let's see how it goes.

.github/workflows/java.yml
```diff
@@ -41,6 +41,7 @@ jobs:
     - env:
         SPARK_VERSION: ${{ matrix.spark }}
         SCALA_VERSION: ${{ matrix.scala }}
+        SPARK_LOCAL_IP: "127.0.0.1"
```
jiayuasu, line 44: You don't need this line.

core/src/test/java/org/apache/sedona/core/serde/SedonaKryoRegistratorTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public static JavaSparkContext sc;
+    public static SparkConf conf;
+
+    @BeforeClass
+    public static void onceExecutedBeforeAll()
+    {
+        conf = new SparkConf()
+                .setAppName(SedonaKryoRegistratorTest.class.getName())
+                .setMaster("local[2]")
```
jiayuasu, line 28: Fix this with the "initialize" example I mentioned.

core/src/test/java/org/apache/sedona/core/serde/SedonaWKBKryoRegistratorTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public static JavaSparkContext sc;
+    public static SparkConf conf;
+
+    @BeforeClass
+    public static void onceExecutedBeforeAll()
+    {
+        conf = new SparkConf()
+                .setAppName(SedonaWKBKryoRegistratorTest.class.getName())
+                .setMaster("local[2]")
```
jiayuasu, line 28: Fix this as well.

jiayuasu (changes requested):

jiayuasu:
@netanel246 What surprised me is that my last commit which only changed a comma in the doc also leads to the **Spark driver cannot find port** issue. But the master brach was built successfully two weeks ago. I guess probably the network setting in GitHub Action got changed from GitHub side. Let me explore it a bit.

jiayuasu:
@netanel246 The issue was due to an issue of the latest GitHub Action Ubuntu image update. It looks like their update leads to the failure of lots of Spark-based CI tests. Their update affects both Ubuntu 20.04 and Ubuntu 18.04 which is the one we used in our Github Action.

https://github.com/actions/virtual-environments/issues/3185

.github/workflows/java.yml
netanel246: This is the same workaround of `"spark.driver.bindAddress", "127.0.0.1"` but we only need to set it once, as an env variable.

As you can see [Here](https://github.com/apache/incubator-sedona/pull/516/checks?check_run_id=2344894037) it works.

I only need to apply that to python and the examples CI's.

Also, they have found the problem, and they will work on that on Monday (as described [here](https://github.com/actions/virtual-environments/issues/3185#issuecomment-821758343))

core/src/test/java/org/apache/sedona/core/serde/SedonaKryoRegistratorTest.java
netanel246: I am using tearDown after each test that closes SparkContext so it should be fine.

I can't use the initialize because I need to init Spark by myself in order to check that the `SedonaWKBKryoRegistrator` is working.

Also, this test is [Ignored](https://github.com/apache/incubator-sedona/pull/516/files/394c496c3a879579047ca42aa807d4230694d9e1#diff-8afb6529894c10113000ff3128d33e00db24f9e122e8282f668c5be64c1d84c6R17) so it wasn't run at all at this commit.

core/src/test/java/org/apache/sedona/core/serde/SedonaWKBKryoRegistratorTest.java
netanel246: Same as [above](https://github.com/apache/incubator-sedona/pull/516#discussion_r615296944)

netanel246:
@jiayuasu , Hi Jia,
I've finished the core implementations. (except for the IndexSerde which I don't know if we want to provide a WKB serde for that also)

Now, I will try to implement it for the SQL module.

If you have time, I will be more than happy that you start to review my code for that part.

**P.S.**
As you can see I am still using the CI, because I fixed the problem(Java and Scala build) with the workaround of `SPARK_LOCAL_IP: "127.0.0.1"`.

jiayuasu:
@netanel246 The current implementation looks good to me! Sedona has its own index serializer. So both WKB serializer and Shape Serializer will re-use the same index serializer.

Looking forward to the SQL part!

jiayuasu (approved):

jiayuasu:
@netanel246 Any update on the SQL part?

netanelma-tr:
Hi @jiayuasu ,
Unfortunately, I didn't have time for this till now.

There is a way to find which kryo serializer is used in runtime (without using the spark conf which isn't available for the scopes that I need)? I tried to search for that on the kryo docs but didn't find anything useful.

> 
> .config("spark.kryo.registrator", classOf[SedonaKryoRegistrator].getName)
> .config("spark.kryo.registrator", classOf[SedonaWKBKryoRegistrator].getName)
> 

otherwise, I won't know which to use for the SQL module.

I think that I need to find which is used and apply the SQL serializer according to this value.

What do you think about that?

netanelma-tr:
Hi @jiayuasu ,
I think I've found a solution. I will use the SedonaConf to get the type of Serializer in runtime.

jiayuasu:
> Hi @jiayuasu ,
> I think I've found a solution. I will use the SedonaConf to get the type of Serializer in runtime.

This is great. I wonder if this will lead to any performance downgrade since we need to fetch the conf for every ser/deser?

netanelma-tr:
> This is great. I wonder if this will lead to any performance downgrade since we need to fetch the conf for every ser/deser?

I was thinking to fetch the conf once only in the package level using the scala package object feature:
[package-objects](https://docs.scala-lang.org/tour/package-objects.html)

But I have two issues now:

1.  I need a SparkConf object which I don't have access to it at the package level.
2.  I don't have access to spark session from the Spark SQL functions, e.g: [Constructors.scala](https://github.com/apache/incubator-sedona/blob/master/sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Constructors.scala)

I would appreciate any thoughts you have.

jiayuasu:
@netanel246 How about something like:  `System.setProperty("sedona.global.charset", "utf8")`? Note that: this system property needs to be set in all executors (probably could be done together in a udf or map).

If this is not the best practice, I think we would rather resort to the old GeometrySer for Sedona SQL. What do you think?

netanelma-tr:
@jiayuasu , I've found a solution using the scala package object feature.
I need to fix 3 tests for scala and java and fix some of the python tests.

After that, I will add more tests and docs to the code.

Please let me know what do you think.

netanel246:
Hi @jiayuasu ,
I'm trying to pass this [test](https://github.com/apache/incubator-sedona/pull/516/commits/7597353d6de83ac0428edd4810cb9df81801186a#diff-5e924912624bfe719e43e0dada306e50a9cbb3681bec134ace263214d85fe0d5R55),  but I encountered a problem because I have two SparkSessions which always the first one (from TestBaseScala.scala) loads first, and failed the test.
It failed because the default serde is SHAPE, and this test will be pass only if the serde is WKB.

Do you have any ideas on how to make it work? I tried to stop the SparkSession before I started, but it's not working.

jiayuasu:
> Hi @jiayuasu ,
> I'm trying to pass this [test](https://github.com/apache/incubator-sedona/pull/516/commits/7597353d6de83ac0428edd4810cb9df81801186a#diff-5e924912624bfe719e43e0dada306e50a9cbb3681bec134ace263214d85fe0d5R55), but I encountered a problem because I have two SparkSessions which always the first one (from TestBaseScala.scala) loads first, and failed the test.
> It failed because the default serde is SHAPE, and this test will be pass only if the serde is WKB.
> 
> I need to find a way to change the spark.kryo.registrator and the sedona.serializer.type to use the WKB serde.
> 
> Do you have any ideas on how to make it work?
> I tried to:
> 
> 1. Stop the SparkSession before I started to run the test (using beforeEach)
> 2. Getting the current sparkSession and stop it (using SparkSession.getActiveSession.get)

I don't have any solution for now. Let me explore it a bit and see what I can do.

netanel246:
@jiayuasu,
Strangely, if the test runs alone, it succeeds. Only when I run it with another test that uses TestBasicScala does the problem occur.
It doesn't help to stop the SparkSession of TestBasicScala. It somehow still exists in my test context. Thus, the test failed because it loads the session with Shape Serde

jiayuasu:
@netanel246 I actually got an idea. Remember in Sedona SQL, we have both Scala and Java test. Why not put WKB SQL serializer test in Java test only? This way, I believe you can bypass this issue

ericsun95:
> @jiayuasu, Thank you.
> 
> 
> 
> Strangely, if the test runs alone, it succeeds. Only when I run it with another test that uses TestBasicScala does the problem occur.
> 
> It doesn't help to stop the SparkSession of TestBasicScala. It somehow still exists in my test context. Thus, the test failed because it loads the session with Shape Serde

This is a very common issue when you didn't close the session between tests. Could you just add one override aftereach function which call sparksession get active session close? Better also move the sparksession create part into beforeeach function.

netanel246:
> > @jiayuasu, Thank you.
> > Strangely, if the test runs alone, it succeeds. Only when I run it with another test that uses TestBasicScala does the problem occur.
> > It doesn't help to stop the SparkSession of TestBasicScala. It somehow still exists in my test context. Thus, the test failed because it loads the session with Shape Serde
> 
> This is a very common issue when you didn't close the session between tests. Could you just add one override aftereach function which call sparksession get active session close? Better also move the sparksession create part into beforeeach function.

Thank you for your response.
I've tried that and still, I have two spark sessions.

@jiayuasu ,
I also tried the java test and that works. I will commit that soon.
Thank you :)

jiayuasu:
@netanel246 Yes, I think you should use WKBserializer in all Java SQL tests including Adapter test. I believe this way you can bypass this issue.

netanel246:
> @netanel246 Yes, I think you should use WKBserializer in all Java SQL tests including Adapter test. I believe this way you can bypass this issue.

That seems to work.
But it implies that the java tests on Spark SQL are using ONLY the WKB serde.
Thank you :)

jiayuasu:
@netanel246 That's true. But I think this is the only solution we have. And I believe WKB serializer is more robust than GeomSerde except its performance. So we can do it this way.

ericsun95:
> > > @jiayuasu, Thank you.
> 
> > > Strangely, if the test runs alone, it succeeds. Only when I run it with another test that uses TestBasicScala does the problem occur.
> 
> > > It doesn't help to stop the SparkSession of TestBasicScala. It somehow still exists in my test context. Thus, the test failed because it loads the session with Shape Serde
> 
> > 
> 
> > This is a very common issue when you didn't close the session between tests. Could you just add one override aftereach function which call sparksession get active session close? Better also move the sparksession create part into beforeeach function.
> 
> 
> 
> Thank you for your response.
> 
> I've tried that and still, I have two spark sessions. Furthermore, the other tests that run after that test keep failing with an error of stopped spark session.
> 
> I've also tried to update the TestBaseClass to use BeforeAndAfterEach, the tests that run after WKBGeometrySerializerTest failed with `Cannot call methods on a stopped SparkContext. ` which I do not understand because the beforeEach contains the creation of a spark session.
> 
> 
> 
> @jiayuasu ,
> 
> I tried to use the java test but it behaves the same. 
> 
> The test runs alone, but with the [adapterTestJava](https://github.com/apache/incubator-sedona/blob/a8c42896a8fc2f15b56fe817121c965adf0dc2f5/sql/src/test/java/org/apache/sedona/sql/adapterTestJava.java) it fails.
> 
> 
> 
> 

I don't think writing in Java could bring any differences. Instead, try just sparksession.close() in after each might help.

jiayuasu:
@netanel246 Hi Netanel, any updates on this one?

netanelma-tr:
@jiayuasu,
We need to fix the python tests that are failing.
I tried to look at that, but I am not sure if I could fix them (I am still trying to run the python tests on my local machine)

If I see that I am not able to fix them, I will ask @Imbruced to help me.

Imbruced:
@netanel246 @jiayuasu I will look at that today evening.

Imbruced:
@netanel246 I was able to quick fix the Python exceptions which occur after changes within this PR. 
* Fast solution require to force user to use "wkb" serializer. I can create PR with that but that cause one issue, Python API will have only one serialization method available (wkb). Selecting "shape" method breaks the API. 
* Second solution is to implement also shape serializer in Python API but... It is ugly solution in my opinion. Why ? The only way how I can get information about serialization methodology is using SedonaRegistrator and store that information after registering additional functions and after that dynamically chose which serializer should be used. Also I can try to get spark session and current arguments, but still this causes code smell. I do not like that solution.
```python

class GeometrySerializer:

    def serialize(self, obj):
        raise NotImplementedError("Geometry serializer should implement serialize method")

    def deserialize(self, datum) -> BaseGeometry:
        raise NotImplementedError("Geometry serializer should implement deserialize method")


class WKBSerializer(GeometrySerializer):

    def serialize(self, obj):
        return dumps(obj)

    def deserialize(self, datum) -> BaseGeometry:
        bytes_data = b''.join([struct.pack('b', el) for el in datum])
        geom = loads(bytes_data)
        return geom


class ShapeSerializer(GeometrySerializer):

    def serialize(self, obj):
        pass

    def deserialize(self, datum) -> BaseGeometry:
        pass


serializers = {
    "wkb": WKBSerializer(),
    "shape": ShapeSerializer()
}


class GeometryType(UserDefinedType):

    @classmethod
    def sqlType(cls):
        return ArrayType(ByteType(), containsNull=False)

    def fromInternal(self, obj):
        return self.deserialize(obj)

    def toInternal(self, obj):
        return [el - 256 if el >= 128 else el for el in self.serialize(obj)]

    def serialize(self, obj):
        from sedona.register import SedonaRegistrator
        serializer = serializers[SedonaRegistrator.serializer]
        return serializer.serialize(obj)

    def deserialize(self, datum):
        from sedona.register import SedonaRegistrator
        deserializer = serializers[SedonaRegistrator.serializer]
        return deserializer.deserialize(datum)

    @classmethod
    def module(cls):
        return "sedona.sql.types"

    def needConversion(self):
        return True

    @classmethod
    def scalaUDT(cls):
        return "org.apache.spark.sql.sedona_sql.UDT.GeometryUDT"
```
* Third, require some minimal code changes on jvm side, can you bring information which serialization method you use as first 4 bytes saved as integer ? Example 1 means wkb, 2 means shape and other numbers for next possible serializers ? What do you think ? I think this is much safer than the previous ones.

netanelma-tr:
@Imbruced , I think the third option is the best.
I will put a PR soon.

netanel246:
@Imbruced , any updates on this?

Imbruced:
> @Imbruced , any updates on this?

I am working on that branch (https://github.com/Imbruced/incubator-sedona/tree/shape-wk-serde-python-dev), major work is done, need to fix 2 tests with shape serde and ~6 with wkb serde. Hope tomorrow is landing day.

Imbruced:
@netanel246 its ready, give me some time to make sure it is 100% correct.

jiayuasu:
Great news! All test cases have passed.

Now let's finalize the docs.

@netanel246 Would you please add some steps to explain how to switch between two SerDe in

1. https://sedona.apache.org/tutorial/rdd/
2. https://sedona.apache.org/tutorial/sql/

@Imbruced Can you add some step to explain how to switch between two SerDe in 

1. https://sedona.apache.org/tutorial/core-python/
2. https://sedona.apache.org/tutorial/sql-python/

Once we have the docs in place, we will ask Yitao, Sedona R maintainer, to take a look at the R part.

Imbruced:
@jiayuasu Thats great, I should be able to add docs updates tomorrow. Hope I have enough privileges to push to this branch, if not I will create additional PR.

jiayuasu (approved):

PR #520, opened 2021-05-13
[SEDONA-29] Upgrade to Spark 3.1.1 and fix ST_Pixelize

jiayuasu:

https://issues.apache.org/jira/browse/SEDONA-29

1. Tweak all workflows to support Spark 2.4, 3.0, 3.1 CI
2. Rewrite ST_Pixelize using Expression instead of Generator to make Lateral View recognize this function

## How was this patch tested?

Passed the unit tests

Yes

### DIFF
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -14,11 +14,9 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        spark: [2.4.7, 3.0.2, 3.1.1]
+        spark: [2.4.7, 3.1.1]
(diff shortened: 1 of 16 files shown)


jiayuasu:
It looks like in Spark 3.1+, ST_Pixelize, the UDF that extends Spark Generator, cannot be recognized by Spark when run Lateral View.

I am still investigating what change on the Spark side leads to this issue. But most likely, I will have to rewrite ST_Pixelize to extend regular Spark Expressions and return an array of Pixels. Then use Spark explode in Lateral View.

PR #525, opened 2021-05-17
[SEDONA-35] address user-data mutability issue with `Adapter.toDF()`

yitao-li:
Signed-off-by: Yitao Li <yitao@rstudio.com>

https://issues.apache.org/jira/browse/SEDONA-35

I think we should create an issue for it.

The problem is currently if `Adapter.toDf()` is called with a `SpatialRDD` carrying non-spatial attributes (a.k.a "user data") in addition to spatial attributes, then the Spark dataframe returned from `toDf()` can only be evaluated successfully once. If it was  evaluated more than once, then a null pointer exception occurrs. If a Spark dataframe is not cached, then the underlying RDD could be evaluated more than once in many scenarios, so, the NPE problem could break quite a few use cases.

Essentially, the changes in this PR ensures `toDf()` respects the immutability of `RDD`s in Spark. The abovementioned NPE issue happens because the `Geometry` objects stored in the input `RDD` have their `userData` set to `null` within a `RDD.map` transformation, and this causes unexpected `null` values (at least when Spark is running in "local" mode) when this transformation is evaluated the second time.

So, the short-term fix (as shown in this PR) is to operate on a copy of the `Geometry` object, rather than modifying the `Geometry` object that is already stored in the `RDD`.

Long-term fix: I think it would be a better practice to work with `RDD[ImmutableGeometry]`s where `ImmutableGeometry` is a sub-class of `Geometry` that does not permit modifications of its existing data fields, but permits copying its underling data to a mutable `Geometry` object. I'm unsure whether there may be noticeable perf impact with this type of change though and I hope to get a second opinion on this. But this feels like, in principle, the correct thing to do, because `RDD`s are by design supposed to be immutable in Spark, and correctness issue arises if the immutability contract is not followed.

## How was this patch tested?

Unit test (in progress)

Not applicable -- there is no change to documentation

### DIFF
--- a/sql/src/main/scala/org/apache/sedona/sql/utils/Adapter.scala
+++ b/sql/src/main/scala/org/apache/sedona/sql/utils/Adapter.scala
@@ -113,11 +113,14 @@ object Adapter {
   }
 
   def toDf[T <: Geometry](spatialRDD: SpatialRDD[T], fieldNames: Seq[String], sparkSession: SparkSession): DataFrame = {
-    val rowRdd = spatialRDD.rawSpatialRDD.rdd.map[Row](f => {
-      var userData = f.getUserData
-      f.setUserData(null)
-      if (userData != null) Row.fromSeq(f +: userData.asInstanceOf[String].split("\t", -1))
-      else Row.fromSeq(Seq(f))
+    val rowRdd = spatialRDD.rawSpatialRDD.rdd.map[Row](geom => {
+      val userData = geom.getUserData
+      val geomWithoutUserData = geom.copy
+      geomWithoutUserData.setUserData(null)
+      if (userData != null)
+        Row.fromSeq(geomWithoutUserData +: userData.asInstanceOf[String].split("\t", -1))
+      else
+        Row.fromSeq(Seq(geom))
     })
     var cols:Seq[StructField] = Seq(StructField("geometry", GeometryUDT))
     if (fieldNames != null && fieldNames.nonEmpty) {
--- a/sql/src/main/scala/org/apache/sedona/sql/utils/Adapter.scala
+++ b/sql/src/main/scala/org/apache/sedona/sql/utils/Adapter.scala
@@ -169,8 +169,9 @@ object Adapter {
     if (fieldNames != null && fieldNames.nonEmpty) {
       val userData = "" + geom.getUserData.asInstanceOf[String]
       val fields = userData.split("\t")
-//      geom.setUserData(null) // Set to null will lead to the null pointer exception of the previous line. Not sure why.
-      (Seq(geom), fields)
+      val geomWithoutUserData = geom.copy
+      geomWithoutUserData.setUserData(null)
+      (Seq(geomWithoutUserData), fields)
     }
     else (Seq(geom), Seq())
   }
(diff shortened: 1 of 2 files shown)


jiayuasu:
@yitao-li This looks good to me. I actually ran into this issue before but wasn't able to identify the issue. Thanks for pointing this out. If you can add some test cases, I will accept this PR immediately.

ImmutableGeometry may not be a good choice because many geoms in JTS directly extend Geometry. If we add a ImmutableGeometry, we have to develop corresponding wrappers for all different geom types, which is not practical I believe

sql/src/main/scala/org/apache/sedona/sql/utils/Adapter.scala
```diff
@@ -166,8 +169,9 @@ object Adapter {
     if (fieldNames != null && fieldNames.nonEmpty) {
       val userData = "" + geom.getUserData.asInstanceOf[String]
       val fields = userData.split("\t")
-//      geom.setUserData(null) // Set to null will lead to the null pointer exception of the previous line. Not sure why.
-      (Seq(geom), fields)
```
yitao-li, line 170: I guess while we are at it, we should also address this issue here as well (even though it's not as bad as a null pointer exception) -- also, see test case with pairrdd that should cover this code path

sql/src/test/scala/org/apache/sedona/sql/adapterTestScala.scala
```diff
@@ -198,5 +199,51 @@ class adapterTestScala extends TestBaseScala with Matchers with GivenWhenThen{
       val spatialDf = Adapter.toDf(spatialRDD, sparkSession)
       assert(spatialDf.schema.fields(1).name == "LST")
     }
+
+    it("can convert spatial RDD with user data to a valid Dataframe") {
```
yitao-li, line 203: I think this should cover the case of spatial RDD -> dataframe

sql/src/test/scala/org/apache/sedona/sql/adapterTestScala.scala
```diff
@@ -198,5 +199,51 @@ class adapterTestScala extends TestBaseScala with Matchers with GivenWhenThen{
...
+        assert(rows.length == 1)
+        val geom = rows(0).get(0).asInstanceOf[Point]
+        assert(geom.getX == 40.7128)
+        assert(geom.getY == -74.006)
+        assert(rows(0).get(1).asInstanceOf[String] == "attr1")
+        assert(rows(0).get(2).asInstanceOf[String] == "attr2")
+      }
+    }
+
+    it("can convert spatial pair RDD with user data to a valid Dataframe") {
```
yitao-li, line 220: I think this should cover the case of pair RDD -> dataframe

sql/src/main/scala/org/apache/sedona/sql/utils/Adapter.scala
```diff
@@ -166,8 +169,9 @@ object Adapter {
...
+      val geomWithoutUserData = geom.copy
+      geomWithoutUserData.setUserData(null)
+      (Seq(geomWithoutUserData), fields)
```
yitao-li, line 174: I guess while we are at it, we should also address this issue here as well (even though it's not as bad as a null pointer exception) -- also, see test case with pairrdd that should cover this code path

yitao-li:
@jiayuasu Thanks for the feedback :slightly_smiling_face:

I had a chat with one of my colleagues about this as well, and we were both thinking if Java or Scala had the equivalent of the `const` qualifier in C++, then there would have been a simpler and more concise way to specify immutability requirements. But because such language construct does not exist in Java/Scala yet, we'll have to go with the current solution.

Also, I fixed a similar (but different) issue with PairRDD and created some test cases covering both code paths.

jiayuasu (approved):

jiayuasu:
@yitao-li Thanks for this PR. I will merge it and enter the release process of Sedona 1.0.1 which is mainly for bug fix purpose
Then I will get back to your R Interface PR.

PR #531, opened 2021-06-30
[SEDONA-30] Changes to Geotiff Loader

shantanuaggarwal2695:

https://issues.apache.org/jira/browse/SEDONA-30

Load bounding boxes into Wkt instead of geometry column in geotiff reader. This is because geometry UDT cannot be mounted in Spark cluster mode.

## How was this patch tested?

Passed unit tests

Yes


jiayuasu (approved):
LGTM

PR #532, opened 2021-07-05
[SEDONA-53]fix SpatialKnnQuery NullPointerException

chenpengchuan:
[[SEDONA-53]](https://issues.apache.org/jira/projects/SEDONA/issues/SEDONA-53?filter=allopenissues)
KnnJudgement.java
` for (int i = 0; i < k; i++) {
      res.add(pq.poll());
  }
`
change to :
`
if (pq.size() >= k) {
    for (int i = 0; i < k; i++) {
        res.add(pq.poll());
    }
} else {
    for (int i = 0; i < pq.size(); i++) {
        res.add(pq.poll());
    }
}
`

## How was this patch tested?
The unit test method testSpatialKnnQueryTopKlargerThanRowNumber() can be used in  test class  PointKnnTest
Not yet, I can add it if necessary.


core/src/test/java/org/apache/sedona/core/rangeJudgement/RangeFilterTest.java
```diff
@@ -0,0 +1,161 @@
+package org.apache.sedona.core.rangeJudgement;
```
jiayuasu, line 1: Could you please remove this test case because I have moved this test to somewhere else since your last PR. Thank you!

jiayuasu (changes requested):
Thanks for your contribution. Please find my comment above.

core/src/test/java/org/apache/sedona/core/rangeJudgement/RangeFilterTest.java
chenpengchuan: Thanks for your reminder, I have deleted the test case RangeFilterTest.java from the current pull request

jiayuasu (approved):

PR #534, opened 2021-08-06
[SEDONA-56] Fix broadcast join with AQE enabled

Kimahriman:
https://issues.apache.org/jira/browse/SEDONA-56

Small fix for SpatialIndexExec to just extend UnaryExecNode instead of Exchange to make it run with AQE enabled.

## How was this patch tested?
New unit test.

None needed, bug fix.


jiayuasu:
@yitao-li Hi Yitao, the R build failed because of some packages issue. Could you please take a look at this?

yitao-li:
@jiayuasu I don't think the R build failure is related to this change. The only thing we need to do is to use `secrets.GITHUB_TOKEN` to fetch R package(s) from GitHub. I'll send a PR to fix this soon.

jiayuasu (approved):

PR #536, opened 2021-08-09
[SEDONA-36] Parquet reader & Writers

swamirishi:

https://issues.apache.org/jira/projects/SEDONA/issues/SEDONA-36
https://issues.apache.org/jira/projects/SEDONA/issues/SEDONA-14

The issue is realted to serializing & deserializing Spatial RDDs into Parquet files with a schema
Serializing & Deserializing issues in Spatial RDDs into Parquet Files.
## How was this patch tested?
Unit Test cases & Integration tests are in place. Not touching existing code flow thus previous functionalities should not get affected
Yes


geoHeil:
Does this follow the https://github.com/geopandas/geo-arrow-spec arrow geo spec? Will it be possible to store metadata about the used projection (CRS) inside the parquet file (like geo-package). I believe geo-pandas is doing this as well.

core/src/main/java/org/apache/sedona/core/io/avro/utils/AvroUtils.java
```diff
@@ -0,0 +1,263 @@
...
+            }
+            case LINESTRING: {
+                geometryData = getCollectionFromLineString((LineString) geometry);
+                break;
+            }
+            case POINT: {
+                geometryData = getMapFromPoint((Point) geometry);
+                break;
+            }
+            case POLYGON:
```
geoHeil, line 247: What about MultiPolygons? Wouldn`t it be great if complex geometries are supported natively?

core/src/main/java/org/apache/sedona/core/spatialRDD/SpatialRDD.java
```diff
@@ -564,7 +588,45 @@ public void saveAsWKT(String outputLocation)
...
+            @Override
+            public Iterator<Tuple2<Void, GenericRecord>> call(Iterator<T> iterator) throws Exception {
+                Schema schema = AvroUtils.getSchema(geometryType, geometryColumn, userColumns, namespace,
+                                                    name);
+                Iterable<T> recordIterable = () -> iterator;
+                try {
+                    return StreamSupport.stream(recordIterable.spliterator(), false).map(geometry -> {
+                        try {
+                            GenericRecord genericRecord =
+                                    AvroUtils.getRecord(geometry, geometryType, geometryColumn, schema);
```
geoHeil, line 615: Why are we not simply choosing WKB for the serialization of the geometry column and keep the rest as a normal parquet file?

jiayuasu:
@swamirishi Thanks for your contribution. However, currently, Sedona DataFrame can be saved and loaded as Parquet files. I am not sure if we need this function for RDD. We have the adapter to travel between RDD and DataFrame.

Could you elaborate more about the purpose of this PR?

jiayuasu:
@swamirishi Thanks for your great work. This PR sounds interesting to me. However, this seems to be a big PR so we need to have a thorough review for this.

So far, this PR contains many file changes: 47 files changed so far. Can you avoid changing POM files unless it is necessary? Currently, you duplicate the dependencies to different Sedona modules. Please only keep your changes to Sedona core module if possible. This will be helpful for us to review.

Can you also make your PR pass our CI test?

@Imbruced @netanel246 @Sarwat Folks, please take a look when you have time.

swamirishi:
> @swamirishi Thanks for your great work. This PR sounds interesting to me. However, this seems to be a big PR so we need to have a thorough review for this.
> 
> So far, this PR contains many file changes: 47 files changed so far. Can you avoid changing POM files unless it is necessary? Currently, you duplicate the dependencies to different Sedona modules. Please only keep your changes to Sedona core module if possible. This will be helpful for us to review.
>  I had to shade Avro Parquet as there is a conflict with the Sernetcdf library. I have moved modules into dependency management. Those are not duplicate now a module would use only the libraries it would require. Thus making the dependency requirement for the project uniform throughout. 
> Can you also make your PR pass our CI test?
> Yeah have fixed the issue will push the code. Some issues after rebasing my branch with the master
> @Imbruced @netanel246 @Sarwat Folks, please take a look when you have time.

swamirishi:
@jiayuasu Is there a code base for org.datasyslab:sernetcdf? I see a bunch of classes google.common classes coming as part of libraries. This has version conflicts with the parquet library due to which I am getting runtime exceptions. In order to do away this I had to shade Parquet libs which isn't possible with the existing pom structure. Dependency management was the way to go about to circumvent the issue without touching the sernetcdf code. If there is a code base for sernetcdf can we look into changing the pom there?

jiayuasu:
@swamirishi The codebase of SerNetCDF is here: https://github.com/jiayuasu/SerNetCDF   which is owned by me. I think you can exclude some libraries in SerNetCDF dependency by using "<exclude></exclude>" tag.

I guess the problem is caused by the dependency in SerNetCDF: https://github.com/jiayuasu/SerNetCDF/blob/master/pom.xml#L46

swamirishi:
@geoHeil  @jiayuasu @Sarwat  I have fixed all build errors and also have reverted the pom changes. Please take a look at the PR.

Imbruced:
This PR is massive, need some time to review it fully.

Imbruced:
What about essential documentation updates ? Am I missing them ?

swamirishi:
> What about essential documentation updates ? Am I missing them ?

@Imbruced I have pushed the changes containing Documentation Comments

swamirishi:
@jiayuasu @Imbruced @netanel246 @Sarwat Did you take a look at the PR? I would like to get this merged as soon as possible so that I can start working on the second phase.

geoHeil:
Would this PR support multiple geometry columns in the parquet file? But one designated one for calculating the statistics at the row-group level?

swamirishi:
> Would this PR support multiple geometry columns in the parquet file? But one designated one for calculating the statistics at the row-group level?

Currently only one single geometry column per record. We can go ahead and support such features in later versions

swamirishi:
> > Would this PR support multiple geometry columns in the parquet file? But one designated one for calculating the statistics at the row-group level?
> 
> Currently only one single geometry column per record. We can go ahead and support such features in later versions. The stats are stored on Schemas Column Path. So for multiple Columns the Stats stored would be stored on different paths.

core/src/main/java/org/apache/sedona/core/io/parquet/ParquetFileReader.java
```diff
@@ -0,0 +1,60 @@
...
+     * @param userColumns
+     * @param inputPaths
+     * @return Avro Record RDD which needs to be deserialized into a GeometryRDD
+     * @throws IOException
+     */
+    public static JavaRDD<GenericRecord> readFile(JavaSparkContext sc,
+                                                  String geometryColumn,
+                                                  List<String> userColumns,
+                                                  String... inputPaths) throws IOException {
+        final Job job = Job.getInstance(sc.hadoopConfiguration());
```
Imbruced, line 40: is that mean that current solution works only in hadoop ecosystem ?

core/src/main/java/org/apache/sedona/core/showcase/Example.java
```diff
@@ -52,99 +52,99 @@
...
      */
     static FileDataSplitter PointRDDSplitter;
-
+    
     /**
      * The Point RDD index type.
      */
     static IndexType PointRDDIndexType;
-
+
```
Imbruced, line 90: I dont like that kind of changes ;)

core/src/test/java/org/apache/sedona/core/io/parquet/ParquetFileReaderTest.java
```diff
@@ -0,0 +1,87 @@
...
+        List<Point> l1 = pointRDD.getRawSpatialRDD().collect().stream().collect(Collectors.toList());
+        Collections.sort(l1, new Comparator<Point>() {
+            @Override
+            public int compare(Point o1, Point o2) {
+                return o1.compareTo(o2);
+            }
+        });
+        
+        SpatialRDD<Point> pointRDD1 = ParquetReader.readToGeometryRDD(sc,
+                                                                        Arrays.stream(testOutputPathDir.listFiles())
```
Imbruced, line 71: is that possible to load parquets from parent directory ex. 
test
    - *.parquet
    - *.parquet
    - *.parquet
    - *.parquet
Can we read this like readToGeometryRDD(..., test) instead of listing all files ?

Imbruced (changes requested):

core/src/main/java/org/apache/sedona/core/formatMapper/ParquetReader.java
```diff
@@ -0,0 +1,38 @@
...
+    public static <T extends Geometry> SpatialRDD<T> createSpatialRDD(JavaRDD rawRDD,
+                                                                      ParquetFormatMapper<T> formatMapper,
+                                                                      GeometryType geometryType) {
+        SpatialRDD spatialRDD = new SpatialRDD<T>(geometryType);
+        spatialRDD.rawSpatialRDD = rawRDD.mapPartitions(formatMapper);
+        return spatialRDD;
+    }
+    
+    public static <T extends Geometry> SpatialRDD<T> readToGeometryRDD(JavaSparkContext sc,
+                                                                       List<String> inputPath,
```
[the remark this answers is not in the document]
swamirishi: Yeah we can have method which can get all files in a directory

core/src/main/java/org/apache/sedona/core/formatMapper/ParquetReader.java
```diff
@@ -0,0 +1,38 @@
...
+public class ParquetReader extends RddReader {
+    public static <T extends Geometry> SpatialRDD<T> createSpatialRDD(JavaRDD rawRDD,
+                                                                      ParquetFormatMapper<T> formatMapper,
+                                                                      GeometryType geometryType) {
+        SpatialRDD spatialRDD = new SpatialRDD<T>(geometryType);
+        spatialRDD.rawSpatialRDD = rawRDD.mapPartitions(formatMapper);
+        return spatialRDD;
+    }
+    
+    public static <T extends Geometry> SpatialRDD<T> readToGeometryRDD(JavaSparkContext sc,
```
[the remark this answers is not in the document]
swamirishi: Currently we need to support reading of random parquet file input as well. We depend on the user to know the data in the parquet file. Otherwise it doesn't really make sense. As a user the user might want to do some projections thus one would need to know the files meta info. Though we can add a feature to identify geometry column in parquet files created by Sedona Framework.

core/src/main/java/org/apache/sedona/core/formatMapper/parquet/ParquetFormatMapper.java
```diff
@@ -0,0 +1,215 @@
...
+     * @return Geometry Object
+     */
+    private Geometry getGeometry(Coordinate[] coordinates, GeometryType type) {
+        Geometry geometry = null;
+        switch (type) {
+            case POINT:
+                geometry = factory.createPoint(coordinates[0]);
+                break;
+            case POLYGON:
+                geometry = factory.createPolygon(coordinates);
```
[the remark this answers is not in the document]
swamirishi: This is a defailt implementation wherein a user might just want convert an array of points into a closed polygon without holes. This is similar to supporting Offsets and number of columns length in text format for geometry column.

core/src/main/java/org/apache/sedona/core/geometryObjects/schema/CoordinateSchema.java
```diff
@@ -0,0 +1,39 @@
...
+            synchronized (CoordinateSchema.class){
+                if(SedonaUtils.isNull(schema)){
+                    schema = new CoordinateSchema();
+                }
+            }
+        }
+        return schema;
+    }
+
+//    public CoordinateSchema(String name, String namespace) {
```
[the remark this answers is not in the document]
swamirishi: Sure

core/src/main/java/org/apache/sedona/core/geometryObjects/schema/PolygonSchema.java
```diff
@@ -0,0 +1,37 @@
...
+    public static final String POLYGON = "polygon";
+    public static final String EXTERIOR_RING = "ex";
+    public static final String HOLES = "holes";
+    
+    private PolygonSchema() {
+        super(AvroConstants.SEDONA_NAMESPACE,POLYGON,
+              Lists.newArrayList(new Field(EXTERIOR_RING, CoordinateArraySchema.getSchema()),
+                                 new Field(HOLES, new UnionSchema(
+                                         new ArraySchema(CoordinateArraySchema.getSchema()),
+                                         new SimpleSchema(AvroConstants.PrimitiveDataType.NULL)))));
```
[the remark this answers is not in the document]
swamirishi: Saving null would be a better approach to go about I think. Though the default implmentation of the geotools returns an empty array. But I feel we should also support null types.

core/src/main/java/org/apache/sedona/core/io/parquet/ParquetFileReader.java
swamirishi: This is to import hadoop based configurations if at all there are any. Nothing else but that. Basically getting all the resource paths based configuration. If you look into the code of sc.textFile() it would be doing the same thing internally

swamirishi: > is that mean that current solution works only in hadoop ecosystem ?
This is to import hadoop based configurations if at all there are any. Nothing else but that. Basically getting all the resource paths based configuration. If you look into the code of sc.textFile() it would be doing the same thing internally

core/src/main/java/org/apache/sedona/core/showcase/Example.java
swamirishi: Sure let me just remove these changes

core/src/test/java/org/apache/sedona/core/io/parquet/ParquetFileReaderTest.java
swamirishi: We can, but if you see the text read and write contracts, they are the same. I wanted to keep the contracts as similar as possible excepting for the geometry column and projection column info. But if you feel we should have a contract for getting files based for FileGlobs we can sure do that.

core/src/main/java/org/apache/sedona/core/spatialRDD/SpatialRDD.java
```diff
@@ -129,9 +138,24 @@
...
     private int sampleNumber = -1;
-
-    public int getSampleNumber()
-    {
+    /**
+     * Geometry Type Defaults to Geometry Collection
+     */
+    private GeometryType geometryType;
+    
+    public SpatialRDD() {
```
jiayuasu, line 146: Is there a reason why you need to add geometry type to SpatialRDD? Because you need to use this information to define the type of geospatial objects in a Parquet file?

I don't think this is a good idea. Currently, this generic SpatialRDD is not bounded to a particular type. It allows mixed spatial objects in an RDD. E.g.,

POINT (XXX)
POLYGON (XXX)
MULTIPOLYGON (XXX)
POINT (XXX)

This scenario is actually quite common in real world. A WKT file of state boundaries or country boundaries may consist of both POLYGON and MULTIPOLYGON.

jiayuasu (changes requested):

jiayuasu:
@swamirishi 

Thank you again for your patience! I traveled back to WSU and the new semester just started. So I was kind of overwhelmed by many things on my plate. Now I am ready to work with you!

I personally are very excited about this PR but I think the reason why several committers hesitate to approve this PR is that 

(1) it does not have a detailed proposal to explain the proposed file structure and how it works. We know the overall idea of having a geospatial Parquet file and it is an awesome idea for sure. But it is unclear to us how this is achieved in this PR.
(2) it contains 44 file changes which are too many to be reviewed by a human... It touches several critical places in Sedona. For instance, it adds a new module called Parquet, adds some Sedona exceptions, change some dependencies. More importantly, it changes the core/SpatialRDD.java file which will affect many places.

Therefore, I believe here are some action items

(1) would you please add a Sedona website doc in this PR to explain 
	* The usage of Sedona Parquet reader and writer
	* The structure of the proposed geo Parquet file, especially the metadata structure
	* the algorithm: how it skips irrelevant data chunks by comparing the spatial query predicate and Parquet metadata

You can put this doc at https://sedona.apache.org/tutorial/ as a programming guide for geospatial Parquet

(2) for the file changes themselves
	* please fix the comments raised by @Imbruced 
	* Please address my comment for SpatialRDD change
	* please remove all whitespace file changes, and changes that just include "re-formatting the code"

Once these changes + doc are done and we fully understand what this PR is doing exactly, I will approve this PR. Let the customers give it a try. I believe once this PR is accepted, your follow-up PRs for the Parquet file will quickly pass our review.

Thank you again for your contribution. We really appreciate your contribution!

Jia Yu

core/src/main/java/org/apache/sedona/core/formatMapper/ParquetReader.java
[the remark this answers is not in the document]
swamirishi: The input Paths can be a path regex. I have changed the testcase accordingly

swamirishi:
@Imbruced I have addressed the review comments. Can you check if I need to change something else

jiayuasu:
@swamirishi 
Can you also address my comments? Then we can move to accept this PR.

swamirishi:
> Once these changes + doc are done and we fully understand what this PR is doing exactly, I will approve this PR. Let the customers give it a try. I believe once this PR is accepted, your follow-up PRs for the Parquet file will quickly pass our review.

@jiayuasu @Imbruced Please take a look at the PR now. I have made the required changes for supporting Generic Geometry Type SpatialRDD. I have added a section pertaining to SpatialRDD - parquet Deserialization & Serialization in the file docs/tutorial/rdd.md.

jiayuasu (approved):

jiayuasu:
@swamirishi 

Thank you for your contribution!

I will accept this PR into a separate branch: parquet-io. I will need to make some changes to this PR myself such as how to package dependencies.

Meanwhile, you can submit new PRs to the sedona/parquet-io branch to continue your next phase implementation.

Thanks,
Jia

PR #537, opened 2021-08-25
Make pyspark dependency of sedona python optional

sebbegg:

Yes - this implements https://issues.apache.org/jira/browse/SEDONA-59

This PR moves the `pyspark` dependency to a extra dependency called spark. Allowing to

- Get apache-sedona and the pyspark pip package in one go:
```bash
pip install apache-sedona[spark]
```
- Or get only apache sedona its dependencies except for pyspark:
```bash
pip install apache-sedona
```

## How was this patch tested?

Built python wheel locally and tested that the two `pip install` commands above work as expected.

Added a note in the install docs.

<sub>Sebastian Eckweiler (sebastian.eckweiler@daimler.com), Mercedes-Benz AG on behalf of MBition GmbH ([Imprint](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md))</sub>

### DIFF
--- a/docs/download/overview.md
+++ b/docs/download/overview.md
@@ -56,6 +56,11 @@ You need to install necessary packages if your system does not have them install
 ```bash
 pip install apache-sedona
 ```
+* Since version 1.1.0 pyspark is an optional dependency since spark comes pre-installed on many spark platforms.
+  To install pyspark along with Sedona Python in one go, use the `spark` extra:
+```bash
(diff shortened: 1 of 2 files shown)


sebbegg:
Note: I also bumped the pyspark dependency in `setup.py` to 2.4, so it's consistent with the `Pipfile`.

PR #538, opened 2021-08-31
Wkb shape serializer python

Imbruced:
SEDONA-28

Fix PYthon API after adding multiple serializers.

## How was this patch tested?
Python unit tests per serializer, included in github worklow matrix


Imbruced:
need to fix python worklow, it should include tests for shape and wkb serializer based on file content.

Imbruced:
And do not know why R workflow does not work.

python/sedona/core/serde/shape/enums.py
```diff
@@ -0,0 +1,22 @@
+from enum import Enum
+
+
+class GeomEnum(Enum):
```
jiayuasu, line 4: @Imbruced Any reason why this only enumerates a few types instead of all types?

Imbruced: I used Shape types fro ShapeType enum within package org.apache.sedona.core.formatMapper.shapefileParser.parseUtils.shp or you meant 
```java
        SHAPE(0),
        CIRCLE(1),
        GEOMETRYCOLLECTION(2),
        ENVELOPE(3);
```

Imbruced:
Did I broke something in R package ? WDYT is better to have shape and wkb serializers in matrix or run tests twice ? (Python API holds on serializers so it is essential to run tests twice per serializer).

Imbruced:
Did I broke something in R package ? WDYT is better to have shape and wkb serializers in matrix or run tests twice ? (Python API holds on serializers so it is essential to run tests twice per serializer).

Imbruced:
I can squash this PR (or we can squash while merging to master), sorry for additional commits, did anyone has a pleasure to use https://github.com/nektos/act to run workflows locally ?

Imbruced:
To be honest, this PR and major one needs documentation update, should I include them here or we update docs in @netanel246 PR ?

Imbruced:
@jiayuasu Can you add @netanel246 to reviewers ?

jiayuasu:
@Imbruced R doc is in the R module folder. Let me merge this PR now and go back to @netanel246 PR.

PR #543, opened 2021-09-16
[SEDONA-63] Skip empty partitions in NestedLoopJudgement

umartin:

https://issues.apache.org/jira/browse/SEDONA-63

Return immediately if one of the sides is empty in NestedLoopJudgement as already implemented in indexed join judgements.

## How was this patch tested?

Current set of test in Sedona should cover this change. 

Small performance improvement (~10%) observed with some of our Sedona jobs.

No documentation changes needed.


jiayuasu (approved):

jiayuasu:
Thanks for your contribution!

PR #545, opened 2021-09-21
[SEDONA-64] Broadcast dedupParams to improve performance

umartin:

https://issues.apache.org/jira/browse/SEDONA-64

Broadcast dedupParams to reduce task size.

## How was this patch tested?

Current tests in Sedona covers the change.

Test with production workloads. We see a 20% reduction in execution time for jobs with a large partition count.

Javadoc for an added public method.


jiayuasu (approved):

jiayuasu:
Thank you for your contribution!

PR #552, opened 2021-10-04
St make polygon

Imbruced:
SEDONA-45
New SQL function
## How was this patch tested?
Scala and Python unit tests
Yes

### DIFF
--- a/sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala
+++ b/sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala
@@ -24,7 +24,7 @@ import org.apache.sedona.sql.utils.GeometrySerializer
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.codegen.{CodegenContext, CodegenFallback, ExprCode}
-import org.apache.spark.sql.catalyst.expressions.{Expression, Generator}
+import org.apache.spark.sql.catalyst.expressions.{BoundReference, Expression, Generator}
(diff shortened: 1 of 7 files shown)


Imbruced:
I am adding new sql functions based on jira tasks. I will implement them one after another.

sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala
```diff
@@ -1167,3 +1167,46 @@ case class ST_SubDivideExplode(children: Seq[Expression]) extends Generator {
...
+      }
+
+    exteriorRing.toGeometry(input) match {
+      case geom: LineString =>
+        try {
+          val poly = new Polygon(geometryFactory.createLinearRing(geom.getCoordinates), holes.toArray, geometryFactory)
+          poly.toGenericArrayData
+        }
+        catch {
+          case e: Exception => null
```
jiayuasu, line 1201: Will it be better if you throw an exception here instead of giving a null?

jiayuasu (changes requested):

sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala
Imbruced: The reason why I changed to null instead of raising exception was to make that fault tollerant. I can change that, but I heard some complaints about that in some cases application is running for several minutes and then fail due to data quality.

jiayuasu: Sounds good to me!

jiayuasu:
I will merge this later once the 1.1.0 vote passes.

PR #555, opened 2021-10-09
[SEDONA-43] ST Geohash and ST_GeomFromGeohash

Imbruced:

https://issues.apache.org/jira/browse/SEDONA-43

ST_GeoHash function
## How was this patch tested?
unit tests scala and python
yes


Imbruced:
@jiayuasu If we can use it, it definitely reduce amount of work needed to complete. But as far I know it also uses protobuf lib.

PR #557, opened 2021-10-22
[SEDONA-67] Support Spark 3.2

Kimahriman:
https://issues.apache.org/jira/browse/SEDONA-67

Adds support for Spark 3.2, but maintains compatibilities.

Specifically to get 3.2 working I had to:
- Rework how functions are registered to use an API present in all supported versions of Spark.
- Add `withNewChildrenInternal` to all expressions. Without using `override`, it works both in the new version to override it as well as in older versions it's just ignored.
- Add a new profile for Spark 3.2 because it needs a newer version of jackson to test with

Additionally I cleaned up a few things:
- Removed assert statements from the eval method of expressions. This is an analysis time check that can be done on instantiation of the case class
- Removed a bunch of return statements to make things consistent

## How was this patch tested?
Existing tests, though I do need help figuring out how the github workflow should be updated.

Haven't looked at the documentation yet, probably need to update a compatibility page still

### DIFF
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        spark: [2.4.7, 3.0.3, 3.1.2, 3.2.0]
+        spark: [2.4.8, 3.0.3, 3.1.2, 3.2.0]
         scala: [2.11.8, 2.12.15]
         exclude:
           - spark: 3.2.0
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -14,12 +14,14 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        spark: [2.4.7, 3.0.2, 3.1.1]
-        scala: [2.11.8, 2.12.8]
+        spark: [2.4.7, 3.0.3, 3.1.2, 3.2.0]
+        scala: [2.11.8, 2.12.15]
         exclude:
-          - spark: 3.1.1
+          - spark: 3.2.0
             scala: 2.11.8
-          - spark: 3.0.2
+          - spark: 3.1.2
+            scala: 2.11.8
+          - spark: 3.0.3
             scala: 2.11.8
 
     steps:
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -15,19 +15,22 @@ jobs:
     strategy:
       matrix:
         include:
-          - spark: 3.1.1
-            scala: 2.12.8
-            python: 3.7
-          - spark: 3.1.1
-            scala: 2.12.8
-            python: 3.8
-          - spark: 3.1.1
+          - spark: 3.2.0
             scala: 2.12.8
             python: 3.9
-          - spark: 3.0.2
+          - spark: 3.2.0
+            scala: 2.12.8
+            python: 3.8
+          - spark: 3.2.0
             scala: 2.12.8
             python: 3.7
-          - spark: 2.4.7
+          - spark: 3.1.2
+            scala: 2.12.8
+            python: 3.7
+          - spark: 3.0.3
+            scala: 2.12.8
+            python: 3.7
+          - spark: 2.4.8
             scala: 2.11.8
             python: 3.7
 
--- a/.github/workflows/r.yml
+++ b/.github/workflows/r.yml
@@ -37,10 +37,10 @@ jobs:
           - spark: 3.0.3
             scala: 2.11.8
             r: release
-          - spark: 2.4.7
+          - spark: 2.4.8
             scala: 2.12.8
             r: oldrel
-          - spark: 2.4.7
+          - spark: 2.4.8
             scala: 2.12.8
             r: release
     env:
--- a/.github/workflows/r.yml
+++ b/.github/workflows/r.yml
@@ -15,20 +15,26 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        spark: [2.4.7, 3.0.2, 3.1.1]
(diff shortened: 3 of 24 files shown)


jiayuasu:
@Kimahriman 

Thanks for your update. This looks good to me.

Do you mind adding me @jiayuasu  and Pawel @Imbruced to be collaborators of your Sedona fork? So I can update your Github Action config and Pawel can fix Python part.

@Imbruced BTW, I cannot add you as a reviewer of Sedona PR because you don't have the "write" access to this repo. Please fix your ASF account.

Kimahriman:
Sent out an email, I'll update this for now to fix

Kimahriman:
Python tests failed because of https://github.com/apache/incubator-sedona/blob/6fafc93217b14745d11819f44e5a166480746baa/python/tests/serialization/test_deserializers.py#L68 which does a geomfromwkt call with no parameters and doesn't make any assertions, is there any point in that? It never gets evaled which is why the assertion didn't get thrown before

jiayuasu:
Scala/Java test passed but Python test failed. This is possibly caused by the Spark 3.2.0 change in UnaryExecNode and BinaryExecNode. So technically, Sedona join extends a different ExecNode in Spark 3.2.0.

Spark < 3.2.0 UnaryExecNode and BinaryExecNode implementation:

image

image

Spark 3.2.0 UnaryExecNode and BinaryExecNode implementation

image

image

This is related to https://github.com/apache/incubator-sedona/pull/558

However, on Sedona's side, this does not require any actual code change, so Scala/Java test passed. But I don't understand why Python test cannot figure this out.

@Kimahriman @Imbruced Can we figure out a solution for this? In the worst case, we have to cut a separate release that is compiled on Spark 3.2+. In short, we have to release Sedona for Spark 2.4, 3.0 (3.0-3.1), 3.2. This is a headache...

Kimahriman:
Hm didn't realize that would cause an issue. I think I saw another project create their own version of UnaryExecNode/BinaryExecNode to get around this, so that might be an option?

On a separate but related note, could consider dropping 2.4 support at some point since it's EOL to ease some of the burden

Kimahriman:
Python tests passed 🎉 

Don't really know about the R build though. Looks like sparklyr maybe has a hard-coded list of available spark versions so it just can't find the new version to download and use?

Also had to not include a few explain improvements added to Unary/BinaryExecNode because they weren't available in Spark 2.4. Could try to add it with the anchors but it involved some multiline strings so would have to rewrite that and add anchors to every line hah.

jiayuasu:
@Kimahriman This is a brilliant idea. Let use this alternative for a few releases. When Spark 3.1 reaches EOL, then we can use Spark's own ExecNode. We may need to periodically pull the latest ExecNode changes from Spark in the future.

@yitao-li Hi Yitao, could you please check how to set the Spark 3.2 support in our GitHub action? Please ping @Kimahriman for the write access in his fork if needed.

Imbruced:
@jiayuasu We can stop supporting spark 2.4, but I am worrying that many users still are using spark 2.4.x.

jiayuasu:
Let's wait for Yitao's response since he is the main contributor of Sedona R

Kimahriman:
This is an attempt at fixing the bug with registering the temp views in SQL: https://github.com/apache/spark/pull/34473. I haven't tested it out yet but I'll confirm if it indeed fixes the issue with the viz test. If so, it'll just be a known-issue with a workaround until 3.2.1 is released.

Kimahriman:
Confirmed that does fix the SQL create view behavior failing the test

yitao-li:
@Kimahriman @jiayuasu I made the necessary changes in `sparklyr`. You can run the R builds again and everything should work.

Kimahriman:
Is it pulling something from master branch in github versus only the latest released sparklyr version? Looks like 3.0.3 is still missing

yitao-li:
@Kimahriman At the moment we are using `sparklyr` from the master branch. I'll add URLs for Spark 3.0.3 asap.

yitao-li:
@Kimahriman Spark 3.0.3 should be working now too.

jiayuasu:
@yitao-li It looks like Spark 2.4.8 is not added to Sparklyr yet. Can you add it too?

yitao-li:
@jiayuasu I pushed some more extensive changes to `sparklyr` in https://github.com/sparklyr/sparklyr/pull/3198. `spark_install()` should be flexible enough to handle new/unknown versions of Apache Spark now.

jiayuasu:
@yitao-li R tests on Spark 3.2.0 failed due to the Spark temp view bug https://github.com/apache/spark/pull/34473

Apparently, it will take some time for Spark to release 3.2.1 so we have to bypass this issue in our test cases. We have bypassed this issue in Scala code: https://github.com/apache/incubator-sedona/pull/557/files#diff-d2c003f25373535c8866fbbbeba085f66985ff2a9f417538d7b60e090839d1bf. Could you please fix it in Sedona R?

I can accept this PR and then you can create a new PR on Sedona to fix Sedona R. What do you think?

Kimahriman:
Should we just take the 3.2 R tests out of the github workflow and add them back in in the follow up?

jiayuasu:
Spark 3.2 R tests have been removed. @yitao-li Please create a Sedona PR to fix Sedona R tests when you have time. Thank you again for your help!

yitao-li:
@jiayuasu OK. I re-enabled R builds for Spark 3.2 in https://github.com/apache/incubator-sedona/pull/561.

arturdryomov:
@jiayuasu, is there a rough timeline when the Spark 3.2 support might be shipped?

jiayuasu:
@arturdryomov Good question. I will start to prepare Sedona 1.1.1 release this weekend.

PR #559, opened 2021-11-02
[SEDONA-72 ] Fix showcase execution bug

Zeratal:

https://issues.apache.org/jira/browse/SEDONA-72

Fixed the bug when   ScalaExample in showcase run
## How was this patch tested?
Run the ScalaExample  in showcase
no

### DIFF
--- a/core/src/main/scala/org/apache/sedona/core/showcase/ScalaExample.scala
+++ b/core/src/main/scala/org/apache/sedona/core/showcase/ScalaExample.scala
@@ -56,7 +56,7 @@ object ScalaExample extends App {
   val PolygonRDDSplitter = FileDataSplitter.CSV
   val PolygonRDDNumPartitions = 5
   val PolygonRDDStartOffset = 0
-  val PolygonRDDEndOffset = 9// 8      modified by wjp 2021.11.2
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/sedona/core/showcase/ScalaExample.scala
```diff
@@ -56,7 +56,7 @@ object ScalaExample extends App {
   val PolygonRDDSplitter = FileDataSplitter.CSV
   val PolygonRDDNumPartitions = 5
   val PolygonRDDStartOffset = 0
-  val PolygonRDDEndOffset = 8
+  val PolygonRDDEndOffset = 9// 8      modified by wjp 2021.11.2
```
jiayuasu, line 59: Please remove the comment in this line

jiayuasu (changes requested):

core/src/main/scala/org/apache/sedona/core/showcase/ScalaExample.scala
Zeratal: done

jiayuasu (approved):

PR #562, opened 2021-11-11
[SEDONA-73] Add support for Scala 2.13 build

Kimahriman:
https://issues.apache.org/jira/browse/SEDONA-73

Adds a Scala 2.13 Maven profile and updates the code base to support 2.13 in addition to existing versions. A new dependency was added that provides backward compatibility for various Java-Scala conversion utilities. I enabled the Java tests to also run with 2.13.

It will be a little trickier to figure out how to run the Python or R tests with Scala 2.13, because I think both just download official releases to do it. It looks like there are official releases for Scala 2.13, but I am not trying to tackle that in this PR. The Python tests would likely have to be updated to download the official release versus installing from Pypi, because that only comes with the default Scala version. This is mostly for experimental purposes, and for making any new additions be Scala 2.13 compatible to make the release in the future easier. 

I also did a little refactoring on the raster functions that I missed last time. They were double eval'ing everything to check the type. I updated to just use the common parent type instead to convert to a double array.

## How was this patch tested?
Existing tests, added 2.13 build to java GitHub workflow tests.

Internal update right now, can update documentation when creating official 2.13 release support.

### DIFF
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -23,6 +23,9 @@ jobs:
             scala: 2.11.8
           - spark: 3.0.3
             scala: 2.11.8
+        include:
+          - spark: 3.2.0
+            scala: 2.13.5
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/r.yml
+++ b/.github/workflows/r.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: true
       matrix:
         spark: [2.4.8, 3.0.3, 3.1.2, 3.2.0]
-        scala: [2.11.8, 2.12.8]
+        scala: [2.11.8, 2.12.15]
         r: [oldrel, release]
         exclude:
           - spark: 3.2.0
@@ -38,10 +38,10 @@ jobs:
             scala: 2.11.8
             r: release
           - spark: 2.4.8
-            scala: 2.12.8
+            scala: 2.12.15
             r: oldrel
           - spark: 2.4.8
-            scala: 2.12.8
+            scala: 2.12.15
             r: release
     env:
       SPARK_VERSION: ${{ matrix.spark }}
--- a/pom.xml
+++ b/pom.xml
@@ -218,6 +218,11 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.scala-lang.modules</groupId>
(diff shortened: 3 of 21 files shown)


.github/workflows/java.yml
```diff
@@ -23,6 +23,9 @@ jobs:
             scala: 2.11.8
           - spark: 3.0.3
             scala: 2.11.8
+        include:
```
jiayuasu, line 26: You need to enable Scala 2.13 in this yml when compile Sedona so the Scala/Java test can actually run. There are some examples in the yml file.

jiayuasu (changes requested):

.github/workflows/java.yml
Kimahriman: Doesn't 
```
run: if [ ${SPARK_VERSION:0:1} == "3" ]; then mvn -q clean install -Dscala=${SCALA_VERSION:0:4} -Dspark=3.0 ; else mvn -q clean install -Dscala=${SCALA_VERSION:0:4} -Dspark=2.4 ; fi
```
already handle that? It looks like that test did run with scala 2.13 in the actions (I see deprecation warnings). Also, since the patch version is ignored in the tests should the scala version just be set to the minor version in the matrix?

jiayuasu (approved):

.github/workflows/java.yml
jiayuasu: Sorry, my bad. Forgot this.

jiayuasu:
@Kimahriman Hi Adam, I received reports from several users.

This PR introduces a dependency [Scala Collection Compat](https://mvnrepository.com/artifact/org.scala-lang.modules/scala-collection-compat) in the main POM.xml. 

The entire [scala-library](https://mvnrepository.com/artifact/org.scala-lang/scala-library) is the compile dependency of  Scala Collection Compat. As a result, Scala library is packaged in all Sedona releases.

This leads to (1) big Sedona jars (2) potential version conflict. Can we exclude Scala library from Scala Collection Compat?

Kimahriman:
Good catch, made #577

achinaou:
Hello!

I just wanted to ask how far away is the Scala 2.13 release and if it will be possible to use it with Scala 3 provided that [there is a way](https://www.47deg.com/blog/using-scala-3-with-spark/) of using Apache Spark with Scala 3.

PR #565, opened 2021-11-29
[SEDONA-75] Add support for "3D" geometries

kseebaldt:

[SEDONA-75](https://issues.apache.org/jira/browse/SEDONA-75)

Preserve Z coordinates on geometries when serializing.
Enhance ST_AsText to output Z coordinate if it exists.  
Add ST_Z function to access Z coordinate
Add ST_3DDistance to compute 3D distance between geometries

## How was this patch tested?

Added unit tests.  Used in internal projects with 2D & 3D datasets

Yes


Imbruced:
@kseebaldt Hi, Thank you for the PR (it's awesome). Can you also add tests for Python ?

kseebaldt:
Sure!  I'll work on that.

jiayuasu:
@Imbruced This looks good to me. What do you think?

Imbruced:
To me it looks good but if you dont mind I will check this branch locally tomorrow.

Imbruced:
One of our action is failing due to depricated exception we should fix that in first place :) I can do that

jiayuasu:
@Imbruced Sure. I will wait for your response tomorrow.

BTW, the failed GitHub action of Example project has been fixed by me. @kseebaldt On your local branch, can you click "fetch upstream updates" on your GitHub interface? This will load my latest commit.

jiayuasu:
@kseebaldt @Imbruced I just realized 3D geometries will not be considered in SpatialJoin. For example, this SQL query won't trigger a Sedona optimized join

```
SELECT t1.geom, t2.geom
FROM t1, t2
WHERE ST_3dDistance(t1.geom, t2.geom)
```

Anyway, I think it is good to accept this PR as is, and have another PR to add the support of 3D spatial join.

Imbruced:
But in case of 3D spatial join we should include tree structures like octree. That sounds interesting :)

jiayuasu:
@Imbruced I would say it is yes and no.

"No": The existing code structure and index structure will work for 3D geoms, and guarantee the correctness of the result.
"Yes": the existing code such as spatial partitioning and indexes does not filter upon the z axis so the join performance is not optimal.

BTW, will you be OK if I accept this PR?

Imbruced:
@jiayuasu I am finishing code review, should be finished within an hour. 
About 3D join, I know it will work as it is now but as you said wont be optimal, bcs we drop one dimension. It would be interesting to add 3D spatial partitioning and indexing.

python/tests/sql/test_constructor_test.py
```diff
@@ -56,6 +60,24 @@ def test_st_geom_from_wkt(self):
...
+            ("Linestring(0 0 1, 1 1 2, 1 0 3)",),
+            ("MULTIPOINT ((10 40 66), (40 30 77), (20 20 88), (30 10 99))",),
+            ("MULTIPOLYGON (((30 20 11, 45 40 11, 10 40 11, 30 20 11)), ((15 5 11, 40 10 11, 10 20 11, 5 10 11, 15 5 11)))",),
+            ("MULTILINESTRING ((10 10 11, 20 20 11, 10 40 11), (40 40 11, 30 30 11, 40 20 11, 30 10 11))",),
+            ("MULTIPOLYGON (((40 40 11, 20 45 11, 45 30 11, 40 40 11)), ((20 35 11, 10 30 11, 10 10 11, 30 5 11, 45 20 11, 20 35 11), (30 20 11, 20 15 11, 20 25 11, 30 20 11)))",),
+            ("POLYGON((0 0 11, 0 5 11, 5 5 11, 5 0 11, 0 0 11), (1 1 11, 2 1 11, 2 2 11, 1 2 11, 1 1 11))",),
+        ], ["wkt"])
+
+        input_df.createOrReplaceTempView("input_wkt")
+        input_df.show()
```
Imbruced, line 76: Can we avoid showing data frame during tests ? I dont know if github actions have simmilar issue to travis, but tests failed due 2 a lot of prints in console.

Imbruced (changes requested):

Imbruced:
@jiayuasu It looks good to me, but we should avoid showing dataframes within the tests or any unnecessary print statements.

python/tests/sql/test_constructor_test.py
kseebaldt: No problem.  I was just following the pattern of the existing tests.

Imbruced (approved):
Awesome, thank you ! :)

jiayuasu (approved):

PR #569, opened 2021-12-09
[SEDONA-18] error reading shapefile

madzik555:
https://issues.apache.org/jira/browse/SEDONA-18

In function getSplits() (core/src/main/java/org/apache/sedona/core/formatMapper/shapefileParser/shapes/ShapeInputFormat.java), the condition that splits file names by name only has been changed. Before the change, files with a different name and a different extension were treated as a separate data package, in which specific files could not be found.

## How was this patch tested?
New java unit test.

No


JassonHua:
您好，您的邮件我已收到，我会尽快给您回复。谢谢!

jiayuasu (changes requested):
Thank you for your contribution! Can you use the existing shapefile (with an additional file inside that has a different file extension) in the resources and use the existing test cases to validate the new patch, instead of introducing a new shapefile?

There are two reasons behind this:
1. A new shapefile introduces to additional copyright issue
2. This new shapefile increases the source code size.

madzik555:
No problem. I can do this

jiayuasu (approved):

PR #575, opened 2022-01-30
[SEDONA-80] Geospatial stream processing support in Flink Table API

jiayuasu:

https://issues.apache.org/jira/browse/SEDONA-80

Add the basic support for Apache Flink

This PR is compiled against Flink Table API 14.3, but it has been tested on Flink Table API 12 and 13 (both with Blink planner) manually.

We might need to add the CI for all recent Flink versions including 12, 13, and 14

## How was this patch tested?

Added a number of unit tests

Will be added as a separate PR


jiayuasu:
@Imbruced This is the Sedona Flink support. I just pushed several basic operators and tried to re-use some functions in Sedona-core. I believe this will be a good start so others who are interested in Flink can contribute more funcs to this.

A doc PR will be published soon. But to give you an overview, this is the functions I added: https://github.com/apache/incubator-sedona/pull/575/files#diff-bcb3952f695c0b65fc27f9e7f30d5d365950823c326504fe0d55fa933bf81471

Imbruced:
@jiayuasu What is better approach, create new package for python flink ? or add the code to exisitng one to python/sedona but create new sub module ?

jiayuasu:
@Imbruced Is there any Sedona python code can be shared between Sedona Spark and SedonaFlink? E.g., the shapely binding of geometry objects?

Then for Python, we probably could have 1. apache-sedona-core, then 2. apache-sedona-spark and 3. apache-sedona-flink.

2 and 3 will use 1 as a required library.

Imbruced:
@jiayuasu For geometry objects at the moment logic is simple (wkb.loads). When I have some working example will let you know about possibilites and then we will decide. WDYT ?

jiayuasu:
@Imbruced Sounds good to me! I will merge the Java API first.

PR #580, opened 2022-02-14
[SEDONA-82] Create symmetrical difference function

madzik555:
[SEDONA-82](https://issues.apache.org/jira/browse/SEDONA-82)

ST_Symmetrical_Difference function

## How was this patch tested?
Scala, Python integration tests

Yes


jiayuasu:
@madzik555 I have fixed the failed docs build. Please pull the latest change from upstream.

madzik555:
@jiayuasu Could you take a look at these errors in Github Checks?

jiayuasu:
@yitao-li Hi Yitao, it seems that R build failed because it cannot download Spark. Could you please fix it?

Imbruced:
@madzik555 Can you add also tests with null values ?

jiayuasu:
@madzik555 Yitao might be busy at other stuff. But apparently your PR passed all other tests. You can just fix what @Imbruced suggested and then I will accept this PR regardless of the R tests.

madzik555:
@Imbruced @jiayuasu I made fixes in ST_SymmetricalDifference to handle null values. Can you check them? :)
Additionally, I would like to introduce the same changes in ST_Difference, should I create a new PR?

yitao-li:
Hey @jiayuasu @madzik555 sorry for the delayed response re: sparklyr

I think the problem is https://dlcdn.apache.org/spark/spark-3.2.0/spark-3.2.0-bin-hadoop3.2.tgz used to exist, and in sparklyr, we essentially made an implicit assumption that this type of tar ball, once existed, will exist forever and be accessible from the same location. In particular, see https://github.com/sparklyr/sparklyr/blob/1c091b92d7454fa8af02abe9777efba8fb6c4faa/inst/extdata/versions.json#L497

yitao-li:
@edgararuiz so continuing with the comment above, I think if you go to https://github.com/sparklyr/sparklyr/blob/main/inst/extdata/versions.json and replace the URL mentioning spark 3.2.0 with https://dlcdn.apache.org/spark/spark-3.2.1/<the tar ball filename>, then it should fix this problem.

@madzik555 @jiayuasu Meanwhile I think you need to switch to Spark 3.2.1 instead. I'm really not sure what happened to Spark 3.2.0. It seems not accessible anymore : /

jiayuasu:
@madzik555 This looks great! One last thing: it seems that this function is identical to ST_SymDifference in PostGIS: https://postgis.net/docs/ST_SymDifference.html , right? If so, can you change the name from "ST_Symmetrical_Difference" to "ST_SymDifference"?

We want to be consistent with PostGIS/SQLMM3/OGC Simple Feature.

Imbruced:
It looks good to me, also I agree that we should stick with postgis function name. It may be confusing.

madzik555:
I changed the name of the function. On Monday I will introduce corrections to ST_Difference and create a PR for it. Thanks for your comment! 😄

jiayuasu (approved):

PR #584, opened 2022-02-21
[SEDONA-82] Fixes in ST_Difference and ST_SymDifference

madzik555:

https://issues.apache.org/jira/browse/SEDONA-82

Fixes in ST_Difference, ST_SymDifference to handle null values.

## How was this patch tested?
New Scala unit tests added 

No

### DIFF
--- a/sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala
+++ b/sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala
@@ -1444,26 +1444,37 @@ case class ST_GeoHash(inputExpressions: Seq[Expression])
   }
 }
 
+/**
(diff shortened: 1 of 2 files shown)


jiayuasu:
@madzik555 Wait, I think we had a miscommunication here. Your implementation for null safe ST_Difference was correct. The expression should extend "BinaryGeometryExpression" which is automatically null safe.

What @Imbruced asked is that just add some "null" test cases in https://github.com/apache/incubator-sedona/pull/580

So for ST_SymDifference, you should also extend "BinaryGeometryExpression". I believe your old implementation was correct, but the one we merged into the master branch didn't extend "BinaryGeometryExpression"...

madzik555:
Okay, my bad. I fix it tomorrow. 😃

jiayuasu (approved):

PR #593, opened 2022-03-20
[SEDONA-93] Add ST_GeomFromGeoJSON to flink API

netanel246:
[SEDONA-93](https://issues.apache.org/jira/browse/SEDONA-93)
Added ST_GeomFromGeoJSON to flink API
## How was this patch tested?
Added test
Yep

### DIFF
--- a/docs/api/flink/Constructor.md
+++ b/docs/api/flink/Constructor.md
@@ -27,6 +27,20 @@ SELECT ST_GeomFromWKB(polygontable._c0) AS polygonshape
 FROM polygontable
 ```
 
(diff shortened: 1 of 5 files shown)


jiayuasu (approved):
Looks great!

jiayuasu:
@netanel246 Thanks for your contribution. This looks good. 

Some of the functions in Sedona Spark SQL are from JTS and some are standalone (without Spark dependency). So I think we can just call those functions in Sedona Flink SQL. No need to create a separate Maven module for it at this moment but maybe in the future.

PR #594, opened 2022-03-22
[DOCS] Fix copyright issues

jiayuasu:

Update the NOTICE file

Add ASF header in shell scripts

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,5 +1,12 @@
 Apache Sedona (incubating)
-Copyright 2015-2021 The Apache Software Foundation
+Copyright 2020 and onwards The Apache Software Foundation
 
 This product includes software developed at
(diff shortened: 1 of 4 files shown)


NOTICE
```diff
@@ -1,5 +1,12 @@
 Apache Sedona (incubating)
-Copyright 2015-2021 The Apache Software Foundation
+Copyright 2020 and onwards The Apache Software Foundation
```
vongosling, line 2: 2015~ 2022 may be better?

vongosling, line 2: 2020 ~ 2022 maybe better?

jiayuasu: I am not sure... I copied the idea from Apache Spark: https://github.com/apache/spark/blob/master/NOTICE#L2

PR #596, opened 2022-03-31
[SEDONA-96] Refactor ST_MakeValid to use GeometryFixer.

umartin:
I will always return a single geometry and works for all geometry types.

[SEDONA-96](https://issues.apache.org/jira/browse/SEDONA-96)

Refactor ST_MakeValid to use GeometryFixer in JTS

## How was this patch tested?

Unit tests added

Yes

### DIFF
--- a/docs/api/sql/Function.md
+++ b/docs/api/sql/Function.md
@@ -162,27 +162,40 @@ FROM polygondf
 
 ## ST_MakeValid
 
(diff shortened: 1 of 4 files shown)


jiayuasu:
@netanel246 The current MakeValid function was written by yo (with a generator). Do you have any suggestion with this new MakeValid function?

jiayuasu (approved):

PR #607, opened 2022-04-15
GeoParquet Reader Implemented (SEDONA-94)

ashar236:
SEDONA-94 
GeoParquet reader support was added.
## How was this patch tested?
The unit test for three different parquet file with geometry column have been added.
Yes.


MulesNose:
Great stuff, found your question from Opengeospatial and almost linked you to your own commit!

cholmes:
Is there any progress on this? I'd love to add Sedona to the list at https://github.com/opengeospatial/geoparquet#current-implementations--examples

JassonHua:
您好，您的邮件我已收到，我会尽快给您回复。谢谢!

jiayuasu:
> Is there any progress on this? I'd love to add Sedona to the list at [opengeospatial/geoparquet#current-implementations--examples](https://github.com/opengeospatial/geoparquet#current-implementations--examples)

We are still working on this. The PR will be updated soon. Will keep you posted

sql/src/main/scala/org/apache/spark/sql/sedona_sql/io/geoparquet/ParquetFileFormat.scala
```diff
@@ -0,0 +1,545 @@
+/*
```
ericsun95, line 1: Looks like all these are duplicated with Spark internal parquet reader implementation. From my understanding the geoparquet is just one specific type of parquet with its only schema. 
Generally, what differs here? how about implementing a very lean layer on top of spark native parquet file reader using DataSourceV2 interface with specific schema?

neontty: @jiayuasu if I start working on geoparquet predicate pushdown, will it be invalidated if you are moving to the datasource V2 api soon? Is there a branch with the v2 api ("[We have a neat version that nicely supports Spark 3.3 only](https://github.com/apache/incubator-sedona/pull/652)")?

jiayuasu: @neontty My understanding is that the merged PR in Sedona already uses Datasource v2 as Spark Parquet reader is using it. But Spark parquet reader is designed in a way that most critical functions are private, so we have to copy lots of their code for adding a few changes on our own.

The version that supports Spark 3.3 only is here: https://github.com/ashar236/incubator-sedona/tree/geoparquet-spark3.3/sql/src/main/scala/org/apache/spark/sql/execution/datasources/parquet

In short, I believe your geoparquet predicate pushdown function will Not be invalidated. But I would suggest that please minimize your change to the codebase.

jiayuasu:
> Is there any progress on this? I'd love to add Sedona to the list at [opengeospatial/geoparquet#current-implementations--examples](https://github.com/opengeospatial/geoparquet#current-implementations--examples)

@cholmes Dear Chris, we have released Sedona 1.3.0 which includes reader/writer for GeoParquet. I created [a PR](https://github.com/opengeospatial/geoparquet/pull/150) on geoparquet repo as well. Please review it when you have time :-)

PR #609, opened 2022-04-15
Feature/dbscan

krishnarb3:
[SEDONA-46](https://issues.apache.org/jira/browse/SEDONA-46)

ST_ClusterDBSCAN

## How was this patch tested?
Added unit tests for DBScan and UnionFind

Waiting for confirmation of the approach


core/src/main/java/org/apache/sedona/core/dbscanJudgement/UnionFindImpl.java
```diff
@@ -0,0 +1,134 @@
...
+        Integer[] orderedComponents = orderedByCluster();
+        Integer[] newIds = new Integer[n];
+        int lastOldId = 0, currentNewId = 0;
+        boolean encounteredCluster = false;
+
+        if (isInCluster == null) {
+            isInCluster = new HashSet<>();
+        }
+
+        for (int i = 0; i < n; i++) {
```
krishnarb3, line 89: As per POSTGIS `getCollapsedClusterIds` will set the cluster id only for indexes that are part of isInCluster or all the indexes if isInCluster is empty.
@jiayuasu Is this the expected behavior?

core/src/main/java/org/apache/sedona/core/dbscanJudgement/DBScanJudgement.java
```diff
@@ -0,0 +1,180 @@
...
+            int numNeighbors = 0;
+            List<Integer> geomsInEnvelope = getGeomsInEnvelope(geoms, i, eps, strtree);
+            if (geomsInEnvelope.size() < minPoints) {
+                continue;
+            }
+
+            for (Integer j : geomsInEnvelope) {
+                if (numNeighbors >= minPoints) {
+                    /*
+                     * If we've already identified p as a core point, and it's already
```
krishnarb3, line 102: @jiayuasu Have added the comments from POSTGIS code for clarity

core/src/main/java/org/apache/sedona/core/spatialOperator/DBScanQuery.java
```diff
@@ -0,0 +1,31 @@
...
+public class DBScanQuery
+        implements Serializable
+{
+    public static <T extends Geometry> List<Integer> SpatialDBScanQuery(SpatialRDD<T> spatialRDD, double eps, int minPoints, boolean useIndex)
+    {
+        if (useIndex) {
+            if (spatialRDD.indexedRawRDD == null) {
+                throw new NullPointerException("Need to invoke buildIndex() first, indexedRDDNoId is null");
+            }
+            JavaRDD<Integer> result = spatialRDD.getRawSpatialRDD().repartition(1).mapPartitions(new DBScanJudgement(eps, minPoints, new HashSet<>()), true);
```
krishnarb3, line 23: Since the implementation of DBScan is not parallel/distributed, it does not work when there are more than one partition(s).
@jiayuasu Do you think this is acceptable?

I will try to come up with an implementation that can be run on multiple partitions later in that case.

krishnarb3:
This follows the POSTGIS implementation of ST_DBScan.
I have not added this as a ST_Function since that would not be able to support running DBScan on multiple partitions in the future.

@jiayuasu What are your thoughts on this?

jiayuasu:
Thanks for your contribution. This PR is huge. I will take some to review it.

core/src/main/java/org/apache/sedona/core/spatialOperator/DBScanQuery.java
jiayuasu: If your implementation cannot work on multiple partitions, then this PR will NOT be accepted. We are looking for a distributed DBScan algorithm. For example, this one: https://github.com/irvingc/dbscan-on-spark

You can incorporate this library into Sedona since this one is also under Apache 2.0 License

jiayuasu (changes requested):

core/src/main/java/org/apache/sedona/core/spatialOperator/DBScanQuery.java
krishnarb3: Thank you @jiayuasu, I will work on this next week

krishnarb3:
I'll pick this PR back up this week

PR #612, opened 2022-04-18
[SEDONA-108] Write Support for GeoTiff Raster Images

kanchanchy:
Yes. [SEDONA-108](https://issues.apache.org/jira/browse/SEDONA-108)

GeoTiff DataFrame can be written as GeoTiff images.

## How was this patch tested?
Yes. Unit tested with various combination of test cases.

Yes. Documentation has been updated.


jiayuasu:
@kanchanchy This PR looks awesome. I have a small request regarding the CRS in GeoTIFF reader and writer. Currently, it is fixed to (1) in the reader, read from the original data files and read to EPSG 4326 (2) in the writer, write to EPSG 4326. This creates some troubles to the end user. Can you provide some optional parameters to the user such as

This will write data to the desired the CRS
```
sparkSession.read.format("geotiff").option("writeFromCRS", "epsg:XXX").save("YOUR_PATH")
sparkSession.read.format("geotiff").option("writeToCRS", "epsg:XXX").save("YOUR_PATH")
```

If the original GeoTiff data files does not have CRS inside, then use the following 
```
sparkSession.read.format("geotiff").option("readFromCRS", "epsg:XXX").load("YOUR_PATH")
sparkSession.read.format("geotiff").option("readToCRS", "epsg:XXX").load("YOUR_PATH")
```

If the original GeoTiff data files have CRS inside, `readFromCRS` will not be in effect.

Both of reader and writer should have an optional parameter which is option("disableErrorInCRS", "true"). This will be the parameter used here: https://github.com/apache/incubator-sedona/blob/master/sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala#L288

The default value should be false

kanchanchy:
Hi @jiayuasu I didn't understand the writeToCRS option. If end user provides writeToCRS option during, the coordinates of the GeoTiff source should be in that CRS format, right?

With this: 'It just writes CRS name together with GeoTiff.', do you mean providing an extra meta-data field for the CRS of the GeoTiff?

sql/src/main/scala/org/apache/spark/sql/sedona_sql/io/GeotiffFileFormat.scala
```diff
@@ -101,4 +126,123 @@ private[spark] class GeotiffFileFormat extends FileFormat with DataSourceRegiste
...
+      for (j <- 0 until tiffWidth) {
+        for (k <- 0 until tiffBands) {
+          pixelVal(k) = tiffData(tiffHeight*tiffWidth*k + i * tiffWidth + j)
+        }
+        raster.setPixel(j, i, pixelVal)
+      }
+    }
+
+    // CRS is always fixed to EPSG:4326
+    val crs = CRS.decode("EPSG:4326", true)
```
jiayuasu, line 202: This should NOT be always fixed to EPSG:4326. The user will use ST_Transform in their DF to convert coordinates to whatever CRS they like. Then call df.write.save to save GeoTiff. GeoTiff meta data has a place for storing the name of the CRS.

kanchanchy: Got it. I have one more question. If the user doesn't specify the 'readToCRS' option during reading, what would be the target default value? In that case, should we use CRS available in the GeoTiff or readFromCRS as the target value?

jiayuasu: @kanchanchy 

In my opinion,

For reader, both "ReadFromCRS" and "ReadToCRS" are optional. We use The following combinations:

1. No "ReadFromCRS" and "ReadToCRS" are provided: we don't perform CRS transformation even if the raw data has CRS info. We just drop the CRS info.
2. "ReadFromCRS" NOT provided, "ReadToCRS" are provided: we use the CRS available in the GeoTiff as source, use "ReadToCRS" as the target. If the GeoTiff does not have CRS info, we throw an exception.
3. "ReadFromCRS" are provided, "ReadToCRS" are  NOT provided: we don't perform CRS transformation even if the raw data has CRS info. We just drop the CRS info.
4. Both "ReadFromCRS" and "ReadToCRS" are provided: we use the "ReadFromCRS" and "ReadToCRS" to perform the transformation, ignoring the CRS in the GeoTiff.

For writer, "writeToCRS" is optional.

1. "writeToCRS" is provided: store CRS info in GeoTiff metadata. No CRS transformation is needed.
2. "writeToCRS" is NOT provided: don't store CRS info in GeoTiff metadata, or store WGS84 (aka epsg:4326) in the metadata. No CRS transformation is needed.

kanchanchy: @jiayuasu When writing a GeoTiff, we usually need to pass the CRS as a parameter when creating the reference envelop. It indicates the CRS of the polygon representing the image source. So, if we need to pass the correct CRS info in line 208 [here](https://github.com/apache/incubator-sedona/pull/612/files/6fe92ec5c2736f8cbd6d9dd381380a1f5ac6aab8#diff-9afdfa8529cd2458278c7f0a2abb09bfaaa0599fe312b02ff6085d72c49fdda2).

kanchanchy: The CRS that we pass there, it should be the correct CRS representing the polygon of image source. So, if users transform the polygon using ST_Transform, they must need to provide the transformed CRS as writeToCRS option. Also, if we use a default CRS in the absence of writeToCRS, that default CRS should be actual CRS of the polygon.

kanchanchy: Is it possible to find the CRS from the coordinates or geometry object in java or scala programming?

jiayuasu: CRS information might be stored in Geometry SRID attribute but this is NOT mandatory. So we cannot find the CRS by ourselves. It has to come from the user.

So let me tweak my proposal for the "writer" a bit:


For writer, "writeToCRS" is optional. Making it optional is kind of important as some GeoTiff users don't know what is called CRS.

"writeToCRS" is provided: store CRS info in GeoTiff metadata. No CRS transformation is needed.
"writeToCRS" is NOT provided: store WGS84 (aka epsg:4326) in the metadata. No CRS transformation is needed.

kanchanchy:
@jiayuasu I finished implementing exactly what you proposed.

However, I need to fix one more issue. If the geometry column is not passed to ST_Geom_From_WKT or ST_Transform then the geometry column represents a String version of Polygon Object. I can convert it easily back to Polygon with WKTReader. On the other hand, if the geometry is transformed using SQL expressions such as ST_Geom_From_WKT or ST_Transform before calling df.write, then the geometry column represents a GeometryUDT. I am not able to retrieve the geometry/polygon from the GeometryUDT. Could you please give some insight on this?

sql/src/main/scala/org/apache/spark/sql/sedona_sql/io/GeotiffFileFormat.scala
```diff
@@ -101,4 +126,123 @@ private[spark] class GeotiffFileFormat extends FileFormat with DataSourceRegiste
...
+    if (fields.contains(colImage)) {
+      schemaFields = dataSchema.fields(dataSchema.fieldIndex(colImage)).dataType.asInstanceOf[StructType]
+      rowFields = row.getStruct(dataSchema.fieldIndex(colImage), 6)
+    }
+
+    val tiffOrigin = rowFields.getString(schemaFields.fieldIndex(colOrigin))
+    val tiffBands = rowFields.getInt(schemaFields.fieldIndex(colBands))
+    val tiffWidth = rowFields.getInt(schemaFields.fieldIndex(colWidth))
+    val tiffHeight = rowFields.getInt(schemaFields.fieldIndex(colHeight))
+    val tiffGeometry = rowFields.getString(schemaFields.fieldIndex(colWkt))
```
jiayuasu, line 184: Will you be able to do rowFields.getAs[Geometry]()?

kanchanchy:
@jiayuasu  org.apache.spark.sql.catalyst.expressions.UnsafeArrayData cannot be cast to org.locationtech.jts.geom.Geometry

sql/src/main/scala/org/apache/spark/sql/sedona_sql/io/GeotiffFileFormat.scala
kanchanchy: @jiayuasu org.apache.spark.sql.catalyst.expressions.UnsafeArrayData cannot be cast to org.locationtech.jts.geom.Geometry

kanchanchy: getAs() method is getting the filed value as UnsafeArraydata

kanchanchy: Solved. getAs[ArrayData] and then GeometrySerializer.deserialize()

jiayuasu: Yes, this is the solution

kanchanchy: @jiayuasu I pushed the new code. I added new unit tests for testing new cases added in this commit. Also, updated the documentation accordingly.

sql/src/main/scala/org/apache/spark/sql/sedona_sql/io/GeotiffFileFormat.scala
```diff
@@ -101,4 +129,127 @@ private[spark] class GeotiffFileFormat extends FileFormat with DataSourceRegiste
...
+  DEFAULT_WRITE_PARAMS.setCompressionType("LZW")
+  DEFAULT_WRITE_PARAMS.setCompressionQuality(0.75F)
+  DEFAULT_WRITE_PARAMS.setTilingMode(ImageWriteParam.MODE_EXPLICIT)
+  DEFAULT_WRITE_PARAMS.setTiling(512, 512)
+
+  private val hfs = new Path(savePath).getFileSystem(context.getConfiguration)
+
+  override def write(row: InternalRow): Unit = {
+    // retrieving the metadata of a geotiff image
+    println("Schema:")
```
jiayuasu, line 165: Don't use println to print log info. Use Log4j instead. See the example: https://github.com/apache/incubator-sedona/blob/master/sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala#L1004

Use "info" level in logging. To see the log in the test, set log level to info level in TestBase.

docs/api/sql/Raster-loader.md
```diff
@@ -142,3 +170,82 @@ Output:
...
+In case, you rename the columns of GeoTiff dataframe, you can set the corresponding column names with the `option` parameter. All available optional parameters are listed below:
+
+```html
+ |-- writeToCRS: (Default value "EPSG:4326") => Coordinate reference system of the geometry coordinates representing the location of the Geotiff.
+ |-- fieldImage: (Default value "image") => Indicates the image column of GeoTiff DataFrame.
+ |-- fieldOrigin: (Default value "origin") => Indicates the origin column of GeoTiff DataFrame.
+ |-- fieldNBands: (Default value "nBands") => Indicates the nBands column of GeoTiff DataFrame.
+ |-- fieldWidth: (Default value "width") => Indicates the width column of GeoTiff DataFrame.
+ |-- fieldHeight: (Default value "height") => Indicates the height column of GeoTiff DataFrame.
+ |-- fieldWkt: (Default value "wkt") => Indicates the wkt column of GeoTiff DataFrame.
```
jiayuasu, line 241: this should be fieldGeometry. Please fix it in the code as well.

jiayuasu (changes requested):
Almost done. Just 2 minor changes.

docs/api/sql/Raster-loader.md
kanchanchy: @jiayuasu I am changing the fieldWkt to fieldGeometry in the code and other places. Should I need to change the wkt column name to geometry in the GeoTiffSchema?

jiayuasu: yes, please change it to geometry.

jiayuasu:
@Imbruced Hi Pawel, your Python test case of "Geom Constructor on Stream" frequently fails because you use a random name to the temp view name. The name was actually not very random so it frequently creates conflicting view names and hence fail. Can you fix your Python tests on Spark Stream?

jiayuasu (approved):

jiayuasu:
@kanchanchy I am gonna merge this PR. Can you also update the notebook example here in another PR? https://github.com/apache/incubator-sedona/blob/master/binder/ApacheSedonaRaster.ipynb

PR #639, opened 2022-06-30
[SEDONA-39] Fix lon/lat order in ST_Transform to be consistent with PostGIS

jiayuasu:

- Yes, the URL of the assoicated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-39. The PR name follows the format `[SEDONA-39] my subject`.

- Yes, I am adding a new API. I am using the [current SNAPSHOT version number](https://github.com/apache/incubator-sedona/blob/master/pom.xml#L29) in since `vX.Y.Z` format.


netanel246:
Hi @jiayuasu ,
I think we should be aligned with PostGIS. 
However, we must be clear about staying with lat/lon order and **emphasize** that users need to update their code if they use this version.

Imbruced:
I agree with @netanel246 we should be consistent with what Postgis does. And definitely we need communicate that clearly with users.

jiayuasu:
@netanel246 @Imbruced Got it. I will then continue working on this PR.

PR #653, opened 2022-08-01
[SEDONA-135] Consolidate and upgrade Hadoop version

sekikn:

- Yes, the URL of the assoicated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-135. The PR name follows the format `[SEDONA-XXX] my subject`.

- This PR consolidates Hadoop versions and upgrade them to the one officially supported as of now.

## How was this patch tested?

- I ran the following commands and ensured all of them successfully finished.

```
$ mvn clean install # same as `mvn clean install -Dhadoop=2.10 -Dspark=3.0 -Dscala=2.12`
$ mvn clean install -Dhadoop=2.10 -Dspark=2.4 -Dscala=2.11
$ mvn clean install -Dhadoop=2.10 -Dspark=3.0 -Dscala=2.12
$ mvn clean install -Dhadoop=3.0  -Dspark=2.4 -Dscala=2.11
$ mvn clean install -Dhadoop=3.0  -Dspark=3.0 -Dscala=2.12
```

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -62,19 +62,7 @@ jobs:
     - env:
         SPARK_VERSION: ${{ matrix.spark }}
         SCALA_VERSION: ${{ matrix.scala }}
-      run: |
-        if [ ${SPARK_VERSION:0:1} == "3" ]; \
-        then \
-          if [ ${HADOOP_VERSION:0:1} == "3" ]; \
-            then mvn -q clean install -DskipTests -Dscala=${SCALA_VERSION:0:4} -Dspark=3.0 -Dhadoop=3.0  -Dgeotools ; \
-            else mvn -q clean install -DskipTests -Dscala=${SCALA_VERSION:0:4} -Dspark=3.0 -Dhadoop=2.10 -Dgeotools ; \
-          fi \
-        else \
-          if [ ${HADOOP_VERSION:0:1} == "3" ]; \
-            then mvn -q clean install -DskipTests -Dscala=${SCALA_VERSION:0:4} -Dspark=2.4 -Dhadoop=3.0  -Dgeotools ; \
-            else mvn -q clean install -DskipTests -Dscala=${SCALA_VERSION:0:4} -Dspark=2.4 -Dhadoop=2.10 -Dgeotools ; \
-          fi \
-        fi
+      run: if [ ${SPARK_VERSION:0:1} == "3" ]; then mvn -q clean install -DskipTests -Dscala=${SCALA_VERSION:0:4} -Dspark=3.0 -Dgeotools ; else mvn -q clean install -DskipTests -Dscala=${SCALA_VERSION:0:4} -Dspark=2.4 -Dgeotools ; fi
     - env:
         SPARK_VERSION: ${{ matrix.spark }}
         HADOOP_VERSION: ${{ matrix.hadoop }}
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -62,7 +62,19 @@ jobs:
     - env:
         SPARK_VERSION: ${{ matrix.spark }}
         SCALA_VERSION: ${{ matrix.scala }}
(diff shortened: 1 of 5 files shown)


sekikn:
Oh, I missed some workflows for GitHub Actions need to be updated. I'll revise the PR later.

core/pom.xml
```diff
@@ -41,9 +41,15 @@
...
             <artifactId>hadoop-minicluster</artifactId>
-            <version>2.7.4</version>
+            <version>${hadoop.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+          <groupId>org.mockito</groupId>
+          <artifactId>mockito-all</artifactId>
+          <version>1.8.5</version>
+          <scope>compile</scope>
```
Kimahriman, line 51: Why did this need to be added as an additional dependency? And if it is needed I assume this should be `test` scope

pom.xml
```diff
@@ -66,6 +66,12 @@
     </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>${guava.version}</version>
+            <scope>${dependency.scope}</scope>
```
Kimahriman, line 73: Why does guava need to be added, doesn't Hadoop have this already?

pom.xml
```diff
@@ -618,5 +612,33 @@
...
+                    <name>hadoop</name>
+                    <value>2.10</value>
+                </property>
+                <activeByDefault>true</activeByDefault>
+            </activation>
+            <properties>
+                <hadoop.version>2.10.2</hadoop.version>
+                <guava.version>11.0.2</guava.version>
+            </properties>
+        </profile>
```
Kimahriman, line 642: Do we need profiles for this? It looks like for the CI (except for python), Spark 3 uses Hadoop 3 and Spark 2 uses Hadoop 2, can we just set the hadoop version in those existing profiles?

.github/workflows/python.yml
```diff
@@ -62,7 +62,19 @@ jobs:
     - env:
         SPARK_VERSION: ${{ matrix.spark }}
         SCALA_VERSION: ${{ matrix.scala }}
-      run: if [ ${SPARK_VERSION:0:1} == "3" ]; then mvn -q clean install -DskipTests -Dscala=${SCALA_VERSION:0:4} -Dspark=3.0 -Dgeotools ; else mvn -q clean install -DskipTests -Dscala=${SCALA_VERSION:0:4} -Dspark=2.4 -Dgeotools ; fi
+      run: |
+        if [ ${SPARK_VERSION:0:1} == "3" ]; \
+        then \
+          if [ ${HADOOP_VERSION:0:1} == "3" ]; \
```
Kimahriman, line 68: I don't think `HADOOP_VERSION` is set to anything, these are populated in the `env` block above. And might not be necessary regarding the comment about the new profiles?

jiayuasu:
Out of curiosity, if we don't consolidate the Hadoop versions, what will happen? Currently, it seems that Sedona can compile on all Spark version since 2.4. And all Hadoop dependency are just provided or test so Sedona does not package Hadoop...

core/pom.xml
sekikn: > Why did this need to be added as an additional dependency?

Some versions of Hadoop 2.x (e.g., 2.10.2) doesn't have Mockito in the minicluster test-jar, and it leads to NoClassDefFoundError unless specifying this dependency explicitly. According to https://issues.apache.org/jira/browse/HADOOP-14857, it's supposed to be fixed in 3.x, but 3.2.4 still seems to have the same problem.

pom.xml
sekikn: > Why does guava need to be added, doesn't Hadoop have this already?

Yes it should do, but for some reason, wrong version of Guava is used for unit test and it causes NoSuchMethodError. Some other libraries that have a transitive dependency on Guava seem to cause this problem. For resolving it, I'll change the dependency declaration order in pom.xml, instead of adding this dependency.

sekikn:
> if we don't consolidate the Hadoop versions, what will happen?

As far as Hadoop API used by Sedona is not changed, nothing will happen. But if incompatible API change is introduced in the future, we may miss some problems caused by it if we are still using the old version of Hadoop. So we should test with recent versions, and consolidating versions makes it easier.

sekikn:
Thank you for the comments @Kimahriman @jiayuasu, all of the concerns you pointed out sound absolutely reasonable. I'll address them in the next commit. I also left answers for some questions.

jiayuasu (approved):

PR #655, opened 2022-08-05
[SEDONA-133] Allow user-defined schemas in Adapter.toDf()

brianrice2:

Yes, I have read [Contributor Rules](https://sedona.apache.org/community/rule/) and [Contributor Development Guide](https://sedona.apache.org/community/develop/)

Yes, the URL of the assoicated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-XXX. The PR name follows the format `[SEDONA-XXX] my subject`.

[Link to original ticket](https://issues.apache.org/jira/browse/SEDONA-133).

This expands the Adapter API to allow for users to convert to DataFrames with a given schema (for both SpatialRDD and JavaPairRDD).

User data is still stored in String format, so these new methods parse/cast the strings to whichever new data type is requested. This is similar to Spark's [UnivocityParser](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/csv/UnivocityParser.scala#L168-L285), which is used to parse CSV files, but unfortunately that functionality is not exposed publicly so I created a barebones version here. I didn't cover _all_ the data types, but tried to cover the key ones. [This page](https://spark.apache.org/docs/2.2.0/api/java/org/apache/spark/sql/Encoders.html) details the encoders that Spark uses and may be helpful to understand the appropriate data types for conversion. We could expand to cover more data types later, or I'm open to it now if you request it.

This also adds some helper private methods and refactors a few common operations into their own functions.

## How was this patch tested?

Added unit tests to confirm that SpatialRDD/JavaPairRDD -> DataFrame conversion works as expected.

**Note**: It doesn't seem to be common practice to test private methods, so I didn't add unit tests for the private methods I introduced. Their behavior is tested implicitly by the public functions.

Yes, I am adding a new API. I am using the [current SNAPSHOT version number](https://github.com/apache/incubator-sedona/blob/master/pom.xml#L29) in since `vX.Y.Z` format.

**Note**: I don't see another appropriate place to change documentation. Please let me know if I missed this!

### 1.

Is the following behavior intentional? I don't have a strong geospatial background.

In the JavaPairRDD -> DataFrame test case (called "can convert JavaPairRDD to DataFrame with user-supplied schema"), you may notice that the left and right dataframes get switched. The SpatialJoinQuery has `pointRDD` on the left and `polygonRDD` on the right, but the final output has `leftGeometry` of type POLYGON, followed by the `polygonRDD` user data fields, and `rightGeometry` of type POINT, followed by the `pointRDD` user data (null).


sql/src/main/scala/org/apache/sedona/sql/utils/Adapter.scala
```diff
@@ -175,4 +256,80 @@ object Adapter {
...
+      case _: FloatType => data.toFloat
+      case _: DoubleType => data.toDouble
+      case _: DateType => Date.valueOf(data)
+      case _: TimestampType => Timestamp.valueOf(data)
+      case _: BooleanType => data.toBoolean
+      case _: StringType => data
+      case _: StructType =>
+        val desiredStructSchema = desiredType.asInstanceOf[StructType]
+        new GenericRowWithSchema(parseStruct(data, desiredStructSchema), desiredStructSchema)
+    }
```
brianrice2, line 315: I couldn't find an elegant way to perform the null-safe conversion, so I have these long `match` statements

jiayuasu:
@brianrice2 

1. For your question `Is the following behavior intentional? I don't have a strong geospatial background.`, yes, it is intentional but it was due to my old silly design in GeoSpark... I didn't change it back due to backwards compatibility... Maybe we should change it back at some point...
 
2. Please add some documentation in https://github.com/apache/incubator-sedona/blob/master/docs/tutorial/sql.md#convert-between-dataframe-and-spatialrdd  to explain your API. Thank u!

brianrice2:
1. Thanks for providing that context! I'll create a followup Jira issue to discuss/prioritize separate from this issue. It would require extra care to be slotted into a major version update and communicated to users, so may be more trouble than it's worth.
2. Thanks for pointing that out&mdash;added some notes on converting with a schema

jiayuasu (approved):

PR #662, opened 2022-08-15
[SEDONA-142] Add ST_Collect to Flink Catalog

Elephantusparvus:

- Yes.

- Yes, the URL of the assoicated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-142. The PR name follows the format `[SEDONA-142] Add ST_Collect to Flink Catalog`.

I actually created to JIRA tickets by accident ([SEDONA-141](https://issues.apache.org/jira/browse/SEDONA-141)). One could/should be deleted. Sorry for that.

This is the first attempt to add ST_Collect aggregation to the flink api.

## How was this patch tested?
In progress
For myself I only tested aggregations with group by.
I guess it would be necessary to test window/over aggregations as well. Does anyone have specific data/queries for these in mind?

Currently not.

In advance this pull request is mainly done to discuss if this is a feasible approach to implement ST_Collect.
The PR would have to be cleaned up (split up), tests would need to be added and it would need to be documented.

I basically took the Internal Collect Aggregator from flink-table and adjusted it's types, letting it return a GeometryCollection.
The getTypeInference needs to be overriden otherwise reflection fails for the mapview / selecting the right serializer fails.
One Problem is/was that this function is not instantiated from a BuiltInFunctionDefinition/, 
so it is not possible to implement it as BuiltInAggregateFunction and let it pass a SpecializedFunction.SpecializedContext to its constructor. And flinks collect function is constructed through its AggFunctionFactory which also passes the argType.
So I needed to pass the Datatype for the geometry manually when constructing and adding it to the catalog, therefore needing a util function in the registrator.

If needed I add separate prs to ST_MakeValid which is an almost 1-t port/copy from the spark version, and the ST_Intersection, ST_Difference which just wrap the jts locationtech functions.

### DIFF
--- a/flink/src/main/java/org/apache/sedona/flink/SedonaFlinkRegistrator.java
+++ b/flink/src/main/java/org/apache/sedona/flink/SedonaFlinkRegistrator.java
@@ -13,12 +13,17 @@
  */
 package org.apache.sedona.flink;
 
+import org.apache.calcite.runtime.Geometries;
+import org.apache.flink.api.java.typeutils.runtime.kryo.KryoSerializer;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.table.api.DataTypes;
 import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
-import org.apache.sedona.common.geometryObjects.Circle;
+import org.apache.flink.table.functions.UserDefinedFunction;
+import org.apache.sedona.core.geometryObjects.Circle;
 import org.apache.sedona.core.geometryObjects.GeometrySerde;
 import org.apache.sedona.core.geometryObjects.SpatialIndexSerde;
 import org.apache.sedona.core.spatialPartitioning.PartitioningUtils;
+import org.apache.sedona.flink.expressions.ST_Collect;
 import org.locationtech.jts.geom.*;
 import org.locationtech.jts.index.quadtree.Quadtree;
 import org.locationtech.jts.index.strtree.STRtree;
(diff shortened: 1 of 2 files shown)


flink/src/main/java/org/apache/sedona/flink/Catalog.java
```diff
@@ -32,6 +32,9 @@ public static UserDefinedFunction[] getFuncs() {
                 new Constructors.ST_GeomFromGeoJSON(),
                 new Constructors.ST_GeomFromGeoHash(),
                 new Functions.ST_Buffer(),
+                new Functions.ST_MakeValid(),
+                new Functions.ST_Intersection(),
+                new Functions.ST_Difference(),
```
Kimahriman, line 37: Can you just make a separate PR for these and add them to the new common module created in https://github.com/apache/incubator-sedona/pull/647? Even though they're pretty simple it will help not to maintain both definitions in flink and spark

Elephantusparvus: Yeah, just saw that #647 got merged earlier this morning, so i have to merge it in my branch anyways. I'll clean it up and make a seperate PR for them.

Elephantusparvus: I pushed a rebased version with #647  only containing the ST_Collect function. 
Apparently #647 broke the build process for me, see: https://github.com/apache/incubator-sedona/pull/647#issuecomment-1216570079.

jiayuasu:
Thanks for your contribution. I have a much neat implementation of spatial aggregation in Flink using the UDAF api. It has been there for a while and I just forgot to create a PR.

I will create a new PR in one or two days. Then you can follow my implementation to tweak your ST_Collect a bit.

jiayuasu:
Please see PR #672 for the updates

Elephantusparvus:
Hi, sorry for the late response.
I saw that this was marked as part of the milestone.
Unfortunately time is limited for me right now and I am not able to work on this for the next 1-2 weeks.
If there is a scheduled timeline for the release of 1.3, I guess this PR should not hold it back.
If in the meantime someone else wants to improve this PR feel free to do so, otherwise I hopefully will be able to do this roughly until the beginning of October.

flink/src/main/java/org/apache/sedona/flink/SedonaFlinkRegistrator.java
```diff
@@ -36,9 +41,24 @@ public static void registerFunc(StreamTableEnvironment tblEnv) {
         );
     }
 
+    /**
+     * This is needed because for Resolving the RAW Geometry DataType in ST_Collect (MapView in Accumulator)
+     * the serializer needs to be known. To get the registered Serializer the ExecutionEnvironment is needed and
+     * to add the UDF the TableEnvironment ist needed.
```
jbampton, line 47:
```suggestion
     * to add the UDF the TableEnvironment is needed.
```

PR #677, opened 2022-09-04
[SEDONA-157] Add coordinate accessors to the Flink API

sekikn:

- Yes, the URL of the assoicated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-157. The PR name follows the format `[SEDONA-XXX] my subject`.

This PR adds ST_X, ST_Y and ST_Z to the Flink API.

## How was this patch tested?

- Ran `mvn clean install` locally
- Ran `mkdocs serve` and checked the generate document

- Yes, I am adding a new API. I am using the [1.3.0](https://github.com/apache/incubator-sedona/blob/master/pom.xml#L29) in since `vX.Y.Z` format.


common/src/main/java/org/apache/sedona/common/Functions.java
```diff
@@ -70,6 +70,27 @@ public static double length(Geometry geometry) {
         return geometry.getLength();
     }
 
+    public static Double x(Geometry geometry) {
```
Kimahriman, line 73: Does this need to be the boxed type instead of the native double like the other functions return?

sekikn: ST_X, ST_Y and ST_Z should return null if the first argument is not a Point, so we have to check the input type somewhere. I chose to do that in the common functions, because if they can direcly return null by using a boxed type, we can simply wrap it with InferredUnaryExpression in the Spark SQL functions. But as you pointed out, returning boxed type looks somewhat strange. Should I change the common functions to return a primitive value? In that case, we have to use UnaryGeometryExpression instead of InferredUnaryExpression so the code will become a bit verbose. Either is fine with me.

Kimahriman: Right that makes sense then, nvm

flink/src/test/java/org/apache/sedona/flink/FunctionTest.java
```diff
@@ -265,6 +265,27 @@ public void testIsEmpty() {
         assertEquals(false, result);
     }
 
+    @Test
+    public void testX() {
```
jiayuasu, line 269: Can you use the same test in testZ for testX and testY?

sekikn: Updated the PR. Thanks for the review!

PR #693, opened 2022-09-21
[SEDONA-166] Support type-safe dataframe style API

douglasdennis:

- Yes, the URL of the assoicated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-166. The PR name follows the format `[SEDONA-XXX] my subject`.

Introduce a type-safe DataFrame style API similar to standard Spark functions. The API can operate on Spark `Column` types directly or function specific native Scala/Java types. For example, the function `ST_CollectionExtract` can be called in the following ways:

```scala
val df = sparkSession.sql("SELECT ST_GeomFromWKT('GEOMETRYCOLLECTION(POINT(0 0), LINESTRING(0 0, 1 0))') AS geom")

// with Column objects and default geomType argument
df.select(ST_CollectionExtract($"geom")

// with a String to specify a column name
df.select(ST_CollectionExtract("geom")

// using Column objects and specifying the geomType argument
df.select(ST_CollectionExtract($"geom", lit(1))

// using a String for the column name and an Int for geomType
df.select(ST_CollectionExtract("geom", 1)
```

The general rule for the API is that methods have two overloaded forms:
1. All `Column` arguments which is the most versatile.
2. Strings for arguments that are commonly column names, and native types for arguments that are commonly constants. For example, the ST_CollectionExtract geomType argument (when given) is generally constant across the dataframe.

This API is made available to Scala, Java, and Python.

## How was this patch tested?

Scala/Java is tested through unit tests in dataFrameAPITestScala.scala and Python is tested through unit tests in test_dataframe_api.py.

- Yes, I am adding a new API. I am using the [current SNAPSHOT version number](https://github.com/apache/incubator-sedona/blob/master/pom.xml#L29) in since `vX.Y.Z` format.

### DIFF
--- a/python/sedona/sql/dataframe_api.py
+++ b/python/sedona/sql/dataframe_api.py
@@ -59,7 +59,9 @@ def _get_type_list(annotated_type: Type) -> Tuple[Type, ...]:
     :return: Tuple of all types covered by the type annotation.
     :rtype: Tuple[Type, ...]
     """
-    if typing.get_origin(annotated_type) is Union:
+    # in 3.8 there is a much nicer way to do this with typing.get_origin
+    # we have to be a bit messy until we drop support for 3.7
+    if isinstance(annotated_type, typing._GenericAlias) and annotated_type.__origin__._name == "Union":
         valid_types = typing.get_args(annotated_type)
     else:
         valid_types = (annotated_type,)
--- a/python/sedona/sql/dataframe_api.py
+++ b/python/sedona/sql/dataframe_api.py
@@ -82,8 +82,8 @@ def _get_readable_name_for_type(type: Type) -> str:
     :return: Human readable name for the type annotation.
     :rtype: str
     """
-    if typing.get_origin(type) is Union:
-        return f"Union[{', '.join((_strip_extra_from_class_name(str(x)) for x in typing.get_args(type)))}]"
+    if isinstance(type, typing._GenericAlias) and type.__origin__._name == "Union":
+        return f"Union[{', '.join((_strip_extra_from_class_name(str(x)) for x in type.__args__))}]"
     return _strip_extra_from_class_name(str(type))
 
 
--- a/python/sedona/sql/dataframe_api.py
+++ b/python/sedona/sql/dataframe_api.py
@@ -1,4 +1,18 @@
-from typing import Any, Tuple, Union, Iterable
+import functools
+import inspect
+import itertools
+from typing import (
+    Any,
+    Callable,
+    Iterable,
+    List,
+    Mapping,
+    Tuple,
+    Type,
+    Union,
+)
+import typing
+
 from pyspark.sql import SparkSession, Column, functions as f
 
 
@@ -22,8 +36,9 @@ def call_sedona_function(object_name: str, function_name: str, args: Union[Any, 
     if spark is None:
         raise ValueError("No active spark session was detected. Unable to call sedona function.")
     
-    if not isinstance(args, Iterable) or isinstance(args, str):
-        args = (args,)
+    # apparently a Column is an Iterable so we need to check for it explicitely
+    if (not isinstance(args, Iterable)) or isinstance(args, str) or isinstance(args, Column):
+        args = [args]
 
     args = map(_convert_argument_to_java_column, args)
 
@@ -32,4 +47,91 @@ def call_sedona_function(object_name: str, function_name: str, args: Union[Any, 
     
     jc = jfunc(*args)
     return Column(jc)
-        
+
+
+def _get_type_list(annotated_type: Type) -> Tuple[Type, ...]:
+    """Convert a type annotation into a tuple of types.
+    For most types this will be a tuple with a single element, but
+    for Union a tuple with multiple elements will be returned.
+
+    :param annotated_type: Type annotation to convert a tuple.
+    :type annotated_type: Type
+    :return: Tuple of all types covered by the type annotation.
+    :rtype: Tuple[Type, ...]
+    """
+    if typing.get_origin(annotated_type) is Union:
+        valid_types = typing.get_args(annotated_type)
+    else:
+        valid_types = (annotated_type,)
+    
+    return valid_types
+
+
+def _strip_extra_from_class_name(class_name):
+    return class_name[len("<class '"):-len("'>")].split(".")[-1]
+
+
+def _get_readable_name_for_type(type: Type) -> str:
+    """Get a human readable name for a type annotation used on a function's parameter.
+
+    :param type: Type annotation for a parameter.
+    :type type: Type
+    :return: Human readable name for the type annotation.
+    :rtype: str
+    """
+    if typing.get_origin(type) is Union:
+        return f"Union[{', '.join((_strip_extra_from_class_name(str(x)) for x in typing.get_args(type)))}]"
+    return _strip_extra_from_class_name(str(type))
+
+
+def _get_bound_arguments(f: Callable, *args, **kwargs) -> Mapping[str, Any]:
+    """Bind the passed arguments to f with actual parameter names, including defaults.
+
+    :param f: Function to bind arguments for.
+    :type f: Callable
+    :return: Dictionary of parameter names to argument values.
+    :rtype: Mapping[str, Any]
+    """
+    f_signature = inspect.signature(f)
+    bound_args = f_signature.bind(*args, **kwargs)
+    bound_args.apply_defaults()
+    return bound_args
+
+
+def _check_bound_arguments(bound_args: Mapping[str, Any], type_annotations: List[Type], function_name: str) -> None:
+    """Check bound arguments against type annotations and raise a ValueError if any do not match.
+
+    :param bound_args: _description_
+    :type bound_args: List[Any]
+    :param type_annotations: _description_
+    :type type_annotations: List[Type]
+    :param function_name: _description_
+    :type function_name: str
+    :raises ValueError: _description_
+    """
+    for bound_arg_name, bound_arg_value in bound_args.arguments.items():
+        annotated_type = type_annotations[bound_arg_name]
+        valid_type_list = _get_type_list(annotated_type)
+        if not any([isinstance(bound_arg_value, valid_type) for valid_type in valid_type_list]):
+            raise ValueError(f"Incorrect parameter type: {bound_arg_name} for {function_name} should be {_get_readable_name_for_type(annotated_type)} but received {_strip_extra_from_class_name(str(type(bound_arg_value)))}.")
+
+
+def validate_argument_types(f: Callable) -> Callable:
+    """Validates types of arguments passed to a dataframe API style function.
+    Arguments will need to be either strings, columns, or match the typehints of f.
+    This function is meant to be used a decorator.
+
+    :param f: Function to validate for.
+    :type f: Callable
+    :return: f wrapped with type validation checks.
+    :rtype: Callable
+    """
+    def validated_function(*args, **kwargs) -> Column:
+        # all arguments are Columns or strings are always legal, so only check types when one of the arguments is not a column
+        if not all([isinstance(x, Column) or isinstance(x, str) for x in itertools.chain(args, kwargs.values())]):
+            bound_args = _get_bound_arguments(f, *args, **kwargs)
+            type_annotations = typing.get_type_hints(f)
+            _check_bound_arguments(bound_args, type_annotations, f.__name__)
+
+        return f(*args, **kwargs)
+    return functools.update_wrapper(validated_function, f)
--- a/python/sedona/sql/dataframe_api.py
+++ b/python/sedona/sql/dataframe_api.py
@@ -1,0 +1,35 @@
+from typing import Any, Tuple, Union, Iterable
+from pyspark.sql import SparkSession, Column, functions as f
+
+
+ColumnOrName = Union[Column, str]
+ColumnOrNameOrNumber = Union[Column, str, float, int]
+
+
+def _convert_argument_to_java_column(arg: Any) -> Column:
+    if isinstance(arg, Column):
+        return arg._jc
+    elif isinstance(arg, str):
+        return f.col(arg)._jc
+    elif isinstance(arg, Iterable):
+        return f.array(*[Column(x) for x in map(_convert_argument_to_java_column, arg)])._jc
+    else:
+        return f.lit(arg)._jc
+
+
+def call_sedona_function(object_name: str, function_name: str, args: Union[Any, Tuple[Any]]) -> Column:
+    spark = SparkSession.getActiveSession()
+    if spark is None:
+        raise ValueError("No active spark session was detected. Unable to call sedona function.")
+    
+    if not isinstance(args, Iterable) or isinstance(args, str):
+        args = (args,)
+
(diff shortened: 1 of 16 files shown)


sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/st_aggregates.scala
```diff
@@ -0,0 +1,40 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.spark.sql.sedona_sql.expressions
+
+import org.apache.spark.sql.catalyst.expressions.Expression
+import org.apache.spark.sql.Column
+import org.apache.spark.sql.sedona_sql.expressions_udaf.{ST_Envelope_Aggr => ST_Envelope_Aggr_udaf, ST_Intersection_Aggr => ST_Intersection_Aggr_udaf, ST_Union_Aggr => ST_Union_Aggr_udaf}
+
+object st_aggregates extends DataFrameAPI {
```
jiayuasu, line 25: Sedona aggregator is already using type-safe aggregator: https://github.com/apache/incubator-sedona/blob/master/sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/AggregateFunctions.scala#L76

The old UDAF-based aggregator (https://github.com/apache/incubator-sedona/blob/master/sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions_udaf/AggregateFunctions.scala#L27) is not used in Sedona for Spark 3.0 version: https://github.com/apache/incubator-sedona/blob/master/sql/src/main/scala/org/apache/sedona/sql/UDF/UdfRegistrator.scala#L44

Spark 2.X support will be dropped in Sedona 1.3.0 (next release). Therefore, I don't think we need this part. What do you think?

douglasdennis:
This is pretty embarrassing. This was meant to PR against my own fork's master to get the build action to run. While most of these changes are approaching done (on the Scala side anyways), it was still subject to change as I'm sorting out the python and Java interface. The cat is out of the bag though so please consider this a draft for the moment.

jiayuasu:
@douglasdennis Haha, no problem. This PR looks pretty nice. Keep up the good work!

sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/st_aggregates.scala
douglasdennis: Are you asking if the st_aggregates object is needed? If so then I think it does. This allows for something like this:
```
df.groupBy().agg(ST_Union_Aggr("geometry").as("geometry"))
```
If that's not what you meant then apologies.

I didn't realize that UserDefinedAggregateFunction was depricated though. I was actually using this tutorial for reference to understand how this works on the Scala side: https://docs.databricks.com/spark/latest/spark-sql/udaf-scala.html

It looks like from what else I've read that I'm supposed to pass an Aggregator to the udaf function, so I'll plan to do that.

jiayuasu: UserDefinedAggregateFunction is deprecated since Spark 3.0 and Spark 3.0 suggests the type-safe aggregator: https://spark.apache.org/docs/latest/sql-ref-functions-udf-aggregate.html#examples

So Sedona for Spark 3.0 uses the type-safe aggregator. Sedona for Spark 2.0 still uses UserDefinedAggregateFunction, which will be completely dropped in Sedona 1.3.0 (next release).

The URL you pasted actually uses the deprecated UserDefinedAggregateFunction as the example.

douglasdennis: I'm shaking my fist at databricks right now...

This makes sense. I'll switch to using the type safe Aggregator classes and using the spark udaf function to allow it to take columns.

douglasdennis:
@jiayuasu Thank you :) I should also mention this is some of my first Scala I've written. So, if there is any language faux pas or better ways to accomplish what I'm doing, then let me know and I'll fix it.

jiayuasu:
@douglasdennis Another interesting finding: since the type-safe dataframe APIs do not need to call "udf.register()" to register all functions, is it possible that, as a side effect, predicate pushdown is finally supported in Sedona?

If you don't know what a predicate pushdown is, see this: https://jaceklaskowski.gitbooks.io/mastering-spark-sql/content/spark-sql-Optimizer-PushDownPredicate.html

Since Sedona 1.3.0 will natively read GeoParquet (PR already merged to the master), Sedona user  RJ Marcus is working hard to get the predicate pushdown work on GeoParquet. See [SEDONA-156](https://issues.apache.org/jira/browse/SEDONA-156).

Quoted my reply to RJ Marcus:

> Pushed filter: UDF function can be pushed down as a filter. Based on my understanding, Sedona ST functions cannot be pushed down because UDFs in pure Spark SQL are blackbox to Spark catalyst unless we do something with the current Sedona ST functions.

> However, Sedona implements all functions in Spark SQL Catalyst "Expressions" [1] instead of the naive UDF. This gives you the possibility to push them down to the data source (see [2]). There is an ongoing effort to enable Sedona ST functions in type-safe format which bypasses the "udf.register" step (see [3])

> So, with the current Sedona GeoParquet reader, and [3], it is possible that the Pushed filter will be finally supported. You might want to check it out and confirm my wild guess.

> [1] https://github.com/apache/incubator-sedona/blob/master/sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Constructors.scala#L45

> [2] https://neapowers.com/apache-spark/native-functions-catalyst-expressions/

> [3] https://github.com/apache/incubator-sedona/pull/693
```

**Comment by Kimahriman** [comment]
> @douglasdennis Another interesting finding: since the type-safe dataframe APIs do not need to call "udf.register()" to register all functions, is it possible that, as a side effect, predicate pushdown is finally supported in Sedona?

I believe the current mechanism doesn't use UDFs technically, it registers SQL functions directly which _should_ theoretically be usable by predicate pushdown (if you can figure out that mechanism in V1). That's why we can do things like join detection with pattern matching on the expressions

**Comment by douglasdennis** [comment]
> Another interesting finding: since the type-safe dataframe APIs do not need to call "udf.register()" to register all functions, is it possible that, as a side effect, predicate pushdown is finally supported in Sedona?

@jiayuasu It does not appear to be so. Here are some results I ran this morning using the example1.parquet in the library.

Checking to make sure predicate pushdown happens with native types:
```scala
val geoparquetdatalocation1: String = resourceFolder + "geoparquet/example1.parquet"
val basicPredicateDf = sparkSession.read.format("geoparquet").load(geoparquetdatalocation1).where(col("name").equalTo("Fiji"))
basicPredicateDf.explain()
```
The plan shows a push down:
```
== Physical Plan ==
*(1) Filter (isnotnull(name#5302) AND (name#5302 = Fiji))
+- FileScan geoparquet [pop_est#5300L,continent#5301,name#5302,iso_a3#5303,gdp_md_est#5304,geometry#5305] Batched: false, DataFilters: [isnotnull(name#5302), (name#5302 = Fiji)], Format: Parquet, Location: InMemoryFileIndex(1 paths)[file:<removed>, PartitionFilters: [], PushedFilters: [IsNotNull(name), EqualTo(name,Fiji)], ReadSchema: struct<pop_est:bigint,continent:string,name:string,iso_a3:string,gdp_md_est:double,geometry:array...
```
\
\
A simple geometry based predicate:
```scala
val geoparquetdatalocation1: String = resourceFolder + "geoparquet/example1.parquet"
val basicGeomPredicateDf = sparkSession.read.format("geoparquet").load(geoparquetdatalocation1).where(ST_GeometryType("geometry").equalTo(lit("ST_Polygon")))
basicGeomPredicateDf.explain()
```
This plan does not show a push down:
```
== Physical Plan ==
Filter (st_geometrytype(geometry#5318) = ST_Polygon)
+- FileScan geoparquet [pop_est#5313L,continent#5314,name#5315,iso_a3#5316,gdp_md_est#5317,geometry#5318] Batched: false, DataFilters: [(st_geometrytype(geometry#5318) = ST_Polygon)], Format: Parquet, Location: InMemoryFileIndex(1 paths)[file:<removed>, PartitionFilters: [], PushedFilters: [], ReadSchema: struct<pop_est:bigint,continent:string,name:string,iso_a3:string,gdp_md_est:double,geometry:array...
```
\
\
And just to be thorough, a more complex geometry based predicate:
```scala
val geoparquetdatalocation1: String = resourceFolder + "geoparquet/example1.parquet"
val basicGeomPredicateDf = sparkSession.read.format("geoparquet").load(geoparquetdatalocation1).where(ST_Distance("geometry", ST_Point(2, 2)) <= (50.0))
basicGeomPredicateDf.explain()
```
As expected, no push down as well:
```
== Physical Plan ==
Filter ( **org.apache.spark.sql.sedona_sql.expressions.ST_Distance**   <= 50.0)
+- FileScan geoparquet [pop_est#5326L,continent#5327,name#5328,iso_a3#5329,gdp_md_est#5330,geometry#5331] Batched: false, DataFilters: [( **org.apache.spark.sql.sedona_sql.expressions.ST_Distance**   <= 50.0)], Format: Parquet, Location: InMemoryFileIndex(1 paths)[file:<removed>, PartitionFilters: [], PushedFilters: [], ReadSchema: struct<pop_est:bigint,continent:string,name:string,iso_a3:string,gdp_md_est:double,geometry:array...
```

neontty:
@douglasdennis this looks like an awesome PR, thank you for putting in this hard work. 

I agree that this PR won't shortcut to geometry based predicate pushdowns; the issue comes with handling Expressions/Predicates correctly during the FileScan since a lot of parquet code deals with "Filter" instead of Expression or Predicate. Thanks Dennis for doing this most recent test to confirm. 

I'm a little bit busy today but I'm taking a look at Jia's comments on SEDONA-156 and making a response to continue the predicate-pushdown conversation in that thread.

douglasdennis:
@jiayuasu This should be, at least structurally, complete. I still have docs and docstrings to complete. A couple questions:

1. I was planning to add a new section to this [page](https://sedona.apache.org/tutorial/sql/) and this [page](https://sedona.apache.org/tutorial/sql-python/) demonstrating how to use this API. Does that seem like a good way to go about it? Or would you prefer something else?
2. Should I add scaladoc to the Scala code?
3. I found sphinx style docstrings in some python code. Is that the style that I should use in Python?

I believe that @Imbruced is the Python reviewer, so I'm pinging them for this as well.

Any comments, suggestions, or code changes needed please let me know.

python/sedona/sql/dataframe_api.py
```diff
@@ -0,0 +1,35 @@
...
+ColumnOrNameOrNumber = Union[Column, str, float, int]
+
+
+def _convert_argument_to_java_column(arg: Any) -> Column:
+    if isinstance(arg, Column):
+        return arg._jc
+    elif isinstance(arg, str):
+        return f.col(arg)._jc
+    elif isinstance(arg, Iterable):
+        return f.array(*[Column(x) for x in map(_convert_argument_to_java_column, arg)])._jc
```
Imbruced, line 15: Whats the advantage of using map and list comprehension in one place instead of applying two functions using list comprehension ?

python/sedona/sql/st_constructors.py
```diff
@@ -0,0 +1,79 @@
...
+    "ST_LineFromText",
+    "ST_LineStringFromText",
+    "ST_Point",
+    "ST_PointFromText",
+    "ST_PolygonFromEnvelope",
+    "ST_PolygonFromText",
+]
+
+
+_call_constructor_function = partial(call_sedona_function, "st_constructors")
```
Imbruced, line 26: I like the idea of using partial :)

python/tests/sql/test_dataframe_api.py
```diff
@@ -0,0 +1,1087 @@
...
+
+from sedona.sql import st_functions as stf
+from sedona.sql import st_constructors as stc
+from sedona.sql import st_aggregates as sta
+from sedona.sql import st_predicates as stp
+from tests.test_base import TestBase
+
+class TestDataFrameAPI(TestBase):
+
+    def test_call_function_using_columns(self):
```
Imbruced, line 13: Do you think parametrised tests can help to simplify the tests ? I mean https://docs.pytest.org/en/6.2.x/parametrize.html ?

Imbruced: Then you can use argument list, function name to apply and also some fixtures to provide dataframe ?

Imbruced:
Great improvement ! The main idea is to provide the type safe function and what I am missing the most is to validate input types and verify against None values. Also I would like to have test cases for that. WDYT ? Maybe in another PR because this is massive :) Thanks for your effort.

Imbruced (changes requested):

python/sedona/sql/dataframe_api.py
douglasdennis: No advantage. Just me being silly when I refactored. I originally wanted this to use recursion in a different way, but I wasn't feeling it so I defaulted to this out of haste :) Will refactor to use function composition. Thanks for catching that.

python/tests/sql/test_dataframe_api.py
douglasdennis: Ooohhhh! I have not used parameterized tests before. I'm excited to give them a shot.

douglasdennis:
> Great improvement ! The main idea is to provide the type safe function and what I am missing the most is to validate input types and verify against None values. Also I would like to have test cases for that. WDYT ? Maybe in another PR because this is massive :) Thanks for your effort. 

Doh! I had meant to have input validation and I completely forgot about it. I'd like to add those in this PR just for completion. The jvm call will throw if it can't find a method that accepts the given argument types but that would be cryptic to the user. For reference, I intend to use a decorator to manage the type checking.

Imbruced:
@douglasdennis Thanks ! Decorator sounds great ! It's good to have Python exception raised which can be easier handled and debugged later by users :)

jiayuasu:
@douglasdennis Please let me know once you finish this PR :-)

jiayuasu:
@douglasdennis Is this PR ready to merge? :-)

douglasdennis:
> @douglasdennis Is this PR ready to merge? :-)

Just working on a couple of the notes now. Will be done by the end of this weekend.

python/tests/sql/test_dataframe_api.py
douglasdennis: Alright. I gave this a shot. Let me know if it isn't what you were hoping for.

Also: I don't know what the proper etiquette is on github. Do I click the "Resolve Conversation" button when I think I did it or do the folks requesting code changes do that when they are satisfied?

python/sedona/sql/dataframe_api.py
douglasdennis: This is fixed.

douglasdennis:
Victory!

@jiayuasu  As best I know this is ready to go, assuming I have addressed the notes from @Imbruced . I can also do a follow up PR after this merges if that would be better.

jiayuasu (approved):

jiayuasu:
This looks good to me. I will merge it for now as a few other PRs are waiting for this. If @Imbruced has any comments, @douglasdennis can make a follow-up PR to address it.

PR #696, opened 2022-10-01
Sedona 173 delta lake tests

Imbruced:

- Yes, the URL of the assoicated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-173. The PR name follows the format `[SEDONA-173] my subject`.

delta lakes tests in scala and python notebook with examples how to use geospatial functions with delta lake.

## How was this patch tested?
Unit tests

### DIFF
--- a/binder/ApacheSedonaDeltaLake.ipynb
+++ b/binder/ApacheSedonaDeltaLake.ipynb
@@ -1,0 +1,754 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "aec5096c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from pyspark.sql import SparkSession\n",
+    "from delta import *\n",
+    "from sedona.register import SedonaRegistrator\n",
+    "from sedona.utils import SedonaKryoRegistrator, KryoSerializer\n",
+    "from dataclasses import dataclass\n",
+    "from shapely.geometry import Point\n",
+    "from shapely.geometry import Polygon\n",
+    "from pyspark.sql import DataFrame\n",
+    "from os import getcwd\n",
+    "import shutil\n",
+    "import os\n",
+    "from sedona.sql.types import GeometryType\n",
+    "from shapely.wkt import loads\n",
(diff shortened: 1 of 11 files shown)


delta/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+            <groupId>org.apache.sedona</groupId>
+            <artifactId>sedona-core-${spark.compat.version}_${scala.compat.version}</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.delta</groupId>
+            <artifactId>delta-core_2.12</artifactId>
+            <version>2.0.0</version>
+        </dependency>
```
jiayuasu, line 36: The scope should be `test`

delta/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+        <artifactId>sedona-parent</artifactId>
+        <groupId>org.apache.sedona</groupId>
+        <version>1.1.1-incubating</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>delta</artifactId>
+
+    <properties>
+        <maven.compiler.source>8</maven.compiler.source>
```
jiayuasu, line 15: Please add `<maven.deploy.skip>false</maven.deploy.skip>` . This will prevent this module from being published to Maven Central.

jiayuasu (changes requested):

Kimahriman:
Just my opinion, this seems like the only thing this is really "testing" is if Delta supports user-defined types, which is more of a thing for the Delta repo than this one, and doesn't really seem like it makes sense as a new module to add. But this makes perfect sense to just be an example notebook on how you can work with geometries natively with Delta.

Imbruced:
@Kimahriman I also agree, the main reason why I created additional module was to figure out later if we can add by default writing to delta in geoparquet data format.

jiayuasu:
@Imbruced Yes, I agree with @Kimahriman . The only purpose of this module is testing. Can you (1) place the Scala/Java part to /example/ folder (2) keep the Python notebook in /binder/  ?

PR #704, opened 2022-10-26
[SEDONA-180] Make ST_* expressions foldable and declares input types for type checking

Kontinuation:

- Yes, the URL of the assoicated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-180.

This patch makes ST_* functions foldable, and declared input types for these functions to make the SQL analyzer perform type checking for these ST_* function calls and emit more friendly error messages. Raster functions and viz functions are still left unchanged.

I've deliberately left `ST_Binary` and `ST_AsEWKB` unfoldable since unit tests would fail after making these expressions foldable.
The reason is that folding binary expressions as constants triggers a function call to [`encodeHexString(byte[], boolean)`](https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Hex.html#encodeHexString-byte:A-boolean-) in spark-sql, while the bundled Apache commons-codec of sernetcdf jar do not have this function defined. We may have to release a new version of sernetcdf with a newer version of / without commons-codec bundled.

I've also moved some implementations of ST functions to `sedona-common`, so that they can be reused by other distributed computation engines.

I also spotted a bug in `ST_MinimumBoundingCircle` dataframe API, it does not honor the second parameter. This patch also includes a fix for that bug.

## How was this patch tested?

Hopefully existing unit tests are sufficient to guarantee that I'm not breaking anything.

### DIFF
--- a/sql/src/main/scala/org/apache/sedona/sql/UDF/Catalog.scala
+++ b/sql/src/main/scala/org/apache/sedona/sql/UDF/Catalog.scala
@@ -18,140 +18,149 @@
  */
 package org.apache.sedona.sql.UDF
 
+import org.apache.spark.sql.catalyst.FunctionIdentifier
+import org.apache.spark.sql.catalyst.expressions.{ExpectsInputTypes, Expression, ExpressionInfo}
+import org.apache.spark.sql.catalyst.expressions.Literal
 import org.apache.spark.sql.catalyst.analysis.FunctionRegistry.FunctionBuilder
 import org.apache.spark.sql.expressions.{Aggregator, UserDefinedAggregateFunction}
 import org.apache.spark.sql.sedona_sql.expressions.{ST_YMax, ST_YMin, _}
 import org.apache.spark.sql.sedona_sql.expressions.collect.{ST_Collect, ST_CollectionExtract}
 import org.apache.spark.sql.sedona_sql.expressions.raster.{RS_Add, RS_Append, RS_Array, RS_Base64, RS_BitwiseAnd, RS_BitwiseOr, RS_Count, RS_Divide, RS_FetchRegion, RS_GetBand, RS_GreaterThan, RS_GreaterThanEqual, RS_HTML, RS_LessThan, RS_LessThanEqual, RS_LogicalDifference, RS_LogicalOver, RS_Mean, RS_Mode, RS_Modulo, RS_Multiply, RS_MultiplyFactor, RS_Normalize, RS_NormalizedDifference, RS_SquareRoot, RS_Subtract}
 import org.locationtech.jts.geom.Geometry
+import org.locationtech.jts.operation.buffer.BufferParameters
+
+import scala.reflect.ClassTag
 
 object Catalog {
-  val expressions: Seq[FunctionBuilder] = Seq(
+
+  type FunctionDescription = (FunctionIdentifier, ExpressionInfo, FunctionBuilder)
+
+  val expressions: Seq[FunctionDescription] = Seq(
     // Expression for vectors
-    ST_PointFromText,
-    ST_PolygonFromText,
-    ST_LineStringFromText,
-    ST_GeomFromText,
-    ST_LineFromText,
-    ST_GeomFromWKT,
-    ST_GeomFromWKB,
-    ST_GeomFromGeoJSON,
-    ST_GeomFromGML,
-    ST_GeomFromKML,
-    ST_Point,
-    ST_PolygonFromEnvelope,
-    ST_Contains,
-    ST_Intersects,
-    ST_Within,
-    ST_Covers,
-    ST_CoveredBy,
-    ST_Disjoint,
-    ST_Distance,
-    ST_3DDistance,
-    ST_ConvexHull,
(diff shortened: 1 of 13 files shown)


sql/src/main/scala/org/apache/sedona/sql/UDF/Catalog.scala
```diff
@@ -18,140 +18,149 @@
...
+import scala.reflect.ClassTag
 
 object Catalog {
-  val expressions: Seq[FunctionBuilder] = Seq(
+
+  type FunctionDescription = (FunctionIdentifier, ExpressionInfo, FunctionBuilder)
+
+  val expressions: Seq[FunctionDescription] = Seq(
     // Expression for vectors
-    ST_PointFromText,
```
jiayuasu, line 31: Is there a specific reason to change the way we define functions in the Catalog?

sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Constructors.scala
```diff
@@ -305,41 +325,29 @@ case class ST_GeomFromGeoJSON(inputExpressions: Seq[Expression])
 /**
   * Return a Point from X and Y
   *
-  * @param inputExpressions This function takes 2 parameter which are point x and y.
+  * @param inputExpressions This function takes 3 parameter which are point x, y and z.
   */
 case class ST_Point(inputExpressions: Seq[Expression])
-  extends Expression with CodegenFallback with UserDataGeneratator {
-  inputExpressions.betweenLength(2, 3)
```
jiayuasu, line 312: Is there a reason why you remove the inputExpression (2, 3) requirement?

I saw similar inputExpression.length assertions were kept in some places and removed in some other places. Is there a reason?

sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Predicates.scala
```diff
@@ -25,39 +25,47 @@ import org.apache.spark.sql.catalyst.expressions.codegen.CodegenFallback
...
 
+/**
+  * Test if leftGeometry full contains rightGeometry
+  *
+  * @param inputExpressions
+  */
+case class ST_Contains(inputExpressions: Seq[Expression])
+  extends ST_Predicate with CodegenFallback {
+
+  override def toString: String = s" **${ST_Contains.getClass.getName}**  "
```
jiayuasu, line 63: the toString() function is only provided in this predicate. Can you put the toString() function to ST_Predicate (e.g., toString(): getClass.getName) ?

jiayuasu (changes requested):
Thank you so much for creating this PR. This PR is huge and it can benefit lots of users!

I only have a few minor comments. Please address them as you see fit.

sql/src/main/scala/org/apache/sedona/sql/UDF/Catalog.scala
Kontinuation: This is for supporting expressions with optional parameters while declaring the types of all parameters in `def inputTypes`. Such as `ST_MakeValid(geom: Geometry, keepCollapsed: Boolean = false)`, where the second parameter is optional.

Before this patch expressions with optional parameters need to check the number of passed in arguments and resolve the values of optional parameters manually. This patch handles optional parameters in function builder registered to the catalog, so that default argument values will be filled in. Now expressions can declare all its parameters in `def inputTypes`, all the arguments will be type-checked, and they don't need to handle default arguments themselves.

sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Constructors.scala
Kontinuation: Assertions for expressions with optional parameters looks strange since all parameters will receive their arguments before evaluation. This assertion became unnecessary since the function builder registered to the catalog will emit error messages for function calls with invalid number of parameters:

```
SELECT ST_Point(1)
>>> function ST_Point takes at least 2 argument(s), 1 argument(s) specified
```

For expressions with optional parameters like `ST_Point` we cannot simply assert `inputExpressions.size == 3` since there's one weird behavior in the implementation of the function builder: when 2 arguments were passed to `ST_Point`, it will construct a `ST_Point` object using these 2 arguments first, then call its `inputTypes` method to acknowledge that it actually has 3 parameters, so that it will construct the final `ST_Point` object with 2 user specified arguments and one default argument.

There're [better ways](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/analysis/FunctionRegistry.scala#L110) to support optional parameters without constructing intermediate invalid expressions, however that may involve huge changes to the definitions of expressions, so I'd like to live with it for now.

sql/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Predicates.scala
Kontinuation: Moved `toString` to `ST_Predicate`.

PR #706, opened 2022-11-01
[SEDONA-186] Fix ordering of columns in results of queries like SELECT * A JOIN B

Kontinuation:

- Yes, the URL of the assoicated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-186

Fixes [SEDONA-186](https://issues.apache.org/jira/browse/SEDONA-186) by properly assembling result rows of join queries when left and right sub plans were swapped.

## How was this patch tested?

New tests were added to `SpatialJoinSuite`.

### DIFF
--- a/sql/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/JoinQueryDetector.scala
+++ b/sql/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/JoinQueryDetector.scala
@@ -164,9 +164,9 @@ class JoinQueryDetector(sparkSession: SparkSession) extends Strategy {
     val relationship = s"ST_$spatialPredicate"
 
     matchExpressionsToPlans(a, b, left, right) match {
-      case Some((planA, planB, _)) =>
+      case Some((planA, planB, swappedLeftAndRight)) =>
         logInfo(s"Planning spatial join for $relationship relationship")
-        RangeJoinExec(planLater(planA), planLater(planB), a, b, spatialPredicate, extraCondition) :: Nil
+        RangeJoinExec(planLater(planA), planLater(planB), a, b, swappedLeftAndRight, spatialPredicate, extraCondition) :: Nil
       case None =>
         logInfo(
           s"Spatial join for $relationship with arguments not aligned " +
@@ -185,13 +185,13 @@ class JoinQueryDetector(sparkSession: SparkSession) extends Strategy {
     val b = children.tail.head
 
     matchExpressionsToPlans(a, b, left, right) match {
-      case Some((planA, planB, _)) =>
+      case Some((planA, planB, swappedLeftAndRight)) =>
         if (distance.references.isEmpty || matches(distance, planA)) {
           logInfo("Planning spatial distance join")
-          DistanceJoinExec(planLater(planA), planLater(planB), a, b, distance, spatialPredicate, extraCondition) :: Nil
+          DistanceJoinExec(planLater(planA), planLater(planB), a, b, swappedLeftAndRight, distance, spatialPredicate, extraCondition) :: Nil
         } else if (matches(distance, planB)) {
           logInfo("Planning spatial distance join")
-          DistanceJoinExec(planLater(planB), planLater(planA), b, a, distance, spatialPredicate, extraCondition) :: Nil
+          DistanceJoinExec(planLater(planB), planLater(planA), b, a, swappedLeftAndRight, distance, spatialPredicate, extraCondition) :: Nil
         } else {
           logInfo(
             "Spatial distance join for ST_Distance with non-scalar distance " +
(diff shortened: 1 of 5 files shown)


sql/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/TraitJoinQueryExec.scala
```diff
@@ -135,27 +129,25 @@ trait TraitJoinQueryExec extends TraitJoinQueryBase {
...
 
-      filtered.map {
-        case (l, r) =>
-          val leftRow = l.getUserData.asInstanceOf[UnsafeRow]
-          val rightRow = r.getUserData.asInstanceOf[UnsafeRow]
-          var joiner = GenerateUnsafeRowJoiner.create(left.schema, right.schema)
-          joiner.join(leftRow, rightRow)
+      extraCondition match {
+        case Some(condition) =>
+          val boundCondition = Predicate.create(condition, output)
```
Kontinuation, line 148: I assume that we don't need to add a SPARK2 anchor here since we'll drop Spark 2 support in the next release.

sql/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/JoinQueryDetector.scala
```diff
@@ -164,9 +164,9 @@ class JoinQueryDetector(sparkSession: SparkSession) extends Strategy {
     val relationship = s"ST_$spatialPredicate"
 
     matchExpressionsToPlans(a, b, left, right) match {
-      case Some((planA, planB, _)) =>
+      case Some((planA, planB, swappedLeftAndRight)) =>
         logInfo(s"Planning spatial join for $relationship relationship")
-        RangeJoinExec(planLater(planA), planLater(planB), a, b, spatialPredicate, extraCondition) :: Nil
+        RangeJoinExec(planLater(planA), planLater(planB), a, b, swappedLeftAndRight, spatialPredicate, extraCondition) :: Nil
```
Kimahriman, line 169: Can these just be handled like the broadcast join below where the swapping is handled here with pattern matching instead of pushed into the Exec node?

Kontinuation: Exec node has to handle the swapping somewhere, unless we invert the spatial predicate in the join query detector.

Broadcast join exec node [resolves if left and right were swapped](https://github.com/apache/incubator-sedona/blob/master/sql/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/BroadcastIndexJoinExec.scala#L117-L122) using `indexBuildSide` and `windowJoinSide`, it does not seems to be more intuitive than simply pushing the swapping indicator to the `TraitJoinQueryExec` node.

Inverting the spatial predicate seems to be a viable approach, I'll submit changes to this PR if you're OK with it.

sql/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/TraitJoinQueryExec.scala
jiayuasu: Exactly, we no longer need the SPARK2 anchor.

sql/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/JoinQueryDetector.scala
jiayuasu: @Kimahriman Adam, does @Kontinuation 's approach sound good to you? :-)

Kimahriman: I'm trying to wrap my head around this a little bit. By inverting the spatial predicate do you mean like rewritting it? As in `ST_Contains(a, b)` -> `ST_Within(b, a)` or something like that?

What would happen if you just swapped the left and right shapes if needed? Like
```scala
    matchExpressionsToPlans(a, b, left, right) match {
      case Some((_, _, swapped)) =>
        logInfo(s"Planning spatial join for $relationship relationship")
        if (swapped) {
          RangeJoinExec(planLater(left), planLater(right), b, a, spatialPredicate, extraCondition) :: Nil
        } else {
          RangeJoinExec(planLater(left), planLater(right), a, b, spatialPredicate, extraCondition) :: Nil
        }
      case None =>
        logInfo(
          s"Spatial join for $relationship with arguments not aligned " +
            "with join relations is not supported")
        Nil
    }
```

I'm trying to think of why that wouldn't work

jiayuasu:
@Kontinuation I will merge this PR for now since we will release 1.3.0 in the next few days. Maybe we can address @Kimahriman 's comment in another PR.

jiayuasu (approved):

PR #711, opened 2022-11-11
[SEDONA-191] Support join types other than inner in BroadcastIndexJoinExec

tanelk:

- Yes, the URL of the assoicated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-191. The PR name follows the format `[SEDONA-XXX] my subject`.

Extended the `BroadcastIndexJoinExec` to support left semi, left anti, left outer and right outer joins if there is a broadcast hint on correct side. Currently those joins are run with Sparks `BroadcastNestedLoopJoinExec`, that can be much slower.

A lot of the code is based on `org.apache.spark.sql.execution.joins.HashJoin`.

## How was this patch tested?
New unittests


tanelk:
This requires more testing before it can be merged. Just wanted to get the draft up.

sql/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/JoinQueryDetector.scala
```diff
@@ -206,14 +223,32 @@ class JoinQueryDetector(sparkSession: SparkSession) extends Strategy {
...
+    children: Seq[Expression],
+    joinType: JoinType,
+    spatialPredicate: SpatialPredicate,
+    indexType: IndexType,
+    broadcastLeft: Boolean,
+    broadcastRight: Boolean,
+    extraCondition: Option[Expression],
+    distance: Option[Expression]): Seq[SparkPlan] = {
+
+    val broadcastSide = joinType match {
```
tanelk, line 238: I'll cover `LeftExistence` in a followup - it's more tedious to implement and test.

tanelk:
@Kimahriman you introduced broadcast joins here, perhaps you are the best person to take a look at this improvement.
Also pinging @jiayuasu - I think this is ready for review now

jiayuasu:
@Kimahriman Adam, would you please share some insights about this?

Kimahriman:
Yeah I peaked at this when it first was put up, I'll look through again with the latest updates

sql/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/JoinQueryDetector.scala
```diff
@@ -206,14 +223,32 @@ class JoinQueryDetector(sparkSession: SparkSession) extends Strategy {
...
+      case Inner if broadcastLeft => Some(LeftSide)
+      case Inner if broadcastRight => Some(RightSide)
+      case LeftSemi if broadcastRight => Some(RightSide)
+      case LeftAnti if broadcastRight => Some(RightSide)
+      case LeftOuter if broadcastRight => Some(RightSide)
+      case RightOuter if broadcastLeft => Some(LeftSide)
```
Kimahriman, line 244: What's the limitation of doing the other jointype/broadcast side combinations? I understand maybe less utility in them, but does enabling them add complexity?

tanelk: I could not figure out how to easily implement other combinations

Kimahriman: Ah yeah I see now how the outer logic only works in one direction

Kimahriman:
Looks good to me. Only other comment would be if the tests could be parameterized at all so to dedupe some of it, and maybe verify the schema and at least one row of data to make sure the output is correct and we don't run into a similar problem as https://github.com/apache/incubator-sedona/pull/706

tanelk:
> Looks good to me. Only other comment would be if the tests could be parameterized at all so to dedupe some of it, and maybe verify the schema and at least one row of data to make sure the output is correct and we don't run into a similar problem as #706

 I'll improve the test coverage in few days

tanelk:
Added additional tests in https://github.com/apache/incubator-sedona/pull/711/commits/2624db496d34621c5e55ad287752adec7564f12e

jiayuasu (changes requested):
Just a minor thing. Would you please update the doc [1] to cover the content introduced in this PR? 

[1] https://sedona.apache.org/api/sql/Optimizer/#broadcast-join

Kimahriman (approved):

jiayuasu (approved):

jiayuasu:
Thanks for your contribution!

michelelaporta:
thanks [petern48](https://github.com/petern48) any example usage?  I post a discussion with a sample here https://github.com/apache/sedona/discussions/2520 by nobody has answer yet.

jiayuasu:
Sorry just replied

PR #739, opened 2022-12-28
[SEDONA-207] Implemented a new geometry serde for supporting Z/M dimensions and better performance

Kontinuation:

- Yes, the URL of the assoicated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-207. The PR name follows the format `[SEDONA-XXX] my subject`.

### New Geometry Serde

The new geometry serde was implemented in `common/src/main/java/org/apache/sedona/common/geometrySerde/`. The ShapeSerde used by Kryo serializer and the WKB based serde used by `GeometryUDT` were replaced by this new serde. Please refer to [SEDONA-207](https://issues.apache.org/jira/browse/SEDONA-207) for a detailed explanation of this new geometry serde.

### GeoParquet

GeoParquet stores geometry objects as WKB binary values, which happens to be the old serialization format of `GeometryUDT` thus no special treatment was needed. This PR changed the serialization format of `GeometryUDT`, so geometry values in GeoParquet files need to be explicitly parsed and serialized.

### GeometryUDT in Python

We've implemented the new serialization format in pure python, it is 2~3x slower than `shapely.wkb.loads/dumps`, which would impact the performance of `collect`, `toPandas` and Python UDFs in pyspark. We'll explore ways to implement it as a CPython extension to achieve good performance.

Python version of the geometry serde does not support M dimension and SRID due to the restrictions of shapely. We'll support M dimension once https://github.com/shapely/shapely/issues/1648 was resolved.

## How was this patch tested?

Unit tests were added to test this patch. This patch was also manually tested on a Spark standalone cluster.

The geometry serde code for Python was manually tested with shapely 2.0. We need to update the python unit tests to be compatible with shapely 2.0 in the future.

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/geometrySerde/UnsafeGeometryBuffer.java
+++ b/common/src/main/java/org/apache/sedona/common/geometrySerde/UnsafeGeometryBuffer.java
@@ -19,39 +19,19 @@
 
 package org.apache.sedona.common.geometrySerde;
 
-import java.lang.reflect.Field;
-
+import org.apache.spark.unsafe.Platform;
 import org.locationtech.jts.geom.Coordinate;
 import org.locationtech.jts.geom.CoordinateSequence;
 import org.locationtech.jts.geom.CoordinateXY;
 import org.locationtech.jts.geom.CoordinateXYM;
 import org.locationtech.jts.geom.CoordinateXYZM;
 import org.locationtech.jts.geom.impl.CoordinateArraySequence;
-import sun.misc.Unsafe;
 
 class UnsafeGeometryBuffer implements GeometryBuffer {
-
-    private static final Unsafe UNSAFE;
-    private static final long BYTE_ARRAY_BASE_OFFSET;
-
-    static {
-        Unsafe unsafe;
-        long byteArrayOffset = 0;
-        try {
-            Field theUnsafe = Unsafe.class.getDeclaredField("theUnsafe");
-            theUnsafe.setAccessible(true);
-            unsafe = (Unsafe) theUnsafe.get(null);
-            byteArrayOffset = unsafe.arrayBaseOffset(byte[].class);
-        } catch (IllegalAccessException | NoSuchFieldException | SecurityException e) {
-            // Unsafe is not available
-            unsafe = null;
-        }
-        UNSAFE = unsafe;
-        BYTE_ARRAY_BASE_OFFSET = byteArrayOffset;
-    }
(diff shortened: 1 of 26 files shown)


Kontinuation:
Example project build failed since the [snapshot artifact for Sedona 1.3.1-incubating](https://repository.apache.org/content/groups/snapshots/org/apache/sedona/sedona-viz-3.0_2.12/1.3.1-incubating-SNAPSHOT/) was gone. Do we need to switch the dependency to a released version?

jiayuasu:
@Kontinuation I just switched all versions to 1.3.1-incubating. If you pull the latest change, you should be fine.

A quick question: does this solve [SEDONA-222](https://issues.apache.org/jira/browse/SEDONA-222)?

Kontinuation:
> @Kontinuation I just switched all versions to 1.3.1-incubating. If you pull the latest change, you should be fine.
> 
> A quick question: does this solve [SEDONA-222](https://issues.apache.org/jira/browse/SEDONA-222)?

It does not resolve SEDONA-222. I'm working on another branch to resolve issues related to GeoParquet, and this issue will be resolved in an upcoming PR.

common/src/main/java/org/apache/sedona/common/geometrySerde/UnsafeGeometryBuffer.java
```diff
@@ -0,0 +1,300 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.sedona.common.geometrySerde;
+
+import org.apache.spark.unsafe.Platform;
```
Kimahriman, line 22: Noticed my build started failing for https://github.com/apache/incubator-sedona/pull/735 because I guess it uses a merged codebase to run off of. The common module is specifically not supposed to have any reference to Spark because:
- It's not versioned by the Scala release
- It's used by Flink as well so this won't even be available

This will need to be removed, either by porting any of the necessary code here directly, or moving this to the core module and making at least the UnsafeGeometryBuffer (or just the whole serializer) a Spark only thing for the time being

Kontinuation: https://github.com/apache/incubator-sedona/pull/741 was submitted to resolve this issue, where `spark.unsafe.Platform` was replaced by direct references to `sun.misc.Unsafe`.

PR #742, opened 2023-01-01
[SEDONA-223] Add ST_Split

douglasdennis:

- Yes

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-223. The PR name follows the format `[SEDONA-XXX] my subject`.

1. Add ST_Split function to sedona-common, sedona-sql, sedona-flink, and the dataframe/python API.
2. Created FunctionsTest.java to house tests for functions being added to sedona-common.

## How was this patch tested?

1. Comprehensive unit tests in sedona-common
2. Integration unit tests in sedona-sql, sedona-flink, and python.

- Yes, I have updated the documentation in docs/api/sql/Function.md.

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/utils/GeometrySplitter.java
+++ b/common/src/main/java/org/apache/sedona/common/utils/GeometrySplitter.java
@@ -25,15 +25,12 @@ import org.locationtech.jts.geom.Coordinate;
 import org.locationtech.jts.geom.Geometry;
 import org.locationtech.jts.geom.GeometryCollection;
 import org.locationtech.jts.geom.GeometryFactory;
-import org.locationtech.jts.geom.Lineal;
 import org.locationtech.jts.geom.LineString;
 import org.locationtech.jts.geom.MultiLineString;
 import org.locationtech.jts.geom.MultiPoint;
 import org.locationtech.jts.geom.MultiPolygon;
 import org.locationtech.jts.geom.Point;
 import org.locationtech.jts.geom.Polygon;
-import org.locationtech.jts.geom.Polygonal;
-import org.locationtech.jts.geom.Puntal;
 import org.locationtech.jts.linearref.LinearGeometryBuilder;
 import org.locationtech.jts.operation.polygonize.Polygonizer;
 
--- a/common/src/main/java/org/apache/sedona/common/utils/GeometrySplitter.java
+++ b/common/src/main/java/org/apache/sedona/common/utils/GeometrySplitter.java
@@ -1,3 +1,16 @@
+/**
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.apache.sedona.common.utils;
 
 import java.util.ArrayDeque;
--- a/common/src/main/java/org/apache/sedona/common/utils/GeometrySplitter.java
+++ b/common/src/main/java/org/apache/sedona/common/utils/GeometrySplitter.java
@@ -20,6 +20,7 @@ import java.util.Deque;
 import java.util.Iterator;
 import java.util.List;
 
+import org.apache.log4j.Logger;
 import org.locationtech.jts.geom.Coordinate;
 import org.locationtech.jts.geom.Geometry;
 import org.locationtech.jts.geom.GeometryCollection;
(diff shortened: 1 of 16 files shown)


douglasdennis:
Ugh. I noticed that my VS Code seems to have gone wild on deleting leading white spaces. I hope that is ok. If the edits seem too messy though then I can try to revert them.

common/src/main/java/org/apache/sedona/common/utils/GeometrySplitter.java
```diff
@@ -0,0 +1,321 @@
...
+        // compute the intersection of inputLines and blade
+        // and pass back to splitLines to handle as points
+        Geometry intersectionWithBlade = inputLines.intersection(blade);
+
+        if (intersectionWithBlade.isEmpty()) {
+            // blade and inputLines are disjoint so just return the input as a multilinestring
+            return (MultiLineString)ensureMultiGeometryOfDimensionN(inputLines, 1);
+        } else if (intersectionWithBlade.getDimension() != 0) {
+            // FIXME need way to access the logger
+            System.err.println("Colinear sections detected between source and blade geometry. Returned null.");
```
douglasdennis, line 154: I wasn't sure of a good way to access a log in sedona-common so I defaulted to standard error out. Is a logging mechanism something that needs to be added? @jiayuasu mentioned `Log.error` in the Jira ticket but I couldn't find the `Log` class/object. I'm happy to change this prior to merge if someone has a better way.

jiayuasu: You can use log4j for that. See an example here: https://github.com/apache/incubator-sedona/blob/master/core/src/main/java/org/apache/sedona/core/joinJudgement/JudgementBase.java#L120

common/src/main/java/org/apache/sedona/common/utils/GeometrySplitter.java
```diff
@@ -0,0 +1,321 @@
+package org.apache.sedona.common.utils;
```
jiayuasu, line 1: Please make sure all new files have Apache License 2.0 header

common/src/main/java/org/apache/sedona/common/utils/GeometrySplitter.java
Kimahriman: I was looking into standardizing the logging dependencies a little bit as a follow on to my last PR, can submit one for that soon. Right now log4j is available when spark is included, but not in common or flink

jiayuasu: Maybe we don't need to include log4j in the common module. Instead, functions in the common module only throw exceptions and sedona-flink / sedona-spark catch them and produce `log.error` using log4j?

douglasdennis: Maybe a new type: `SedonaFunctionException`. With a constructor like: `SedonaFunctionException(String message, Object result)`. Where `message` gets written to the appropriate log and `result` is the result of the function? I feel like there may sometimes be situations where a log entry needs to be made but a non-null result is still viable.

common/src/main/java/org/apache/sedona/common/utils/GeometrySplitter.java
douglasdennis: That file did look a little funny at the top. Added it.

common/src/main/java/org/apache/sedona/common/utils/GeometrySplitter.java
Kimahriman: I made https://github.com/apache/incubator-sedona/pull/743 if we want to just be able to log in the common module

Kimahriman: And if you're going to add logging it might be a good chance to start using slf4j instead of log4j directly

jiayuasu:
@douglasdennis PR https://github.com/apache/sedona/pull/743 has been merged. If you pull from the upstream, you can actually now use log.error to handle the issue. Adam might need to make some changes later on but the dependencies should be the same.

common/src/main/java/org/apache/sedona/common/utils/GeometrySplitter.java
```diff
@@ -0,0 +1,331 @@
...
+    private MultiLineString splitLinesByLines(Geometry inputLines, Geometry blade) {
+        // compute the intersection of inputLines and blade
+        // and pass back to splitLines to handle as points
+        Geometry intersectionWithBlade = inputLines.intersection(blade);
+
+        if (intersectionWithBlade.isEmpty()) {
+            // blade and inputLines are disjoint so just return the input as a multilinestring
+            return (MultiLineString)ensureMultiGeometryOfDimensionN(inputLines, 1);
+        } else if (intersectionWithBlade.getDimension() != 0) {
+            Logger.getLogger(GeometrySplitter.class).error("Colinear sections detected between source and blade geometry. Returned null.");
```
Kimahriman, line 164: 3 things:
Can you try changing the common dependency from
```
        <dependency>
            <groupId>org.apache.logging.log4j</groupId>
            <artifactId>log4j-1.2-api</artifactId>
        </dependency>
```
to
```
            <dependency>
                <groupId>org.slf4j</groupId>
                <artifactId>slf4j-api</artifactId>
            </dependency>
```
and try using the slf4j logger API instead? I think doing that in the common module would be a good place to try to stop using the log4j 1.x API directly.

Create the Logger as a static variable at the top of the class, I think that's the recommended way.

And kinda just my opinion, but if it doesn't break/stop the execution, I feel like it should be.a warning message and not an error message

douglasdennis: @Kimahriman Double check what I just pushed, please :)

I followed [this](https://www.slf4j.org/api/org/slf4j/Logger.html) which seemed to recommend using this `LoggerFactory` class as well. I don't mind iterating on this till you're good with it.

Kimahriman: Yep that's exactly what I was picturing!

jiayuasu:
Awesome. Will merge.

jiayuasu (approved):

PR #748, opened 2023-01-18
[SEDONA-233] Incorrect results for several joins in a single stage

umartin:

- Yes, the URL of the assoicated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-233. The PR name follows the format `[SEDONA-XXX] my subject`.

This patch changes how the deduplication gets it partition id. The previous method of getting it from TaskContext was unreliable. Now it uses mapPartitionsWithIndex. The documentation clearly states that is uses the _original_ partition id. https://spark.apache.org/docs/latest/api/scala/org/apache/spark/rdd/RDD.html#mapPartitionsWithIndex[U](f:(Int,Iterator[T])=%3EIterator[U],preservesPartitioning:Boolean)(implicitevidence$9:scala.reflect.ClassTag[U]):org.apache.spark.rdd.RDD[U]

Deduplication is refactored out of the join judgement into a separate DuplicatesFilter.

Deduplication code that is used in sedona-flink is moved to common.

## How was this patch tested?

Unit test added


core/src/main/java/org/apache/sedona/core/joinJudgement/DuplicatesFilter.java
```diff
@@ -0,0 +1,48 @@
+package org.apache.sedona.core.joinJudgement;
+
```
jiayuasu, line 2: Please add Apache License header here.

core/src/test/java/org/apache/sedona/core/spatialOperator/JoinQueryDeduplicationTest.java
```diff
@@ -0,0 +1,75 @@
+/*
```
jiayuasu, line 1: Could you please add another test in Sedona Spark SQL to verify that this bug is eliminated?

jiayuasu (changes requested):

core/src/main/java/org/apache/sedona/core/joinJudgement/DuplicatesFilter.java
umartin: Yes

core/src/test/java/org/apache/sedona/core/spatialOperator/JoinQueryDeduplicationTest.java
umartin: Sure! Hopefully I have time to add it tomorrow

jiayuasu (approved):

jiayuasu:
@yitao-li 

Dear Yitao, Sedona R build started to fail since this PR. But the PR is not relevant to the R side, could you please take a look?

Thanks,
Jia

yitao-li:
Hello Jia,
    Apologies for the delayed reply! I'll take a look as soon as I can.
    I don't have a lot of bandwidth at the moment though, mainly busy with
work, and then helping take care of my 6-month-old before and after work,
among other things... I wonder whether there are other folks such as the
current maintainer of sparklyr who might be able to help?
    Also, while we are on the topic of sparklyr: TBH 2 things that always
weighed on my mind while working with sparklyr and sparklyr-related R
packages are (1) sparklyr uses java reflection to invoke JVM methods and it
is so easy to break abstraction with it if one is not careful, and then (2)
it might also run into subtle interop issues with Scala. I don't know
whether there is a great solution to any of those 2 potential issues at the
moment though.
    Anyhow, I'll look into this issue as soon as I can.
    Best regards,
Yitao

PR #770, opened 2023-02-16
[SEDONA-243] R features: read/write geoparquet, get names from shapefiles

gregleleu:
Updates to the R package:
* reading/writing geoparquet files 
* getting names to sdf from cases where "fieldNames" exists (shapefile)

## How was this patch tested?
Added/updated the relevant R tests

Updated the R documentation only


R/R/data_interface.R
```diff
@@ -439,6 +439,52 @@ sedona_read_shapefile <- function(sc,
...
+#'
+#' if (!inherits(sc, "test_connection")) {
+#'   input_location <- "/dev/null" # replace it with the path to your input file
+#'   rdd <- sedona_read_geoparquet(sc, location = input_location)
+#' }
+#'
+#' @family Sedona data interface functions
+#'
+#' @export
+sedona_read_geoparquet <- function(sc,
```
jiayuasu, line 467: Can we add a function that takes any format string? This way, the R interface will be able to automatically invoke any new reader we add in the future?

The example usage could be:

Read GeoParquet

```
sdf = sedona_read(sc, "geoparquet", location, name)
```

Read GeoTiff

```
sdf = sedona_read(sc, "geotiff", location, name)
```

R/R/data_interface.R
```diff
@@ -589,6 +635,52 @@ sedona_save_spatial_rdd <- function(x,
...
+#'     tbl %>% dplyr::mutate(id = 1),
+#'     output_location = "/tmp/pts.geoparquet"
+#'   )
+#' }
+#'
+#' @family Sedona data interface functions
+#'
+#' @export
+sedona_save_geoparquet <- function(x,
+                                   output_location) {
```
jiayuasu, line 671: Same here:

We could make this save function more generic:

```
sedona_save(df, "geoparqet", location)
```

jiayuasu (changes requested):

gregleleu:
I created an account, and wanted to create an issue, but JIRA is so complicated I gave up. I'll try again next time.

Regarding the "generic" function, it's not really the R way, cf. sparklyr's documentation (https://spark.rstudio.com/packages/sparklyr/latest/reference/)

Any idea why the checks failed? I'm using the same test as the scala version – an exception with the words "GeoParquet file does not contain valid geo metadata" – but it raised another kind of exception for spark 3.0.3

jiayuasu:
@gregleleu 

Spark Analysis exception has been there for a while: https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/AnalysisException.scala

Not sure why it failed. In fact, we have the same Scala tests on Spark 3.0.3 and it can successfully capture the exception. @Kontinuation Hi Kristin, any idea why this AnalysisException was not found in Spark 3.0.3?

R/R/data_interface.R
jiayuasu: Thanks for the explanation. Makes sense to me. Would you please add one more reader/writer `sedona_read_geotiff` and `sedona_write_geotiff`?

It works the same way as the geoparquet: https://sedona.apache.org/latest-snapshot/api/sql/Raster-loader/#geotiff-dataframe-loader.

If you think this is too complicated, we certainly can leave it to another PR.

Kontinuation:
> @gregleleu
> 
> Spark Analysis exception has been there for a while: https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/AnalysisException.scala
> 
> Not sure why it failed. In fact, we have the same Scala tests on Spark 3.0.3 and it can successfully capture the exception. @Kontinuation Hi Kristin, any idea why this AnalysisException was not found in Spark 3.0.3?

`AnalysisException` has some ABI-breaking changes after 3.0, so the code compiled with spark 3.3 could break on spark 3.0. Can we simply replace them with `IllegalArgumentException`?

jiayuasu:
@Kontinuation Yes, let's change it to `IllegalArgumentException`. Can you make a PR to fix it?

Kontinuation:
> Do we keep the informative message ("GeoParquet file does not contain valid geo metadata") if we change the class of the exception?

Yes, the message will be kept.

Kontinuation:
> @Kontinuation Yes, let's change it to `IllegalArgumentException`. Can you make a PR to fix it?

Submitted https://github.com/apache/sedona/pull/771

jiayuasu:
@gregleleu #771 has been merged. Please pull from the upstream. This will trigger GitHub actions of this PR.

gregleleu:
@jiayuasu Made a few changes, notably changed the name of the write function to be coherent with the rest of the sparklyr ecosystem.

R/R/data_interface.R
gregleleu: Regarding the Geotiff: I started working on it. Reading the geotiff works fine.
There is one complexity given the created dataframe is a nested object. There is a package to help with that case: `sparklyr.nested`. I'll need more time to figure out whether we need to list it as a dependency or just suggest its use. And if it is required for the testing.

jiayuasu (approved):

PR #783, opened 2023-03-02
[SEDONA-250][SEDONA-253] R - Update dependencies + drop 2.4

gregleleu:
https://issues.apache.org/jira/projects/SEDONA/issues/SEDONA-250?

Update dependencies to 28.2
Remove support for Spark 2.4

## How was this patch tested?
Ran the tests


gregleleu:
Unclear why the tests crashed, maybe again an issue with the cache since we changed dependencies?

jiayuasu:
@gregleleu I have cleaned up the cache and re-ran the test. The test still failed because it cannot find resolve-pom.xml in spark-shaded. Like we discussed in the JIRA ticket, you might want to use regex to locate the jar instead of relying on resolve-pom.

jiayuasu (approved):

PR #791, opened 2023-03-08
[SEDONA-229] Fix Linter Issues in Viz

douglasdennis:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-229.

This PR addresses several linter issues in the sedona viz module. All of the fixes for these issues involve no real structural code changes, just formatting and being more explicit with imports. The types of changes in this PR:

- Ignore imports from AWT, which is apparently an "illegal" import according to the linter. These seem mandatory for the viz module.
- Wildcard imports were replaced with explicit imports for the classes needed.
- An import was moved from inside of a method to be at the top.
- Add spaces around a comment and plus operators.
- Fixed long lines. I attempted to use the style guide located at https://docs.scala-lang.org/style/ If I got something wrong then please let me know.
- Added newlines to the ends of files.

## How was this patch tested?

Standard build process with unit tests.


jiayuasu:
@douglasdennis This is perfectly fine. However, I wonder if scala has something similar to https://code.revelc.net/formatter-maven-plugin/ in the Java world. This will automatically fix styling issues instead of manually doing the job.

jiayuasu (approved):

douglasdennis:
@jiayuasu I found scalafmt (https://scalameta.org/scalafmt/). It looks like spotless (https://github.com/diffplug/spotless/tree/main/plugin-maven) uses it to handle Scala code as a maven plugin. I can explore that for handling the formatting related linter messages.

 Would there be any interest to add this as a check to the build process?

jiayuasu:
@douglasdennis I think spotless is nice. Let's have it in Sedona. I think we should put it in the build process. Should we allow the build process automatically fix the format, or let the build process fail and we manually call `mvn spotless:apply` to fix it?

I kind of prefer the auto fix.

PR #835, opened 2023-05-20
[SEDONA-281] Support geodesic / geography functions

jiayuasu:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-281. The PR name follows the format `[SEDONA-281] my subject`.

### ST_AreaSpheroid

Return the geodesic area of A using WGS84 spheroid. Unit is meter. Works better for large geometries (country level) compared to `ST_Area` + `ST_Transform`. It is equivalent to PostGIS `ST_Area(geography, use_spheroid=true)` function and produces nearly identical results.

### ST_LengthSpheroid

Return the geodesic perimeter of A using WGS84 spheroid. Unit is meter. Works better for large geometries (country level) compared to `ST_Length` + `ST_Transform`. It is equivalent to PostGIS `ST_Length(geography, use_spheroid=true)` and `ST_LengthSpheroid` function and produces nearly identical results.

### ST_DistanceSphere

Return the haversine / great-circle distance of A using a given earth radius (default radius: 6378137.0). Unit is meter. Works better for large geometries (country level) compared to `ST_Distance` + `ST_Transform`. It is equivalent to PostGIS `ST_Distance(geography, use_spheroid=false)` and `ST_DistanceSphere` function and produces nearly identical results. It provides faster but less accurate result compared to `ST_DistanceSpheroid`.

### ST_DistanceSpheroid

Return the geodesic distance of A using WGS84 spheroid. Unit is meter. Works better for large geometries (country level) compared to `ST_Distance` + `ST_Transform`. It is equivalent to PostGIS `ST_Distance(geography, use_spheroid=true)` and `ST_DistanceSpheroid` function and produces nearly identical results. It provides slower but more accurate result compared to `ST_DistanceSphere`.

## How was this patch tested?

Added many unit tests.

- Yes, I am adding a new API. I am using the [current SNAPSHOT version number](https://github.com/apache/sedona/blob/master/pom.xml#L29) in since `vX.Y.Z` format.


common/src/main/java/org/apache/sedona/common/sphere/Haversine.java
```diff
@@ -0,0 +1,65 @@
...
+                * sin(lngDistance / 2) * sin(lngDistance / 2);
+        double c = 2 * atan2(sqrt(a), sqrt(1 - a));
+        return AVG_EARTH_RADIUS * c * 1.0;
+    }
+
+    // Calculate the distance between two points on the earth using the "haversine" formula.
+    // The radius of the earth is 6371.0 km
+    public static double distance(Geometry geom1, Geometry geom2)
+    {
+        return distance(geom1, geom2, 6378137.0);
```
Kontinuation, line 63: I suggest using `6371008` as the default radius according to [PostGIS ST_DistanceSphere](https://postgis.net/docs/ST_DistanceSphere.html) and https://en.wikipedia.org/wiki/Earth_radius

PR #845, opened 2023-05-29
[SEDONA-286] Support optimized distance join on ST_DistanceSphere and ST_DistanceSpheroid

jiayuasu:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-286. The PR name follows the format `[SEDONA-286] my subject`.

1. Added patterns for ST_DistanceSpheroid and ST_DistanceSphere based join
2. Expand buffer area by following the formula provided by Kristin:

$$
\begin{align*}
\Delta \textrm{lat} &= \frac{D}{R} \\
\Delta \textrm{lon} &= \max (\frac{D}{R \cos(\textrm{lat} + \Delta \textrm{lat})}, \frac{D}{R \cos(\textrm{lat} - \Delta \textrm{lat})})
\end{align*}
$$

where $\textrm{lat}$, $\textrm{lon}$ is the center of the envelope; $R$ is the mean radius of the earth. The center point is expanded by $\Delta \textrm{lat}$, $\Delta \textrm{lon}$ to obtain the envelope for coarse-grained filtering. The envelope is expanded by a constant factor `1.1` to get rid of false negatives.

For spheroid and sphere distance, we use the same formula with the same $R$ value. We take the polar radius of the spheroid as $R$ (approximately 6357 km) and the resulting envelope should be inclusive.

Now the users can trigger our advanced spatial join algorithm (distance join and broadcast index join) by a ST_DistanceSpheroid and ST_DistanceSphere predicate.

```
SELECT *
FROM df1, df2
WHERE ST_DistanceSpheroid(df1.geom, df2.geom) < 100
``` 

`100` indicates 100 meters.

## How was this patch tested?

Passed new unit tests

### DIFF
--- a/docs/api/sql/Function.md
+++ b/docs/api/sql/Function.md
@@ -51,7 +51,7 @@ FROM polygondf
 
 ## ST_AreaSpheroid
 
-Introduction: Return the geodesic area of A using WGS84 spheroid. Unit is meter. Works better for large geometries (country level) compared to `ST_Area` + `ST_Transform`. It is equivalent to PostGIS `ST_Area(geography, use_spheroid=true)` function and produces nearly identical results.
+Introduction: Return the geodesic area of A using WGS84 spheroid. Unit is square meter. Works better for large geometries (country level) compared to `ST_Area` + `ST_Transform`. It is equivalent to PostGIS `ST_Area(geography, use_spheroid=true)` function and produces nearly identical results.
 
 Geometry must be in EPSG:4326 (WGS84) projection and must be in ==lat/lon== order. You can use ==ST_FlipCoordinates== to swap lat and lon.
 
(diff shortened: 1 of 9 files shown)


Kontinuation:
The transformation from sphere/spheroid distance to radius distance may not be an inclusive transformation, which means that $\forall (p_0, p_1, d) [ \textrm{DistanceSphere}(p_0, p_1) < d \implies \textrm{Distance}(p_0, p_1) < \frac{d}{111000} ]$ does not hold. Applying the optimization may result in fewer results being produced. If this is the case, we have to document that such queries may not produce accurate results, especially for geometries near the poles.

jiayuasu:
@Kontinuation Thanks. Is there a better solution than this? More accurate but also easy to implement?

Kontinuation:
We can derive a more accurate formula of lat-lon envelope for haversine distance $D$:

$$
\begin{align*}
\Delta \textrm{lat} &= \frac{D}{R} \\
\Delta \textrm{lon} &= \max (\frac{D}{R \cos(\textrm{lat} + \Delta \textrm{lat})}, \frac{D}{R \cos(\textrm{lat} - \Delta \textrm{lat})})
\end{align*}
$$

where $\textrm{lat}$, $\textrm{lon}$ is the center of the envelope; $R$ is the radius of the earth. The center point is expanded by $\Delta \textrm{lat}$, $\Delta \textrm{lon}$ to obtain the envelope for coarse-grained filtering. The envelope can also be expanded by a constant factor slightly greater than 1 to get rid of false negatives.

For spheroid distance, we can use the same formula for sphere distance, but with a different $R$ value. We can take the equatorial radius of the spheroid as $R$ (approximately 6378 km) and the resulting envelope should be inclusive.

I'm not sure if I'm doing the math correctly, I can find some other references on this topic.

Kontinuation (approved):

PR #862, opened 2023-06-13
[SEDONA-302] Add ST_Translate | [SEDONA-196] Update tests for ST_Force3D

iGN5117:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-302 and https://issues.apache.org/jira/browse/SEDONA-196. The PR name follows the format `[SEDONA-XXX] my subject`.

- Implement ST_Translate and port it to Sedona flink and spark
- Update test cases for ST_Force3D

## How was this patch tested?
- Comprehensive correctness tests in java
- Integration tests

- Yes, I am adding a new API. I am using the 1.4.1 (https://github.com/apache/sedona/blob/master/pom.xml#L29) in since `v1.4.1` format.


common/src/main/java/org/apache/sedona/common/Functions.java
```diff
@@ -881,6 +881,38 @@ public static Integer nRings(Geometry geometry) throws Exception {
         return numRings;
     }
 
+    public static Geometry translate(Geometry geometry, double deltaX, double deltaY, double deltaZ) {
+        if (!geometry.isEmpty()) {
+            if (geometry instanceof GeometryCollection) {
+                GeometryCollection geometryCollection = (GeometryCollection) geometry;
```
jiayuasu, line 887: GeometryCollection can have child GeometryCollection. The process to extract GeometryCollection is more complicated than you thought. See here: https://github.com/apache/sedona/blob/master/common/src/main/java/org/apache/sedona/common/utils/GeomUtils.java#L411

Please fix it accordingly. We usually use a queue or recursive function to implement similar logics. In this case, I would recommend a recursive function. With such a function, the code could be even further simplified.

common/src/test/java/org/apache/sedona/common/FunctionsTest.java
```diff
@@ -731,4 +731,87 @@ public void nRingsUnsupported() {
         assertEquals(expected, e.getMessage());
     }
 
+    @Test
```
jiayuasu, line 734: Please also create test cases that have GEOMETRYCOLLECTION(MULTILPOLYGON(), GEOMETRYCOLLECTION())

docs/api/flink/Function.md
```diff
@@ -993,6 +993,31 @@ FROM polygondf
...
     The detailed EPSG information can be searched on [EPSG.io](https://epsg.io/).
 
+## ST_Translate
+Introduction: Returns the input geometry with its X, Y and Z coordinates (if present in the geometry) translated by deltaX, deltaY and deltaZ (if specified)
+
+If the geometry is 2D, and a deltaZ parameter is specified, no change is done to the Z coordinate of the geometry and the resultant geometry is also 2D.
+
+If the geometry is empty, no change is done to it.
+
+If a geometry collection is given, all geometries of the collection are individually translated.
```
jiayuasu, line 1003: This should apply to all geometries such as multipoint, multipolygon, multilinestring, and geometrycollection. Not only geometrycollection.

sql/common/src/test/scala/org/apache/sedona/sql/dataFrameAPITestScala.scala
```diff
@@ -978,5 +978,14 @@ class dataFrameAPITestScala extends TestBaseScala {
       val actual = df.take(1)(0).getInt(0)
       assert(expected == actual)
     }
+
+    it("Passed ST_Translate") {
```
jiayuasu, line 982: Please add a test for the case that has no optional parameter.

sql/common/src/test/scala/org/apache/sedona/sql/functionTestScala.scala
```diff
@@ -1954,4 +1954,22 @@ class functionTestScala extends TestBaseScala with Matchers with GeometrySample
...
+
+  it ("should pass ST_Translate") {
+    val geomTestCases = Map(
+      ("'POINT (1 1 1)'") -> ("'POINT Z(2 2 2)'", "'POINT Z(2 2 1)'"),
+      ("'POLYGON ((1 0, 1 1, 2 1, 2 0, 1 0))'") -> ("'POLYGON ((2 1, 2 2, 3 2, 3 1, 2 1))'", "'POLYGON ((2 1, 2 2, 3 2, 3 1, 2 1))'"),
+      ("'LINESTRING EMPTY'") -> ("'LINESTRING EMPTY'", "'LINESTRING EMPTY'")
+    )
+    for (((geom), expectedResult) <- geomTestCases) {
+      val df = sparkSession.sql(s"SELECT ST_AsText(ST_Translate(ST_GeomFromWKT($geom), 1, 1, 1)) AS geom, " + s"$expectedResult")
+      val dfDefaultValue = sparkSession.sql(s"SELECT ST_AsText(ST_Translate(ST_GeomFromWKT($geom), 1, 1)) AS geom, " + s"$expectedResult")
```
jiayuasu, line 1966: Please add a test for the case that has no optional parameter.

jiayuasu (changes requested):

common/src/main/java/org/apache/sedona/common/Functions.java
jiayuasu: On this regard, I think the approach used in ST_Force3D are actually better since it does not need to fetch individual geometries.

In addition, I noticed that ST_Force3D does not have test cases for GeomtryCollection/MultiPolygon and such. Would you please confirm ST_Force3D works for these cases as well?

sql/common/src/test/scala/org/apache/sedona/sql/functionTestScala.scala
iGN5117: This already has tests for cases with no optional parameter in df

common/src/test/java/org/apache/sedona/common/FunctionsTest.java
```diff
@@ -731,4 +771,90 @@ public void nRingsUnsupported() {
...
+    public void translateEmptyObjectNoDeltaZ() {
```
jiayuasu, line 775: Please change the test cases to use different values for deltaX and deltaY (currently they are all `1`) or add a test case that uses different values for detlaX and deltaY.

This will test if x and y are handled correctly.

jiayuasu (changes requested):

common/src/test/java/org/apache/sedona/common/FunctionsTest.java
iGN5117: Done

sql/common/src/test/scala/org/apache/sedona/sql/dataFrameAPITestScala.scala
iGN5117: Done

common/src/main/java/org/apache/sedona/common/Functions.java
iGN5117: Fixed ST_Translate to handle recursive geometries in a better way.
Updated test cases for ST_Force3D

jiayuasu (approved):

PR #877, opened 2023-06-28
[SEDONA-298] Implement ST_ClosestPoint

yyy1000:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-298. The PR name follows the format `[SEDONA-XXX] my subject`.

Add ST_ClosestPointfunction to [sedona-common](https://issues.apache.org/jira/browse/SEDONA-common), [sedona-sql](https://issues.apache.org/jira/browse/SEDONA-sql), [sedona-flink](https://issues.apache.org/jira/browse/SEDONA-flink), and the dataframe/python API.

## How was this patch tested?
1. Comprehensive unit tests in [sedona-common](https://issues.apache.org/jira/browse/SEDONA-common)
2. Integration unit tests in [sedona-sql](https://issues.apache.org/jira/browse/SEDONA-sql), [sedona-flink](https://issues.apache.org/jira/browse/SEDONA-flink), and python.

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/Functions.java
+++ b/common/src/main/java/org/apache/sedona/common/Functions.java
@@ -30,6 +30,7 @@ import org.locationtech.jts.geom.util.GeometryFixer;
 import org.locationtech.jts.io.gml2.GMLWriter;
 import org.locationtech.jts.io.kml.KMLWriter;
 import org.locationtech.jts.linearref.LengthIndexedLine;
+import org.locationtech.jts.operation.distance.DistanceOp;
 import org.locationtech.jts.operation.distance3d.Distance3DOp;
 import org.locationtech.jts.operation.linemerge.LineMerger;
 import org.locationtech.jts.operation.valid.IsSimpleOp;
@@ -445,6 +446,12 @@ public class Functions {
             coordinates.add(c);
         }
         return GEOMETRY_FACTORY.createLineString(coordinates.toArray(new Coordinate[0]));
+    }
+
+    public static Geometry closestPoint(Geometry left, Geometry right) {
+        DistanceOp distanceOp = new DistanceOp(left, right);
+        Coordinate[] closestPoints = distanceOp.nearestPoints();
+        return GEOMETRY_FACTORY.createPoint(closestPoints[0]);
     }
 
     public static Geometry concaveHull(Geometry geometry, double pctConvex, boolean allowHoles){
--- a/common/src/main/java/org/apache/sedona/common/Functions.java
+++ b/common/src/main/java/org/apache/sedona/common/Functions.java
@@ -450,8 +450,13 @@ public class Functions {
 
     public static Geometry closestPoint(Geometry left, Geometry right) {
         DistanceOp distanceOp = new DistanceOp(left, right);
-        Coordinate[] closestPoints = distanceOp.nearestPoints();
-        return GEOMETRY_FACTORY.createPoint(closestPoints[0]);
+        try {
+            Coordinate[] closestPoints = distanceOp.nearestPoints();
+            return GEOMETRY_FACTORY.createPoint(closestPoints[0]);
+        }
+        catch (Exception e) {
+            return null;
+        }
     }
 
     public static Geometry concaveHull(Geometry geometry, double pctConvex, boolean allowHoles){
--- a/common/src/main/java/org/apache/sedona/common/Functions.java
+++ b/common/src/main/java/org/apache/sedona/common/Functions.java
@@ -455,7 +455,7 @@ public class Functions {
             return GEOMETRY_FACTORY.createPoint(closestPoints[0]);
         }
         catch (Exception e) {
-            return null;
+            throw new IllegalArgumentException("ST_ClosestPoint doesn't support empty geometry object.");
         }
     }
 
--- a/common/src/test/java/org/apache/sedona/common/FunctionsTest.java
+++ b/common/src/test/java/org/apache/sedona/common/FunctionsTest.java
@@ -309,7 +309,7 @@ public class FunctionsTest {
         Integer expectedResult = 0;
         assertEquals(actualResult, expectedResult);
 
-        LineString lineString3D = GEOMETRY_FACTORY.createLineString(coordArray(1, 0, 1, 1, 1, 2));
+        LineString lineString3D = GEOMETRY_FACTORY.createLineString(coordArray3d(1, 0, 1, 1, 1, 2));
         actualResult = Functions.dimension(lineString3D);
         expectedResult = 1;
         assertEquals(actualResult, expectedResult);
@@ -1214,7 +1214,7 @@ public class FunctionsTest {
         assertEquals(expected2, actual2);
         Point expectedPoint3 = GEOMETRY_FACTORY.createPoint(new Coordinate(125.75342465753425, 115.34246575342466));
         Double expected3 = Functions.closestPoint(lineString2, point2).distance(expectedPoint3);
-        assertEquals(expected3, 0, 1e-3);
+        assertEquals(expected3, 0, 1e-6);
 
         Point point4 = GEOMETRY_FACTORY.createPoint(new Coordinate(80, 160));
         Polygon polygonA = GEOMETRY_FACTORY.createPolygon(coordArray(190, 150, 20, 10, 160, 70, 190, 150));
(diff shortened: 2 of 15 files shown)


common/src/test/java/org/apache/sedona/common/FunctionsTest.java
```diff
@@ -1200,6 +1199,31 @@ public void geometryTypeWithMeasuredCollection() {
...
+        assertEquals(expected1, actual1);
+
+        Point point2 = GEOMETRY_FACTORY.createPoint(new Coordinate(160, 40));
+        LineString lineString2 = GEOMETRY_FACTORY.createLineString(coordArray(10, 30, 50, 50, 30, 110, 70, 90, 180, 140, 130, 190));
+        String expected2 = "POINT (160 40)";
+        String actual2 = Functions.closestPoint(point2, lineString2).toText();
+        assertEquals(expected2, actual2);
+        Point expectedPoint3 = GEOMETRY_FACTORY.createPoint(new Coordinate(125.75342465753425, 115.34246575342466));
+        Double expected3 = Functions.closestPoint(lineString2, point2).distance(expectedPoint3);
+        assertEquals(expected3, 0, 1e-3);
```
iGN5117, line 1217: Why is the tolerance level 1e-3 here? please standardize to 1e-6 or we can even go upto 1e-9.

common/src/test/java/org/apache/sedona/common/FunctionsTest.java
```diff
@@ -1200,6 +1199,31 @@ public void geometryTypeWithMeasuredCollection() {
         assertEquals(expected4, actual4);
     }
 
+    @Test
+    public void closestPoint() {
+        Point point1 = GEOMETRY_FACTORY.createPoint(new Coordinate(1, 1));
+        LineString lineString1 = GEOMETRY_FACTORY.createLineString(coordArray(1, 0, 1, 1, 2, 1, 2, 0, 1, 0));
```
iGN5117, line 1205: What happens if one or both the geometries are empty? Please add unit tests testing that.

common/src/main/java/org/apache/sedona/common/Functions.java
```diff
@@ -447,6 +448,12 @@ public static Geometry lineFromMultiPoint(Geometry geometry) {
         return GEOMETRY_FACTORY.createLineString(coordinates.toArray(new Coordinate[0]));
     }
 
+    public static Geometry closestPoint(Geometry left, Geometry right) {
+        DistanceOp distanceOp = new DistanceOp(left, right);
+        Coordinate[] closestPoints = distanceOp.nearestPoints();
```
iGN5117, line 453: Since we're using JTS DistanceOp, what happens if the one or both the input geometries are 3D? Is the z ordinate ignored, or it is supported?
We will have to update documentation and tests accordingly.

common/src/test/java/org/apache/sedona/common/FunctionsTest.java
```diff
@@ -1200,6 +1199,31 @@ public void geometryTypeWithMeasuredCollection() {
...
+        Double expected3 = Functions.closestPoint(lineString2, point2).distance(expectedPoint3);
+        assertEquals(expected3, 0, 1e-3);
+
+        Point point4 = GEOMETRY_FACTORY.createPoint(new Coordinate(80, 160));
+        Polygon polygonA = GEOMETRY_FACTORY.createPolygon(coordArray(190, 150, 20, 10, 160, 70, 190, 150));
+        Geometry polygonB = Functions.buffer(point4, 30);
+        Point expectedPoint4 = GEOMETRY_FACTORY.createPoint(new Coordinate(131.59149149528952, 101.89887534906197));
+        Double expected4 = Functions.closestPoint(polygonA, polygonB).distance(expectedPoint4);
+        assertEquals(expected4, 0, 1e-6);
+    }
```
iGN5117, line 1225: Please add a unit test case testing geometry collections

docs/api/flink/Function.md
```diff
@@ -363,6 +363,31 @@ Input: `MULTILINESTRING((0 0, 10 0, 10 10, 0 10, 0 0),(10 10, 20 10, 20 20, 10 2
...
+
+Introduction: Returns the 2-dimensional point on geom1 that is closest to geom2. This is the first point of the shortest line between the geometries.
+
+Format: `ST_ClosestPoint(g1: geomtry, g2: geometry)`
+
+Since: `1.5.0`
+
+Example1:
+```sql
+SELECT ST_AsText( ST_ClosestPoint(g1, g2)) As ptwkt;
```
iGN5117, line 376: This documentation needs to be updated about 3D geometries, based on testing JTS DistanceOp

iGN5117 (changes requested):

common/src/test/java/org/apache/sedona/common/FunctionsTest.java
yyy1000: OK, but I'm a little confused. Though it could work. I search it but can't find any usage about geometry collections. XD

common/src/main/java/org/apache/sedona/common/Functions.java
```diff
@@ -447,6 +448,17 @@ public static Geometry lineFromMultiPoint(Geometry geometry) {
...
     }
 
+    public static Geometry closestPoint(Geometry left, Geometry right) {
+        DistanceOp distanceOp = new DistanceOp(left, right);
+        try {
+            Coordinate[] closestPoints = distanceOp.nearestPoints();
+            return GEOMETRY_FACTORY.createPoint(closestPoints[0]);
+        }
+        catch (Exception e) {
+            return null;
```
iGN5117, line 458: Do not throw null here, throw a meaningful exception. If this is thrown if the geometry is empty, throw an IllegalArgumentException with an appropriate message

common/src/test/java/org/apache/sedona/common/FunctionsTest.java
iGN5117: There could be many possible use cases. Like given a geometry collection representing a neighborhood (polygon/multipolygon for houses), find closest point to a given external point.

common/src/test/java/org/apache/sedona/common/FunctionsTest.java
```diff
@@ -1200,6 +1199,74 @@ public void geometryTypeWithMeasuredCollection() {
...
+    @Test
+    public void closestPointEmpty() {
+        // One of the object is empty
+        Point point = GEOMETRY_FACTORY.createPoint(new Coordinate(1, 1));
+        LineString emptyLineString = GEOMETRY_FACTORY.createLineString();
+        Geometry actual = Functions.closestPoint(point, emptyLineString);
+        assertNull(actual);
+
+        // Both objects are empty
+        Polygon emptyPolygon = GEOMETRY_FACTORY.createPolygon();
```
iGN5117, line 1265: refactor these test cases to catch the meaningful exception.
There would be 2 asserts, 1 for the type of exception thrown, 1 for the message in the exception.
Check implementation and tests of nRings for reference

docs/api/flink/Function.md
iGN5117: Update documentation here to include the thrown exception if any of the geometry is empty

docs/api/sql/Function.md
```diff
@@ -381,6 +381,32 @@ SELECT ST_Centroid(polygondf.countyshape)
 FROM polygondf
 ```
 
+## ST_ClosestPoint
+
+Introduction: Returns the 2-dimensional point on geom1 that is closest to geom2. This is the first point of the shortest line between the geometries. If using 3D geometries, the Z coordinates will be ignored. If you have a 3D Geometry, you may prefer to use ST_3DClosestPoint.
+
```
iGN5117, line 387: Update documentation here to include the thrown exception if any of the geometry is empty

iGN5117 (changes requested):

common/src/main/java/org/apache/sedona/common/Functions.java
yyy1000: Should we throw an exception or just return a null value? I think postgis return an empty value:
```
postgres=# select  ST_AsText(ST_ClosestPoint(ST_GeomFromText('POINT EMPTY'), 
                                   ST_GeomFromText('LINESTRING (0 0, 1 0, 2 0, 3 0, 4 0, 5 0)')));
 st_astext 
-----------
 
(1 row)
```

iGN5117: It is better to throw an exception, in a case where multiple functions are stringed together, which is what happens usually, throwing null means user would see an unexplainable NullPointerException eventually.

By throwing an IllegalArgumentException with a proper message, we will end up helping the user.

yyy1000: Good idea!

docs/api/sql/Function.md
yyy1000: I tried to resolve all your comments! You can review again.  :) Thanks!

common/src/main/java/org/apache/sedona/common/Functions.java
yyy1000: Based on this, I tried to resolve all your comments! You can review again.  :) Thanks! 😁

jiayuasu (approved):

PR #887, opened 2023-07-04
[SEDONA-303] Port all Sedona Spark functions to Sedona Flink -- Step 4

yyy1000:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-303. The PR name follows the format `[SEDONA-XXX] my subject`.

1. Port missing ST_functions including 
{ST_PrecisionReduce and ST_MinimumBoundingCircle}to [sedona-flink](https://issues.apache.org/jira/browse/SEDONA-flink).
2. Change the default value of the second param of ST_MinimumBoundingCircle from 8 to 48 and its tests. See https://postgis.net/docs/ST_MinimumBoundingCircle.html for reference.
3. Change the name of ST_PrecisionReduce to ST_ReducePrecision, and its tests and docs, see https://postgis.net/docs/ST_ReducePrecision.html for reference. 

## How was this patch tested?
1. Integration unit tests in [sedona-flink](https://issues.apache.org/jira/browse/SEDONA-flink).
2. Integration unit tests in [sedona-sql](https://issues.apache.org/jira/browse/SEDONA-sql), [sedona-flink](https://issues.apache.org/jira/browse/SEDONA-flink), and python.


jiayuasu:
@yyy1000 Please fix the failed Python tests

yyy1000:
> @yyy1000 Please fix the failed Python tests

Fixed it. :)

docs/api/sql/Function.md
```diff
@@ -1552,18 +1552,18 @@ SELECT ST_AsText(ST_PointOnSurface(ST_GeomFromText('LINESTRING(0 5 1, 0 0 1, 0 1
 
 ```
 
-## ST_PrecisionReduce
+## ST_ReducePrecision
 
 Introduction: Reduce the decimals places in the coordinates of the geometry to the given number of decimal places. The last decimal place will be rounded.
 
-Format: `ST_PrecisionReduce (A:geometry, B:int)`
```
jiayuasu, line 1559: Please add one more sentence saying that the name of this function was called ST_PrecisionReduce in versions prior to v1.5.0.

sql/common/src/main/scala/org/apache/sedona/sql/UDF/Catalog.scala
```diff
@@ -102,7 +102,7 @@ object Catalog {
     function[ST_ClosestPoint](),
     function[ST_Boundary](),
     function[ST_MinimumBoundingRadius](),
-    function[ST_MinimumBoundingCircle](BufferParameters.DEFAULT_QUADRANT_SEGMENTS),
```
jiayuasu, line 105: Please update the doc in Sedona Spark saying that the default param has been changed.

jiayuasu (changes requested):

docs/api/sql/Function.md
yyy1000: I just noticed that `ST_ReducePrecision(geometry g, float8 gridsize);` in postgis needs float as the second param. That means use '0.1' as 1, '0.01' as 2. Should we follow the function in postgis?

jiayuasu: em... I think the Sedona implementation of this function is very different from PostGIS. I suggest that, in this PR, we don't adopt the PostGIS implementation. We will address this in a different PR.

yyy1000: Yeah, I think so. Here we can keep Sedona implementation. Should we use the keep the function name the same as PostGIS like 'ST_ReducePrecision '? Or we can revert to the old name 'ST_PrecisionReduce ', which doesn't exsit in PostGIS.

jiayuasu: @yyy1000 I think changing to `ST_ReducePrecision` is fine. No need to revert it back. Just update the doc then we are good to go.

yyy1000:
I updated the doc. :)

docs/api/flink/Function.md
```diff
@@ -1244,6 +1244,19 @@ Result:
 +------------------+------------------------+
 ```
 
+## ST_MinimumBoundingCircle
+
+Introduction: Returns the smallest circle polygon that contains a geometry. The optional quadrantSegments parameter determines how many segments to use per quadrant and the default number of segments is 48. 
+
+Format: `ST_MinimumBoundingCircle(geom: geometry, [Optional] quadrantSegments:int)`
+
+Since: `v1.0.1`
```
jiayuasu, line 1253: Why is this version reverted back to v1.0.1? All these functions should be v1.5.0 @yyy1000

yyy1000: Ah, sorry for missing that. Updated.

jiayuasu (approved):

jiayuasu (approved):

PR #898, opened 2023-07-11
[SEDONA-323] Add keplergl wrapper

iGN5117:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-323. The PR name follows the format `[SEDONA-XXX] my subject`.

- This PR adds SedonaKepler in sedona.maps. SedonaKepler is a wrapper for creating keplergl maps in sedona. 

## How was this patch tested?
- Refactored notebook example to include SedonaKepler, and verified correct map rendering.
- Ran basic test case testing SedonaKepler map object and keplergl map object.

- Yes, I have added documentation for using SedonaKepler.


python/tests/sql/test_sedonakepler_visualization.py
jiayuasu, line 1: Please move this test to `python/tests/maps/`

python/__init__.py
jiayuasu, line 1: This file is not needed as the `Python` folder per se is not a package.

python/sedona/maps/SedonaKepler.py
jiayuasu, line 1: Please use create_map() and add_df()

jiayuasu (changes requested):

python/__init__.py
iGN5117: This was not intended to be added, removed this.

python/sedona/maps/SedonaKepler.py
iGN5117: Done

python/tests/sql/test_sedonakepler_visualization.py
iGN5117: Done

python/sedona/maps/SedonaKepler.py
```diff
@@ -0,0 +1,58 @@
+from keplergl import KeplerGl
+import geopandas as gpd
```
jiayuasu, line 2: You introduced 4 new files in this PR. All new files must have Apache License header. Please refer to existing python files in Sedona.

jiayuasu (changes requested):

python/sedona/maps/SedonaKepler.py
iGN5117: Added

python/tests/maps/test_sedonakepler_visualization.py
```diff
@@ -0,0 +1,28 @@
...
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+
+from keplergl import KeplerGl
+from sedona.maps.SedonaKepler import SedonaKepler
+from tests.test_base import TestBase
+
+
+class TestVisualization(TestBase):
```
Kontinuation, line 23: Is it possible to add more tests for `SedonaKepler`, such as creating a map with an initial dataframe or adding a dataframe to an existing map?

Kontinuation: The test fails due to some known bugs of keplergl or geopandas (https://github.com/keplergl/kepler.gl/issues/1240), we need to work around this problem. We can check whether the data configuration with a specific dataId is added to the map config `kepler_map.config`, instead of checking the result of `kepler_map._repr_html()`.

iGN5117: I do not think thats the case as the tests pass in my local machine, and here even the map_creation_with_df test fails which does not add the df separately.
I am investigating the root cause

iGN5117: I cannot work with config as that would require me to create config objects for the map everytime I add a new df. As this bug only affects when adding the same dataset twice, I believe getting around this by adding a copy of the dataset is okay for these tests?

Kontinuation: I found that `_repr_html()` only works with pandas>=2.0.0. If we upgrade pandas in Pipfile, we have to drop tests for python 3.7.

If upgrading pandas is not an option, the best thing we can do is to check the keys of `kepler_map.data`, or simply don't validate the `kepler_map` since the constructor of `KeplerGl` has already validated the data. A successful map object construction indicates that everything is fine.

iGN5117: I am able to get around this issue by exclusively using add_data of keplerGL instead of its df constructor. Please review this if the tests succeed @Kontinuation @jiayuasu

Kontinuation (approved):

docs/tutorial/sql.md
```diff
@@ -505,6 +505,81 @@ The details of a join query is available here [Join query](../api/sql/Optimizer.
...
+		```python
+		map = SedonaKepler.create_map()
+		SedonaKepler.add_df(map, groupedresult, name="AirportCount", geometry_col="country_geom")
+		map
+		```
+
+### Setting a config via the map 
+A map rendered by accessing the map object created by SedonaKepler includes a config panel which can be used to customize the map
+
+<img src="../../image/sedona_customization.gif" width="1000">
```
jiayuasu, line 573: @iGN5117 This doc references to this image but this image was not committed together with this PR.

jiayuasu (changes requested):

docs/tutorial/sql.md
iGN5117: Sorry about that, pushed the gif

jiayuasu (approved):

PR #978, opened 2023-08-18
[SEDONA-368] Add RS_SummaryStats

furqaankhan:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-368. The PR name follows the format `[SEDONA-XXX] my subject`.

- Add RS_SummaryStats

## How was this patch tested?

- Passed all added tests


common/src/main/java/org/apache/sedona/common/raster/RasterBandAccessors.java
```diff
@@ -78,6 +80,68 @@ public static long getCount(GridCoverage2D raster, int band) {
...
+//        return getSummaryStats(raster, 1, excludeNoDataValue);
+//    }
+
+    private static double getStddev(double[] pixels, double mean) {
+        double stddev = 0;
+        for(double pixel: pixels){
+             stddev += Math.pow(pixel - mean, 2);
+        }
+        return Math.sqrt(stddev/(pixels.length - 1));
+    }
```
Kontinuation, line 143: Do we need to exclude pixels with nodata value when computing std-dev?

common/src/main/java/org/apache/sedona/common/raster/RasterBandAccessors.java
```diff
@@ -78,6 +80,68 @@ public static long getCount(GridCoverage2D raster, int band) {
...
+                    sum += pixel;
+                    if (pixel < min) {
+                        min = pixel;
+                    }
+                    if (pixel > max) {
+                        max = pixel;
+                    }
+                }
+            }
+            mean = sum / count;
```
Kontinuation, line 104: We need to be aware of the cases when all pixels in the band are nodata value. PostGIS returns all statistics except count as `NULL`. I think both NULL and NaN are sane choices.

furqaankhan: Okay, will add a check before the compute code and return NULL.

common/src/test/java/org/apache/sedona/common/raster/RasterBandAccessorsTest.java
```diff
@@ -74,6 +74,116 @@ public void testBandNoDataValueIllegalBand() throws FactoryException, IOExceptio
         assertEquals("Provided band index 2 is not present in the raster", exception.getMessage());
     }
 
+    @Test
```
iGN5117, line 77: For the tests, since we're dealing with means and std which are small, can we see if delta of 1e-6 is sufficient instead of 0.1d?

iGN5117 (changes requested):

iGN5117 (approved):

PR #1001, opened 2023-09-05
[SEDONA-383] Add RS_Band

furqaankhan:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-383. The PR name follows the format `[SEDONA-XXX] my subject`.

- Add RS_Band

## How was this patch tested?

- Passed added tests

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/raster/RasterBandAccessors.java
+++ b/common/src/main/java/org/apache/sedona/common/raster/RasterBandAccessors.java
@@ -22,6 +22,7 @@ import org.apache.sedona.common.utils.RasterUtils;
 import org.geotools.coverage.GridSampleDimension;
 import org.geotools.coverage.grid.GridCoordinates2D;
 import org.geotools.coverage.grid.GridCoverage2D;
+import org.opengis.referencing.FactoryException;
 
 import java.awt.image.Raster;
 
@@ -133,6 +134,25 @@ public class RasterBandAccessors {
 //    public static double[] getSummaryStats(GridCoverage2D raster, boolean excludeNoDataValue) {
 //        return getSummaryStats(raster, 1, excludeNoDataValue);
 //    }
+
+    public static GridCoverage2D getBand(GridCoverage2D rasterGeom, int[] bands) throws FactoryException {
+        double[] metadata = RasterAccessors.metadata(rasterGeom);
+        GridCoverage2D resultRaster = RasterConstructors.makeEmptyRaster(bands.length, (int) metadata[2], (int) metadata[3],
+                metadata[0], metadata[1], metadata[4], metadata[5], metadata[6], metadata[7], (int) metadata[8]);
+        double[] curBandData;
+        Double noDataValue;
+        for (int curBand: bands) {
+            RasterUtils.ensureBand(rasterGeom, curBand);
+            curBandData = MapAlgebra.bandAsArray(rasterGeom, curBand);
+            noDataValue = RasterBandAccessors.getBandNoDataValue(rasterGeom, curBand);
+            if (noDataValue != null) {
+                resultRaster = MapAlgebra.addBandFromArray(resultRaster, curBandData, curBand, noDataValue);
+                continue;
+            }
+            resultRaster = MapAlgebra.addBandFromArray(resultRaster, curBandData, curBand);
+        }
+        return resultRaster;
+    }
 
     public static String getBandType(GridCoverage2D raster, int band) {
         RasterUtils.ensureBand(raster, band);
--- a/common/src/main/java/org/apache/sedona/common/raster/RasterBandAccessors.java
+++ b/common/src/main/java/org/apache/sedona/common/raster/RasterBandAccessors.java
@@ -25,9 +25,9 @@ import org.geotools.coverage.grid.GridCoverage2D;
 import org.opengis.referencing.FactoryException;
 
 import java.awt.image.Raster;
-import java.awt.image.RenderedImage;
 import java.awt.image.WritableRaster;
 import java.util.Arrays;
+import java.util.HashMap;
 
 public class RasterBandAccessors {
 
@@ -147,10 +147,10 @@ public class RasterBandAccessors {
 
         // Get band data that's required
         int[] bandsDistinct = Arrays.stream(bandIndexes).distinct().toArray();
-        double [][] bandData = new double[bandsDistinct.length][height * width];
+        HashMap<Integer, double[]> bandData = new HashMap<>();
         for (int curBand: bandsDistinct) {
             RasterUtils.ensureBand(rasterGeom, curBand);
-            bandData[curBand - 1] = MapAlgebra.bandAsArray(rasterGeom, curBand);
+            bandData.put(curBand - 1, MapAlgebra.bandAsArray(rasterGeom, curBand));
         }
 
         // Get Writable Raster from the resultRaster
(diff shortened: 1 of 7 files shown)


common/src/main/java/org/apache/sedona/common/raster/RasterBandAccessors.java
```diff
@@ -134,6 +135,25 @@ public static double[] getSummaryStats(GridCoverage2D raster) {
...
+        GridCoverage2D resultRaster = RasterConstructors.makeEmptyRaster(bands.length, (int) metadata[2], (int) metadata[3],
+                metadata[0], metadata[1], metadata[4], metadata[5], metadata[6], metadata[7], (int) metadata[8]);
+        double[] curBandData;
+        Double noDataValue;
+        for (int curBand: bands) {
+            RasterUtils.ensureBand(rasterGeom, curBand);
+            curBandData = MapAlgebra.bandAsArray(rasterGeom, curBand);
+            noDataValue = RasterBandAccessors.getBandNoDataValue(rasterGeom, curBand);
+            if (noDataValue != null) {
+                resultRaster = MapAlgebra.addBandFromArray(resultRaster, curBandData, curBand, noDataValue);
```
jiayuasu, line 149: You should not directly use `addBandFromArray`. Each function call will copy and create a new raster, which is very slow. You should implement a version of `copyRasterAndReplaceBand` or borrow some logic from that to set pixel values of all bands in one go.

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/InferredExpression.scala
```diff
@@ -100,6 +100,10 @@ object InferrableType {
     new InferrableType[Array[Byte]] {}
   implicit val longArrayInstance: InferrableType[Array[java.lang.Long]] =
     new InferrableType[Array[java.lang.Long]] {}
+  implicit val intArrayInstance: InferrableType[Array[Int]] =
+    new InferrableType[Array[Int]] {}
+  implicit val javaIntArrayInstance: InferrableType[Array[java.lang.Integer]] =
```
jiayuasu, line 105: Why do you need both `Int` and `java.lang.Integer`?

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/raster/RasterBandAccessors.scala
```diff
@@ -45,6 +52,34 @@ case class RS_SummaryStats(inputExpressions: Seq[Expression]) extends InferredEx
   }
 }
 
+case class RS_Band(inputExpressions: Seq[Expression]) extends Expression with SerdeAware with CodegenFallback {
```
jiayuasu, line 55: Why not use the inferredexpression?

jiayuasu (changes requested):

common/src/main/java/org/apache/sedona/common/raster/RasterBandAccessors.java
furqaankhan: Sure, I was going to do that but I thought to reuse.

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/raster/RasterBandAccessors.scala
furqaankhan: That makes it so it can't find the RS_Band function, have to use Expression.

jiayuasu: Why is this happening? We need to use inferredExpression whenever we can.

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/InferredExpression.scala
jiayuasu: The existing approach is using Array[java.lang.Long]. Why not follow the same style?

furqaankhan: My bad, will remove the `Int` and `java.lang.Integer` and replace it with `java.lang.Long`

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/raster/RasterBandAccessors.scala
furqaankhan: Oh didn't know that, Will look into it. I referred to a function that use Expression instead of InferredExpression, ST_Collect being one. Will ping you when I know more.

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/InferredExpression.scala
jiayuasu: Dude, what I mean is that why not use Array[java.lang.Integer] just like Array[java.lang.Long]

furqaankhan: No particular reason, scala by default takes `Int`. I guess the same reason we support `Double` and `java.lang.Double`

jiayuasu: OK got it @Kontinuation Can you confirm this?

Kontinuation: Scala `Int` is a primitive type, which is different from `java.lang.Integer`. It is necessary to have InferrableType instances for both `Array[Int]` and `Array[java.lang.Integer]` to support passing both `int` and `int[]` parameters.

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/raster/RasterBandAccessors.scala
```diff
@@ -45,6 +52,28 @@ case class RS_SummaryStats(inputExpressions: Seq[Expression]) extends InferredEx
   }
 }
 
+case class RS_Band(inputExpressions: Seq[Expression]) extends InferredExpression(RasterBandAccessors.getBand _) with SerdeAware with CodegenFallback {
```
jiayuasu, line 55: No need to add SerdeAware and CodgenFallback here as InferredExpression already has those.

jiayuasu (changes requested):

common/src/main/java/org/apache/sedona/common/raster/RasterBandAccessors.java
```diff
@@ -134,6 +138,39 @@ public static double[] getSummaryStats(GridCoverage2D raster) {
...
+        int width = (int) metadata[2], height = (int) metadata[3];
+        GridCoverage2D resultRaster = RasterConstructors.makeEmptyRaster(bandIndexes.length, width, height,
+                metadata[0], metadata[1], metadata[4], metadata[5], metadata[6], metadata[7], (int) metadata[8]);
+
+        // Get band data that's required
+        int[] bandsDistinct = Arrays.stream(bandIndexes).distinct().toArray();
+        double [][] bandData = new double[bandsDistinct.length][height * width];
+        for (int curBand: bandsDistinct) {
+            RasterUtils.ensureBand(rasterGeom, curBand);
+            bandData[curBand - 1] = MapAlgebra.bandAsArray(rasterGeom, curBand);
```
jiayuasu, line 153: The code here has a bug. It cannot handle cases such as `3, 3, 3` and will throw index out of bound error

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/raster/RasterBandAccessors.scala
```diff
@@ -45,6 +52,28 @@ case class RS_SummaryStats(inputExpressions: Seq[Expression]) extends InferredEx
   }
 }
 
+case class RS_Band(inputExpressions: Seq[Expression]) extends InferredExpression(RasterBandAccessors.getBand _) {
```
jiayuasu, line 55: Please optimize the imports and remove unnecessary imports.

jiayuasu (changes requested):

jiayuasu (approved):

PR #1011, opened 2023-09-07
[SEDONA-388] Add RS_AsRaster

furqaankhan:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-388. The PR name follows the format `[SEDONA-XXX] my subject`.

- Add RS_AsRaster
- Fix alphabetic order of Raster writer docs
- Add gt-process-feature dependency to maven
- Add new function RasterOutputs.toGeotiff for vizualizing a raster internally (Not exposed to the public)

## How was this patch tested?

- Passed added tests


common/src/test/java/org/apache/sedona/common/raster/RasterConstructorsTest.java
```diff
@@ -55,6 +58,22 @@ public void fromGeoTiff() throws IOException, FactoryException {
...
+        assertArrayEquals(expected, actual);
```
Kontinuation, line 68: I suggest asserting on the band values instead of the byte sequence of GeoTiff

docs/api/sql/Raster-writer.md
```diff
@@ -183,3 +183,59 @@ sparkSession.read.format("binaryFile").load("my_raster_file/*")
 Then you can create Raster type in Sedona like this `RS_FromGeoTiff(content)` (if the written data was in GeoTiff format).
 
 The newly created DataFrame can be written to disk again but must be under a different name such as `my_raster_file_modified`
+
+### Write Geometry to Raster dataframe
+
+#### RS_AsRaster
+
+Introduction: Converts a Geometry to a Raster dataset. Defaults to using `1.0` for cell `value` and `null` for `noDataValue` if not provided. Supports all geometry types.
```
jiayuasu, line 191: Please explain the possible values for pixel type

common/src/main/java/org/apache/sedona/common/raster/RasterConstructors.java
```diff
@@ -45,6 +55,94 @@ public static GridCoverage2D fromGeoTiff(byte[] bytes) throws IOException {
...
+        width = (int) (width / scaleX);
+        height = (int) (height / scaleY);
+
+        VectorToRasterProcess rasterProcess = new VectorToRasterProcess();
+        GridCoverage2D rasterized = rasterProcess.execute(featureCollection, width, height, "value", Double.toString(value), bound, null);
+        if (noDataValue != null) {
+            rasterized = RasterBandEditors.setBandNoDataValue(rasterized, 1, noDataValue);
+        }
+        WritableRaster writableRaster = RasterFactory.createBandedRaster(RasterUtils.getDataTypeCode(pixelType), width, height, 1, null);
+        double [] samples = rasterized.getRenderedImage().getData().getSamples(0, 0, width, height, 0, (double[]) null);
```
jiayuasu, line 110: Please use `RasterUtils.getRaster` function to replace rasterized.getRenderImage().getData

common/src/main/java/org/apache/sedona/common/raster/RasterConstructors.java
```diff
@@ -45,6 +55,94 @@ public static GridCoverage2D fromGeoTiff(byte[] bytes) throws IOException {
...
+        // To preserve scale of reference raster
+        width = (int) (width / scaleX);
+        height = (int) (height / scaleY);
+
+        VectorToRasterProcess rasterProcess = new VectorToRasterProcess();
+        GridCoverage2D rasterized = rasterProcess.execute(featureCollection, width, height, "value", Double.toString(value), bound, null);
+        if (noDataValue != null) {
+            rasterized = RasterBandEditors.setBandNoDataValue(rasterized, 1, noDataValue);
+        }
+        WritableRaster writableRaster = RasterFactory.createBandedRaster(RasterUtils.getDataTypeCode(pixelType), width, height, 1, null);
```
jiayuasu, line 109: Why does this call `RasterFactory.createBandedRaster` to get the GridCoverage2D? Is it because the original GridCoverage2D object is not banded?

common/src/main/java/org/apache/sedona/common/raster/RasterOutputs.java
```diff
@@ -74,6 +76,26 @@ public static byte[] asGeoTiff(GridCoverage2D raster) {
...
+    * Creates a GeoTiff file with the provided raster. Primarily used for testing.
+     * @param raster The raster to be stored as .tiff
+     * @param filePath The path where the .tiff should be stored.
+     *
+     * @return true if file is created, otherwise throws IOException
+     * @throws IOException
+     * */
+    public static boolean toGeoTiff(GridCoverage2D raster, String filePath) {
+        File outputFile = new File(filePath);
+        GeoTiffFormat format = new GeoTiffFormat();
```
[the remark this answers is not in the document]
jiayuasu: And this function is not tested. Please add a test.

common/src/test/java/org/apache/sedona/common/raster/RasterConstructorsTest.java
```diff
@@ -55,6 +58,22 @@ public void fromGeoTiff() throws IOException, FactoryException {
         assertEquals(4, gridCoverage2D.getNumSampleDimensions());
     }
 
+    @Test
+    public void testAsRasterWithEmptyRaster() throws FactoryException, ParseException {
```
jiayuasu, line 62: Please add a few more tests using the real test geotiff images

jiayuasu (changes requested):

common/src/main/java/org/apache/sedona/common/raster/RasterConstructors.java
furqaankhan: Will do a back merge and use it.

common/src/main/java/org/apache/sedona/common/raster/RasterConstructors.java
```diff
@@ -45,6 +55,94 @@ public static GridCoverage2D fromGeoTiff(byte[] bytes) throws IOException {
...
+
+        double[] metadata = RasterAccessors.metadata(raster);
+        // The current implementation doesn't support rasters with properties below
+        // It is not a problem as most rasters don't have these properties
+        // ScaleX < 0
+        if (metadata[4] < 0) {
+            throw new IllegalArgumentException(String.format("ScaleX %f of the raster is negative, it should be positive", metadata[4]));
+        }
+        // ScaleY > 0
+        if (metadata[5] > 0) {
```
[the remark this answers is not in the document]
furqaankhan: If the ScaleX < 0 and ScaleY > 0 then we would have to flip the raster. To keep the function simple, we decided to not do it for now. I consulted Kristin on this.

spark/common/pom.xml
```diff
@@ -122,6 +122,10 @@
             <groupId>org.geotools</groupId>
             <artifactId>gt-geotiff</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.geotools</groupId>
+            <artifactId>gt-process-feature</artifactId>
```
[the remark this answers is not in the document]
Kontinuation: I think we need to add it to geotools-wrapper. I'm afraid adding it to spark-shaded will have any licensing issues.

common/src/main/java/org/apache/sedona/common/raster/RasterConstructors.java
furqaankhan: No, to define the pixel type of the raster. Is there another way to define it?

spark/common/pom.xml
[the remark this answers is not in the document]
Kontinuation: It reminds me that we also need to add this to geotools-wrapper.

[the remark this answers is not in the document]
jiayuasu: @Kontinuation The geotools scope in spark-shaded is `provided` so we should be fine. Putting it in `spark-shaded` will help Python and R tests in the future since they can simply do `mvn clean install -Dgeotools`

common/src/main/java/org/apache/sedona/common/raster/RasterConstructors.java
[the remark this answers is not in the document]
jiayuasu: Got it. But since your code does two separate if condition check, this logic only allows `ScaleX >=0 and ScaleY <= 0` and disabled all other possibility. Is this an expected behavior? @Kontinuation

common/src/test/java/org/apache/sedona/common/raster/RasterConstructorsTest.java
jiayuasu: Agree. To make things easier, add a test case that uses a Point geometry and checks if the corresponding pixel has the expected rasterized value.

jiayuasu (changes requested):

common/src/main/java/org/apache/sedona/common/raster/RasterConstructors.java
jiayuasu: seems to be fine.

common/src/main/java/org/apache/sedona/common/raster/RasterConstructors.java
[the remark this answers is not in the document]
Kontinuation: Yes. This is the most common case for north-up images. For any other cases, we need to flip the resulting raster.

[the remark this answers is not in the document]
jiayuasu: Got it. Thanks!

common/src/main/java/org/apache/sedona/common/raster/RasterConstructors.java
```diff
@@ -45,6 +55,116 @@ public static GridCoverage2D fromGeoTiff(byte[] bytes) throws IOException {
...
+        int width = (int) bound.getWidth(), height = (int) bound.getHeight();
+
+        if (geom.getGeometryType().equalsIgnoreCase(Geometry.TYPENAME_POINT)) {
+            width = 1;
+            height = 1;
+        } else {
+            // To preserve scale of reference raster
+            width = (int) (width / scaleX);
+            height = (int) (height / scaleY);
+        }
```
Kontinuation, line 109: There are some special treatments for points. I think it is also needed for horizontal and vertical line strings. We also need to add tests for these cases.

common/src/test/java/org/apache/sedona/common/raster/RasterConstructorsTest.java
```diff
@@ -55,6 +58,99 @@ public void fromGeoTiff() throws IOException, FactoryException {
...
+        assertArrayEquals(expected, actual, 0.1d);
```
Kontinuation, line 86: These walls of texts as expected results are very confusing.
I suggest rendering smaller rasters, or picking some representative points and only testing them.

furqaankhan: I agree, altering RN.

common/src/main/java/org/apache/sedona/common/raster/RasterConstructors.java
Kontinuation: Actually, any kind of geometry can have a sub-pixel size. I suggest simply checking the width/height of the raster to determine if we need to extend its width or height to at least 1 pixel, so that we don't need special treatments for different types of geometries.

common/src/main/java/org/apache/sedona/common/raster/RasterOutputs.java
```diff
@@ -74,6 +77,25 @@ public static byte[] asGeoTiff(GridCoverage2D raster) {
...
 
+    /**
+    * Creates a GeoTiff file with the provided raster. Primarily used for testing.
+     * @param raster The raster to be stored as .tiff
+     * @param filePath The path where the .tiff should be stored.
+     *
+     * @return true if file is created, otherwise throws IOException
+     * @throws IOException
+     * */
+    public static boolean toGeoTiff(GridCoverage2D raster, String filePath) {
```
jiayuasu, line 88: This function should be called `writeToDiskFile`. It takes a byte[] array as input and write the file to disk. This is a generic function that works with the result of asGeoTiff, asArcGrid, and so on.

common/src/test/java/org/apache/sedona/common/raster/RasterOutputTest.java
```diff
@@ -44,6 +44,19 @@ public void testAsGeoTiff() throws IOException {
         assertEquals(rasterTest.getEnvelope().toString(), rasterOg.getEnvelope().toString());
     }
 
+    @Test
+    public void testToGeoTiff() throws IOException {
+        GridCoverage2D rasterOg = rasterFromGeoTiff(resourceFolder + "raster/test1.tiff");
+        String filePath = System.getProperty("user.dir") + "/../spark/common/src/test/resources/testToGeoTiffFunction/test1.tiff";
```
jiayuasu, line 50: The file path is wrong. It should be `System.getProperty("user.dir") + "/target/estToGeoTiffFunction/test1.tiff";`. Do not add content to the test resource folder.

spark/common/src/test/resources/testToGeoTiffFunction/test1.tiff
jiayuasu, line 1: This file is generated by your `toGeoTiff`. It should not be committed. And again, the test should not write result back to the resources folder.

jiayuasu (changes requested):

Kontinuation (approved):

PR #1039, opened 2023-09-28
[DOCS] Mention behavior of string returning functions

furqaankhan:

- No, this is a documentation update. The PR name follows the format `[DOCS] my subject`.

- `show` function escapes special characters, this messes up with functions that return string with special characters.

### DIFF
--- a/docs/api/sql/Raster-operators.md
+++ b/docs/api/sql/Raster-operators.md
@@ -194,11 +194,26 @@ Introduction: Returns the georeference metadata of raster as a string in GDAL or
 
 !!!note
     If you are using `show()` to display the output, it will show special characters as escape sequences. To get the expected behavior use the following code:
+    
+    === "Scala"
 
-    ```sql
-    print(df.selectExpr("RS_GeoReference(rast)").sample(0.5).collect().mkString(""))
-    ```
-    Use `sample` and `collect` to print the georeference.
+        ```scala
+        println(df.selectExpr("RS_GeoReference(rast)").sample(0.5).collect()(0))
+        ```
+    
+    === "Java"
+    
+        ```java
+        System.out.println(df.selectExpr("RS_GeoReference(rast)").sample(0.5).collect()[0])
+        ```
+    
+    === "Python"
+    
+        ```python
+        print(df.selectExpr("RS_GeoReference(rast)").sample(0.5).collect()[0])
+        ```
+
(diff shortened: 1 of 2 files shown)


furqaankhan:
Please let me know if there are more functions that return strings with special characters.

docs/api/sql/Raster-operators.md
```diff
@@ -192,6 +192,14 @@ Provided band index 3 does not lie in the raster
 
 Introduction: Returns the georeference metadata of raster as a string in GDAL or ESRI format. Default is GDAL if not specified.
 
+!!!note
+    If you are using `show()` to display the output, it will show special characters as escape sequences. To get the expected behavior use the following code:
+
+    ```sql
+    print(df.selectExpr("RS_GeoReference(rast)").sample(0.5).collect().mkString(""))
```
jiayuasu, line 199: This is not a SQL example. This should be a language-specific example. You need to provide examples for (Scala/Java) and (Python). Please also say `sample()` here is to reduce data to be collected. Other functions such as filter() will work as well.

docs/api/sql/Raster-visualizer.md
```diff
@@ -68,6 +68,14 @@ RS_AsMatrix expects a raster, and optionally a band (default: 1) and postDecimal
 !!!Note
     If the provided raster has integral values, postDecimalPrecision (if any) is simply ignored and integers are printed in the resultant string
 
+!!!note
+    If you are using `show()` to display the output, it will show special characters as escape sequences. To get the expected behavior use the following code:
+
+    ```sql
+    print(df.selectExpr("RS_AsMatrix(rast)").sample(0.5).collect().mkString(""))
```
jiayuasu, line 75: This is not a SQL example. This should be a language-specific example. You need to provide examples for (Scala/Java) and (Python). Please also say `sample()` here is to reduce data to be collected. Other functions such as filter() will work as well.

jiayuasu (changes requested):

docs/api/sql/Raster-operators.md
furqaankhan: I don't think the filter function makes much sense here. As there's nothing to filter, the function returns one row, that's a string with special characters for formatting.

jiayuasu: @furqaankhan sample and filter are operations that can reduce the number of resulting rows. Without this kind of commands, a direct collect() on a DataFrame that has millions of row will crash the driver program.

jiayuasu (approved):

PR #1130, opened 2023-11-22
[SEDONA-434] Improve reliability by resolve the nondeterministic of the order of the Map

MyEnthusiastic:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-434

## Problem :

when the program reading from the newRdd, it assumes the order of the data to read into the Map https://github.com/apache/sedona/blob/a64972027aa4c0fea354898da67a2dd4e6cde46b/spark/common/src/test/java/org/apache/sedona/core/formatMapper/GeoJsonIOTest.java#L106

However, when it write the data, it did not assume the order according to the [Oracle's official document](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html#:~:text=This%20class%20makes%20no%20guarantees%20as%20to%20the%20order%20of%20the%20map)

https://github.com/apache/sedona/blob/a64972027aa4c0fea354898da67a2dd4e6cde46b/spark/common/src/main/java/org/apache/sedona/core/spatialRDD/SpatialRDD.java#L571-L572

So it will cause the ERROR when the environment has been changed

```
org.junit.ComparisonFailure: expected:<0[1]> but was:<0[23]>

	at org.junit.Assert.assertEquals(Assert.java:117)
	at org.junit.Assert.assertEquals(Assert.java:146)
	at org.apache.sedona.core.formatMapper.GeoJsonIOTest.testReadWriteGeoJson(GeoJsonIOTest.java:115)
```

# Solution

So we can change the Map to use the LinkedHashMap to guarantee the order to solve the problem 

## How was this patch tested?

This error can be reproduced with the [NondexTool](https://github.com/TestingResearchIllinois/NonDex) with the following command (you can try that without modifying the pom, but feel free to use the plugin to protect your project and make it safer 😊 ), and this kind problem is widely documented in [International Dataset of Flaky Tests (IDoFT)](https://github.com/TestingResearchIllinois/idoft)

- No, this PR does not affect any functionality public API, only improve the security by making the order deterministic

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -256,6 +256,7 @@
                 <groupId>it.geosolutions.jaiext.jiffle</groupId>
                 <artifactId>jt-jiffle-language</artifactId>
                 <version>${jt-jiffle.version}</version>
+                <scope>${geotools.scope}</scope>
                 <exclusions>
                     <exclusion>
                         <groupId>*</groupId>
--- a/spark/common/pom.xml
+++ b/spark/common/pom.xml
@@ -135,6 +135,10 @@
             <artifactId>gt-arcgrid</artifactId>
         </dependency>
         <dependency>
+            <groupId>it.geosolutions.jaiext.jiffle</groupId>
+            <artifactId>jt-jiffle-language</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.locationtech.jts</groupId>
             <artifactId>jts-core</artifactId>
         </dependency>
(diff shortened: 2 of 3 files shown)


Kontinuation:
Thank you for identifying a bug in the GeoJSON reader. While the proposed fix is appreciated, there is one more important issue we need to fix. The key issue is that the GeoJSON reader should have the capability to process GeoJSONs with properties in random orders within the same dataset.

Additionally, ensuring the consistent ordering of properties in GeoJSON files generated by Sedona is a valuable feature. It appears that this pull request addresses that aspect.

jiayuasu:
@MyEnthusiastic Thanks for the PR. We will address the other problem mentioned by @Kontinuation separately.

jiayuasu (approved):

jiayuasu (approved):

PR #1158, opened 2023-12-21
[SEDONA-453] Fix Quadtree index efficiency degrade when indexing points

Kontinuation:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-453. The PR name follows the format `[SEDONA-XXX] my subject`.

We found that the index efficiency of Quadtree drastically degrades when indexing datasets made up of points. The index returns way more candidates than expected when querying the Quadtree using envelopes. The reason is that JTS Quadtree automatically expands indexed envelopes by 0.5 if the envelope has zero width and height (see [Quadtree.java#L61-L96](https://github.com/locationtech/jts/blob/1.19.0/modules/core/src/main/java/org/locationtech/jts/index/quadtree/Quadtree.java#L61-L96)), this makes the indexed envelopes of points are way larger than necessary, especially when indexed points are WGS84 coordinates.

Suppose that we are indexing the following dataset using Quadtree:

Screenshot 2023-12-21 at 9 10 58 PM

The envelopes indexed by Quadtree happens to be something like this:

Screenshot 2023-12-21 at 9 12 16 PM

This PR workarounds this problem by manually extendinging envelopes with 0 width or height by 1e-3. This will prevent JTS Quadtree from extending the envelopes by 0.5, and 1e-3 is small enough to cope with the most common use cases. However, this will significantly increase the size of Quadtree since the tree will be deeper.

## How was this patch tested?

Add test to verify Quad tree index efficiency for PointRDD.

### DIFF
--- a/spark/common/src/main/java/org/apache/sedona/core/joinJudgement/DynamicIndexLookupJudgement.java
+++ b/spark/common/src/main/java/org/apache/sedona/core/joinJudgement/DynamicIndexLookupJudgement.java
@@ -123,7 +123,7 @@ public class DynamicIndexLookupJudgement<T extends Geometry, U extends Geometry>
             if (index instanceof Quadtree && (envelope.getWidth() == 0 || envelope.getHeight() == 0)) {
                 // Quadtree works poorly with envelopes with 0 extent. Such envelopes are usually obtained from
                 // points. We expand the envelope by a small amount to avoid this problem.
-                envelope.expandBy(1e-6);
+                envelope.expandBy(1e-3);
             }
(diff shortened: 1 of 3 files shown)


Kontinuation:
Manually reducing the size of the envelopes for points may significantly increase the size of the Quadtree. This may cause executors to run out of memory or exceed the kryo buffer limit when serializing the spatial index. STRtree is a better option for spatial indexing. We've already made STRtree as the default configuration in https://github.com/apache/sedona/pull/1159, so I'd like to keep the Quadtree indexing unchanged.

PR #1162, opened 2023-12-27
[SEDONA-429][SEDONA-430] Support specifying GeoParquet spec version number and CRS

Kontinuation:
**Note**: This PR depends on https://github.com/apache/sedona/pull/1161

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-429 and https://issues.apache.org/jira/browse/SEDONA-430. The PR name follows the format `[SEDONA-XXX] my subject`.

* Bumped the default GeoParquet version number from `1.0.0-beta.1` to `1.0.0`
* Allow specifying GeoParquet version number using `geoparquet.version` option
* Allow specifying CRS metadata for geometry columns using `geoparquet.crs` option

## How was this patch tested?

Add new tests for GeoParquet metadata.


docs/tutorial/sql.md
```diff
@@ -656,6 +656,30 @@ Since v`1.3.0`, Sedona natively supports writing GeoParquet file. GeoParquet can
 df.write.format("geoparquet").save(geoparquetoutputlocation + "/GeoParquet_File_Name.parquet")
 ```
 
+Since v`1.5.1`, Sedona supports writing GeoParquet files with custom GeoParquet spec version and crs.
```
jiayuasu, line 659: Please also add 

`You can find the projjson string of a specific CRS from here: https://epsg.io/ (click the JSON option at the bottom of the page. You can also customize your projjson string as needed.`

`Please note that Sedona currently cannot set/get a projjson string to/from a CRS. Its geoparquet reader will ignore the projjson metadata and you will have to set your CRS via ST_SetSRID after reading the file. Its geoparquet writer will not leverage the SRID field of a geometry so you will have to always set the geoparquet.crs option manually when writing the file.`

jiayuasu (changes requested):

spark/common/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/GeoParquetMetaData.scala
```diff
@@ -14,12 +14,7 @@
 package org.apache.spark.sql.execution.datasources.parquet
```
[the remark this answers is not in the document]
jiayuasu: Created a JIRA ticket for this: https://issues.apache.org/jira/browse/SEDONA-455

spark/spark-3.5/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/GeoParquetWriteSupport.scala
```diff
@@ -172,9 +186,10 @@ class GeoParquetWriteSupport extends WriteSupport[InternalRow] with Logging {
         val bbox = if (geometryTypes.nonEmpty) {
           Seq(columnInfo.bbox.minX, columnInfo.bbox.minY, columnInfo.bbox.maxX, columnInfo.bbox.maxY)
         } else Seq(0.0, 0.0, 0.0, 0.0)
-        columnName -> GeometryFieldMetaData("WKB", geometryTypes, bbox)
+        val crs = geoParquetColumnCrsMap.get(columnName).orElse(defaultGeoParquetCrs)
```
jiayuasu, line 189: According to the geoparquet spec, if a CRS is not given, then the crs field of the geoparquet should be omitted assuming this is CRS84. If a CRS is set null, then the crs field of the geoparquet should be set null.

The question is how a user can specify the null here. I think the user need to manually specify `geoparquet.crs` -> `null` via the options. What do you think?

jiayuasu (changes requested):

spark/spark-3.5/src/main/scala/org/apache/spark/sql/execution/datasources/parquet/GeoParquetWriteSupport.scala
Kontinuation: Currently the `crs` field is always present (written as `null`). This is because early version of geopandas (for example, 0.10.2 used by python tests) cannot read geoparquet files without `crs` metadata:

```python
>>> geopandas.read_parquet('gp_sample2.parquet')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/homebrew/lib/python3.11/site-packages/geopandas/io/arrow.py", line 461, in _read_parquet
    return _arrow_to_geopandas(table)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/geopandas/io/arrow.py", line 318, in _arrow_to_geopandas
    _validate_metadata(metadata)
  File "/opt/homebrew/lib/python3.11/site-packages/geopandas/io/arrow.py", line 162, in _validate_metadata
    raise ValueError(
ValueError: 'geo' metadata in Parquet/Feather file is missing required key 'crs' for column 'geometry'
```

`crs` field is optional in recent versions of GeoParquet standard, and setting it as `null` has a different meaning from omitting it. We can omit `crs` by default. This requires us to upgrade geopandas to 0.13.2, and drop support for Python 3.7 since geopandas dropped support for Python 3.7 since 0.11.

The value for `geoparquet.crs` or `geoparquet.crs.<column_name>` option user specified can be one of the following values:
* `""` (empty string): omit the `crs` metadata
* `"null"`: explicitly setting `crs` as `null`
* `"{...PROJJSON...}"`: explicitly setting `crs` to specified PROJJSON object

jiayuasu: This sounds good to me but GeoPandas > 0.11 requires shapely 2.0, and our other Python functions will not work in this case.

So maybe we still set the crs `null` by default and change the default behavior in the next major release 1.6.0, in which we will add Shapely 2.0 support?

Other parts look good to me:

```
The value for geoparquet.crs or geoparquet.crs.<column_name> option user specified can be one of the following values:

"" (empty string): omit the crs metadata
"null": explicitly setting crs as null
"{...PROJJSON...}": explicitly setting crs to specified PROJJSON object

```

Kontinuation: GeoPandas 0.13.2 works with shapely >=1.7.1, so it does not require shapely 2.0. I've confirmed that it works fine with shapely 1.8.5 and could pass all the tests after applying https://github.com/apache/sedona/pull/1163.

If we decide to upgrade GeoPandas and drop Python 3.7 in the next release, we can keep setting `crs` as `null` by default for now, and allow user to explicitly omit `crs` by setting `geoparquet.crs` to `""`.

jiayuasu: Yes, let's postpone the GeoPandas upgrade to 1.6.0. Let's keep setting crs as null by default for now, and allow user to explicitly omit crs by setting geoparquet.crs to ""

docs/tutorial/sql.md
```diff
@@ -656,6 +656,37 @@ Since v`1.3.0`, Sedona natively supports writing GeoParquet file. GeoParquet can
...
+The default GeoParquet spec version is `1.0.0` and the default crs is `null`. You can specify the GeoParquet spec version and crs as follows:
+
+```scala
+val projjson = "{...}" // PROJJSON string for all geometry columns
+df.write.format("geoparquet")
+		.option("geoparquet.version", "1.0.0")
+		.option("geoparquet.crs", projjson)
```
jiayuasu, line 666: Can you also add some text explaining the behavior of Sedona.

"geoparquet.crs","" will generate a geoparquet file without the `crs` field meaning this is CRS84
"geoparquet.crs","null" will generate a geoparquet file with `null` in its `crs` field meaning the CRS is unknown

jiayuasu (changes requested):

jiayuasu (approved):

PR #1175, opened 2024-01-01
[SEDONA-369] Add ST_DWithin

iGN5117:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/projects/SEDONA/issues/SEDONA-369 The PR name follows the format `[SEDONA-XXX] my subject`.

- Add ST_DWithin
- Add PredicatesTest.java for correctness test cases in Java
- Added a TestBase.java to reuse helpers between PredicatesTest and FunctionsTest.java

## How was this patch tested?
- Added new tests

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/Predicates.java
+++ b/common/src/main/java/org/apache/sedona/common/Predicates.java
@@ -49,4 +49,11 @@ public class Predicates {
     public static boolean orderingEquals(Geometry leftGeometry, Geometry rightGeometry) {
         return leftGeometry.equalsExact(rightGeometry);
     }
+    public static boolean dWithin(Geometry leftGeometry, Geometry rightGeometry, double distance) {
+        int leftSrid = leftGeometry.getSRID(), rightSrid = rightGeometry.getSRID();
+        if (leftSrid != rightSrid) {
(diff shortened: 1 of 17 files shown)


common/src/main/java/org/apache/sedona/common/Predicates.java
```diff
@@ -49,4 +49,11 @@ public static boolean disjoint(Geometry leftGeometry, Geometry rightGeometry) {
     public static boolean orderingEquals(Geometry leftGeometry, Geometry rightGeometry) {
         return leftGeometry.equalsExact(rightGeometry);
     }
+    public static boolean dWithin(Geometry leftGeometry, Geometry rightGeometry, double distance) {
+        int leftSrid = leftGeometry.getSRID(), rightSrid = rightGeometry.getSRID();
```
jiayuasu, line 53: I suggest that we don't perform CRS check in Predicates because all other predicates do not check it.

If we want to enforce this check, please raise another PR for this specific issue. Moreover, what is the default value of SRID if the user don't specify it at all? I think it is `0`? If so, this check will work. If it is `null`, then this check will not work.

jiayuasu (changes requested):
Just a reminder, since `ST_DWithIn` as a predicate, we need to support it in our optimized spatial join. You can address this issue in another PR.

jiayuasu (approved):

PR #1221, opened 2024-01-30
[SEDONA-475] Add RS_NormalizeAll

prantogg:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-475. The PR name follows the format `[SEDONA-XXX] my subject`.

Adds RS_NormalizeAll, to normalize values in all bands of a raster between a given normalization range. 
Key features - 
- **Flexible Normalization Range**: Optionally specify normalization range through `minLim` and `maxLim`. By default, the values are normalized to range [0, 255].
- **Handling Uniform Bands**: For bands where all pixel values are identical, normalized values are set to lower limit of normalization range.
- **Handling NoDataValues**: Allows defining `noDataValue` to be used for missing or invalid data in raster bands. By default, noDataValue is set to maxLim.
- **Optional Min/Max range of raster**: Min and Max range of the input raster's pixel values can be specified via `minValue` and `maxValue` to skip computation of them during runtime.
- **Data Type Integrity**: The function maintains the data type of raster values ensuring normalized values are cast back to the original raster's data type.
- **Normalization across bands**: `normalizeAcrossBands` flag, when set, normalization is performed across all bands based on global min and max values. If false, each band is normalized individually based on its own min and max values.

## How was this patch tested?

Passes new and existing tests

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/raster/MapAlgebra.java
+++ b/common/src/main/java/org/apache/sedona/common/raster/MapAlgebra.java
@@ -438,6 +438,45 @@ public class MapAlgebra
     }
 
     /**
+     *
+     * @param rasterGeom Raster to be normalized
+     * @return a raster with all values in all bands normalized between [0 - 255]
+     */
+    public static GridCoverage2D normalizeAll(GridCoverage2D rasterGeom) {
+        return normalizeAll(rasterGeom, 0d, 255d);
+    }
+
+    /**
+     *
+     * @param rasterGeom Raster to be normalized
+     * @param minLim Lower limit of normalization range
+     * @param maxLim Upper limit of normalization range
+     * @return a raster with all values in all bands normalized between minLim and maxLim
+     */
+    public static GridCoverage2D normalizeAll(GridCoverage2D rasterGeom, double minLim, double maxLim) {
+        int numBands = rasterGeom.getNumSampleDimensions();
+
+        for (int bandIndex = 1; bandIndex <= numBands; bandIndex++) {
+            // Get the band values as an array
+            double[] bandValues = bandAsArray(rasterGeom, bandIndex);
+
+            // Find min and max values in the band
+            double minValue = Arrays.stream(bandValues).min().orElse(Double.NaN);
+            double maxValue = Arrays.stream(bandValues).max().orElse(Double.NaN);
+
+            // Normalize the band values
+            for (int i = 0; i < bandValues.length; i++) {
+                bandValues[i] = minLim + ((bandValues[i] - minValue) * (maxLim - minLim)) / (maxValue - minValue);
+            }
+
+            // Update the raster with the normalized band
+            rasterGeom = addBandFromArray(rasterGeom, bandValues, bandIndex);
+        }
+
+        return rasterGeom;
+    }
+
+    /**
      * @param band1 band values
      * @param band2 band values
      * @return an array with the normalized difference of the provided bands
--- a/common/src/main/java/org/apache/sedona/common/raster/MapAlgebra.java
+++ b/common/src/main/java/org/apache/sedona/common/raster/MapAlgebra.java
@@ -454,6 +454,10 @@ public class MapAlgebra
      * @return a raster with all values in all bands normalized between minLim and maxLim
      */
     public static GridCoverage2D normalizeAll(GridCoverage2D rasterGeom, double minLim, double maxLim) {
+        if (minLim > maxLim) {
+            throw new IllegalArgumentException("minLim cannot be greater than maxLim");
+        }
+
         int numBands = rasterGeom.getNumSampleDimensions();
 
         for (int bandIndex = 1; bandIndex <= numBands; bandIndex++) {
--- a/common/src/main/java/org/apache/sedona/common/raster/MapAlgebra.java
+++ b/common/src/main/java/org/apache/sedona/common/raster/MapAlgebra.java
@@ -437,13 +437,16 @@ public class MapAlgebra
         return result;
     }
 
-    /**
-     *
-     * @param rasterGeom Raster to be normalized
-     * @return a raster with all values in all bands normalized between [0 - 255]
-     */
     public static GridCoverage2D normalizeAll(GridCoverage2D rasterGeom) {
-        return normalizeAll(rasterGeom, 0d, 255d);
+        return normalizeAll(rasterGeom, 0d, 255d, -9999, -99999, -99999);
+    }
+
+    public static GridCoverage2D normalizeAll(GridCoverage2D rasterGeom, double minLim, double maxLim) {
+        return normalizeAll(rasterGeom, minLim, maxLim, -9999, -99999, -99999);
+    }
+
+    public static GridCoverage2D normalizeAll(GridCoverage2D rasterGeom, double minLim, double maxLim, double noDataValue) {
+        return normalizeAll(rasterGeom, minLim, maxLim, noDataValue, -99999, -99999);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


common/src/main/java/org/apache/sedona/common/raster/MapAlgebra.java
```diff
@@ -437,6 +437,49 @@ public static double[] normalize(double[] band) {
...
+            // Get the band values as an array
+            double[] bandValues = bandAsArray(rasterGeom, bandIndex);
+
+            // Find min and max values in the band
+            double minValue = Arrays.stream(bandValues).min().orElse(Double.NaN);
+            double maxValue = Arrays.stream(bandValues).max().orElse(Double.NaN);
+
+            // Normalize the band values
+            for (int i = 0; i < bandValues.length; i++) {
+                bandValues[i] = minLim + ((bandValues[i] - minValue) * (maxLim - minLim)) / (maxValue - minValue);
```
jiayuasu, line 473: What will happen if all band values are the same. Then the denominator becomes zero and hence will lead to unexpected behavior.

common/src/test/java/org/apache/sedona/common/raster/MapAlgebraTest.java
```diff
@@ -321,6 +323,40 @@ public void testNormalize() {
...
+        GridCoverage2D normalizedRaster2 = MapAlgebra.normalizeAll(raster, 256d, 511d);
+
+        for (int band = 1; band <= 3; band++) {
+            double[] normalizedBand1 = MapAlgebra.bandAsArray(normalizedRaster1, band);
+            double[] normalizedBand2 = MapAlgebra.bandAsArray(normalizedRaster2, band);
+            double normalizedMin1 = Arrays.stream(normalizedBand1).min().getAsDouble();
+            double normalizedMax1 = Arrays.stream(normalizedBand1).max().getAsDouble();
+            double normalizedMin2 = Arrays.stream(normalizedBand2).min().getAsDouble();
+            double normalizedMax2 = Arrays.stream(normalizedBand2).max().getAsDouble();
+            double[] expected1 = {0.0, 17.0, 34.0, 51.0, 68.0, 85.0, 102.0, 119.0, 136.0, 153.0, 170.0, 187.0, 204.0, 221.0, 238.0, 255.0};
```
jiayuasu, line 348: 1. Please add tests for bands that have all same values.
2. Please add tests for rasters that have different pixel types. `NormalizeAll` is not supposed to change the pixel types. Moreover, the data truncation should happen as expected. If the input raster is integer, the normalized output raster should still have integer values, which means some decimal places derived from normalization will get truncated. One example is from MapAlgebra: https://sedona.apache.org/1.5.1/api/sql/Raster-map-algebra/#map-algebra

jiayuasu (changes requested):

common/src/main/java/org/apache/sedona/common/raster/MapAlgebra.java
jiayuasu: What if all band values are zero?

jiayuasu: What happens to the NoDataValue of a raster after normalization? @rbavery any experience?

jiayuasu: After discussing with @rbavery, we should exclude NoDataValue when performing the normalization. Moreover, we should provide an optional parameter for users to set the new NoDataValue.

common/src/main/java/org/apache/sedona/common/raster/MapAlgebra.java
```diff
@@ -437,6 +437,57 @@ public static double[] normalize(double[] band) {
...
+        int numBands = rasterGeom.getNumSampleDimensions();
+
+        for (int bandIndex = 1; bandIndex <= numBands; bandIndex++) {
+            // Get the band values as an array
+            double[] bandValues = bandAsArray(rasterGeom, bandIndex);
+
+            // Find min and max values in the band
+            double minValue = Arrays.stream(bandValues).min().orElse(Double.NaN);
+            double maxValue = Arrays.stream(bandValues).max().orElse(Double.NaN);
+            System.out.println("minValue: "+minValue);
```
jiayuasu, line 470: Please do not commit `System.out.println`

jiayuasu (changes requested):

common/src/main/java/org/apache/sedona/common/raster/MapAlgebra.java
```diff
@@ -437,6 +437,57 @@ public static double[] normalize(double[] band) {
...
+            System.out.println("maxValue: "+maxValue);
+
+            if (minValue == maxValue) {
+                // If all values in the band are same - set middle value of range as default value.
+                double defaultValue = (minLim + maxLim) / 2.0;
```
jiayuasu, line 475: If this happens, please choose `minLim` as the default value.

Please describe the behavior in the document.

common/src/main/java/org/apache/sedona/common/raster/MapAlgebra.java
```diff
@@ -437,6 +437,57 @@ public static double[] normalize(double[] band) {
...
+            // Get the band values as an array
+            double[] bandValues = bandAsArray(rasterGeom, bandIndex);
+
+            // Find min and max values in the band
+            double minValue = Arrays.stream(bandValues).min().orElse(Double.NaN);
+            double maxValue = Arrays.stream(bandValues).max().orElse(Double.NaN);
+            System.out.println("minValue: "+minValue);
+            System.out.println("maxValue: "+maxValue);
+
+            if (minValue == maxValue) {
```
jiayuasu, line 473: Please add 2 more optional parameters to this function which takes the minValue and maxValue from users. If the 2 values are not given, we compute the min and max ourselves.

jiayuasu (changes requested):
In short, this function should have the following variants:

`RS_NormalizeAll (raster: Raster)`

`RS_NormalizeAll (raster: Raster, minLim: Double, maxLim: Double)`

`RS_NormalizeAll (raster: Raster, minLim: Double, maxLim: Double, NoDataValue:Double)`

`RS_NormalizeAll (raster: Raster, minLim: Double, maxLim: Double, NoDataValue:Double, minValue:Double, maxValue:Double)`

common/src/main/java/org/apache/sedona/common/raster/MapAlgebra.java
prantogg: Sure. Have described the current behavior in the PR description above.

common/src/main/java/org/apache/sedona/common/raster/MapAlgebra.java
prantogg: @jiayuasu Just curious, why is taking minValue and MaxValue necessary? The current implementation i'm working on is able to compute minValue and maxValue excluding the NoDataValue.

jiayuasu: There will be cases that users know the expected value range of all band data in advance. In this case, finding the min max values ourselves does not make sense.

prantogg: To handle all pixel data types, would having a pixelType argument like in RS_MapAlgebra help?

jiayuasu: No need. Because this NormalizeAll function is not supposed to change the original pixel data type.

jiayuasu (changes requested):

common/src/main/java/org/apache/sedona/common/raster/MapAlgebra.java
```diff
@@ -437,6 +437,90 @@ public static double[] normalize(double[] band) {
         return result;
     }
 
+    public static GridCoverage2D normalizeAll(GridCoverage2D rasterGeom) {
+        return normalizeAll(rasterGeom, 0d, 255d, -9999, -99999, -99999);
```
jiayuasu, line 441: Please don't use magic numbers here. Can you use null?

common/src/main/java/org/apache/sedona/common/raster/MapAlgebra.java
```diff
@@ -437,6 +437,90 @@ public static double[] normalize(double[] band) {
...
+
+        int numBands = rasterGeom.getNumSampleDimensions();
+        RenderedImage renderedImage = rasterGeom.getRenderedImage();
+        int rasterDataType = renderedImage.getSampleModel().getDataType();
+
+        for (int bandIndex = 1; bandIndex <= numBands; bandIndex++) {
+            // Get the band values as an array
+            double[] bandValues = bandAsArray(rasterGeom, bandIndex);
+
+            // Find min and max values in the band, excluding NoDataValue
```
jiayuasu, line 478: You should fetch the `NoDataValue` from each band. There are operators functions in RasterUtils can already do it. The `NoDataValue` of the input is to be used when storing in the raster band.

common/src/main/java/org/apache/sedona/common/raster/MapAlgebra.java
```diff
@@ -437,6 +437,90 @@ public static double[] normalize(double[] band) {
...
+     * @param maxValue Maximum value in raster
+     * @return a raster with all values in all bands normalized between minLim and maxLim
+     */
+    public static GridCoverage2D normalizeAll(GridCoverage2D rasterGeom, double minLim, double maxLim, double noDataValue, double minValue, double maxValue) {
+        if (minLim > maxLim) {
+            throw new IllegalArgumentException("minLim cannot be greater than maxLim");
+        }
+
+        // Unset minmaxFlag if minValue and maxValue are not given;
+        boolean minmaxFlag = minValue != -99999;
```
jiayuasu, line 468: Please check `null` instead of a magic number.

common/src/main/java/org/apache/sedona/common/raster/MapAlgebra.java
```diff
@@ -437,6 +437,90 @@ public static double[] normalize(double[] band) {
...
+            }
+
+            if (minValue == maxValue) {
+                // Set default value for constant bands to minLim
+                Arrays.fill(bandValues, minLim);
+            } else {
+                // Normalize the band values, setting NoDataValue to maxLim
+                for (int i = 0; i < bandValues.length; i++) {
+                    if (bandValues[i] == noDataValue) {
+                        bandValues[i] = maxLim;
```
jiayuasu, line 491: If the band value equals to the no data value of a band, then you should set it as `noDataValue` from the input.

If the input `noDataValue` is not given, then set it to maxLim.

jiayuasu (changes requested):

jiayuasu (approved):

common/src/main/java/org/apache/sedona/common/raster/MapAlgebra.java
rbavery: @jiayuasu setting noDataValue to the maxLim by default will corrupt real data values.

instead I think the default should be to scale real values from 0, 254 and set the noDataValue to 255 if it is not given.

Or, scale real values between 1-255, set noDataValue to 0 by default. this makes more sense to me, 0 is used as a noDataValue more often than 255.

PR #1234, opened 2024-02-08
[SEDONA-493] Update default behavior of RS_NormalizeAll

prantogg:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-493. The PR name follows the format `[SEDONA-XXX] my subject`.

This PR, updates the default behavior of RS_NormalizeAll - 
- A safety mode is introduced for cases where the `noDataValue` is null, setting it to `maxLim` and normalizing data values to the range [minLim, maxLim-1] to prevent the loss of valid data that matches the new `noDataValue`.
- Additionally, refactors the function signatures to facilitate invoking safety mode for both normalization methods - individually on bands and across bands.

## How was this patch tested?


docs/api/sql/Raster-operators.md
```diff
@@ -2482,14 +2482,16 @@ SELECT RS_Normalize(band)
...
 - `minLim` and `maxLim` (Optional): The lower and upper limits of the normalization range. By default, normalization range is set to [0, 255].
-- `noDataValue` (Optional): Defines the value to be used for missing or invalid data in raster bands. By default, noDataValue is set to `maxLim`.
+- `noDataValue` (Optional): Defines the value to be used for missing or invalid data in raster bands. By default, noDataValue is set to `maxLim` and Safety mode is triggered.
 - `minValue` and `maxValue` (Optional): Optionally, specific minimum and maximum values of the input raster can be provided. If not provided, these values are computed from the raster data.
 - `normalizeAcrossBands` (Optional): A boolean flag to determine the normalization method. If set to true (default), normalization is performed across all bands based on global min and max values. If false, each band is normalized individually based on its own min and max values.
 
+A safety mode is triggered when `noDataValue` is null, setting it to `maxLim` and normalizing data values to the range [minLim, maxLim-1]. This is to avoid replacing valid data that might coincide with the new `noDataValue`.
```
jiayuasu, line 2493: Does your current implementation allow `null` as the input of a Sedona Spark (not just sedona-common)? Do we have a test?

If yes, then the sentence here should be `when `noDataValue` is null or not given). If no, it should be `when `noDataValue` is not given`

jiayuasu (changes requested):

docs/api/sql/Raster-operators.md
prantogg: This is an important consideration, thanks!
Current implementation allows `noDataValue` to be set to `null` in Sedona Spark as well. Confirmed this by testing it in Sedona-spark.
Have added tests for the same under Sedona-common. Would it be better to add tests under Sedona-spark as well?

prantogg: Accordingly updated the documentation too

jiayuasu: @prantogg https://github.com/apache/sedona/blob/master/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/raster/MapAlgebra.scala#L182

InferredExpression is null safe, which means it returns null directly if one of the input is null. Please double check.

In addition, `RS_NormarlizeAll` should not be put in `MapAlgebra.scala`. It should be put in `RasterEditors.scala`.

prantogg: My bad, I misunderstood

prantogg: InferredExpression is indeed null safe. Have updated the docs accordingly and moved `RS_NormalizeAll` to RasterEditors.

docs/api/sql/Raster-operators.md
```diff
@@ -1434,6 +1434,49 @@ Output:
...
+- `minValue` and `maxValue` (Optional): Optionally, specific minimum and maximum values of the input raster can be provided. If not provided, these values are computed from the raster data.
+- `normalizeAcrossBands` (Optional): A boolean flag to determine the normalization method. If set to true (default), normalization is performed across all bands based on global min and max values. If false, each band is normalized individually based on its own min and max values.
+
+A safety mode is triggered when `noDataValue` is not given. This sets `noDataValue` to `maxLim` and normalizes valid data values to the range [minLim, maxLim-1]. This is to avoid replacing valid data that might coincide with the new `noDataValue`.
+
+!!! Warning
+Using a noDataValue that falls within the normalization range can lead to loss of valid data. If any data value within a raster band matches the specified noDataValue, it will be replaced and cannot be distinguished or recovered later. Exercise caution in selecting a noDataValue to avoid unintentional data alteration.
```
jiayuasu, line 1450: There is no indention here. Doc website won't show this correctly.

jiayuasu (changes requested):

jiayuasu (approved):

PR #1251, opened 2024-02-22
[SEDONA-498] Add Spheroidal ST_Buffer

prantogg:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-498. The PR name follows the format `[SEDONA-XXX] my subject`.

- Adds Spheroidal support for `ST_Buffer`. Functions as follows,
  - Uses `ST_BestSRID` to determine the optimal SRID based on the geometry's extent and location.
  - Performs `ST_ShiftLongitude` on geometries crossing the IDL. `ST_CrossesDateLine` is used to determine if geometry crosses the IDL. 
  - Transforms the geometry to the selected SRID for buffering, defaulting to WGS 84 (SRID 4326) if no SRID is set.
  - Applies standard planar buffer operation in the chosen coordinate system.
  - Transforms the buffered geometry back to its original SRID, or to WGS 84 if the original SRID was not set.
  - Normalize longitudes of resultant geometry back to [-180, 180].
  - Behaves similar to PostGIS's `ST_Buffer(Geography)`.

- Updates ST_BestSRID as follows,
  - Handle geometries crossing the dateline. Performs `ST_ShiftLongitude` on geometries crossing the IDL determined by `ST_CrossesDateLine`
  - Update UTM zone calculation for geometries with shifted longitudes.

## How was this patch tested?


docs/api/sql/Function.md
```diff
@@ -561,6 +562,16 @@ The optional third parameter controls the buffer accuracy and style. Buffer accu
 - `mitre_limit=#.#` : mitre ratio limit and it only affects mitred join style. `miter_limit` is an accepted synonym for `mitre_limit`.
 - `side=both|left|right` : The option `left` or `right` enables a single-sided buffer operation on the geometry, with the buffered side aligned according to the direction of the line. This functionality is specific to LINESTRING geometry and has no impact on POINT or POLYGON geometries. By default, square end caps are applied.
 
+Mode of buffer calculation (Since: `v1.6.0`):
+
+- Planar Buffering (default): When `useSpheroid` is false, `ST_Buffer` performs standard planar buffering based on the provided parameters.
+- Spheroidal Buffering:
+    - When `useSpheroid` is set to true, the function returns the spheroidal buffer polygon for more accurate representation over the Earth.
+    - ST_Buffer first determines the most appropriate Spatial Reference Identifier (SRID) for a given geometry, based on its spatial extent and location, using `ST_BestSRID`.
```
jiayuasu, line 570: ST_BestSRID is only used when the input geom has no SRID or SRID == 4325. Please rephrase the sentences to make it clear.

In addition, what will happen to geometries that cross the dateline?

jiayuasu (changes requested):

common/src/main/java/org/apache/sedona/common/FunctionsGeoTools.java
```diff
@@ -134,4 +136,31 @@ public static Geometry voronoiPolygons(Geometry geom, double tolerance, Geometry
         }
         return builder.getDiagram(FunctionsGeoTools.GEOMETRY_FACTORY);
     }
+
+    public static Geometry bufferSpheroid(Geometry geometry, double radius, BufferParameters params){
+        // Determine the best SRID for spheroidal calculations
+        int bestCRS = Functions.bestSRID(geometry);
```
jiayuasu, line 142: Why not first check if the originalCRS is in WGS84 or not? If not, there is no need to call `BestSRID`. To make things neat, I recommend that we implement some sanity checks:

Sanity check Step 1 of `BestSRID`:

If the SRID is not given, we assume it is 4326.
If the SRID is given but is not 4326, we throw an exception saying that this function requires WGS84 / EPSG:4326 coordinates, and your SRID is XXX.

Sanity check Step 2 of `BestSRID`:

Given a geometry of 4326, its X must be in [-180, 180], its Y must be in [-90, 90]. If not, we throw an exception saying that this function requires WGS84 / EPSG:4326 coordinates but the bounding box of your geometry is XXXX.

docs/api/flink/Function.md
```diff
@@ -563,18 +565,31 @@ The optional third parameter controls the buffer accuracy and style. Buffer accu
...
+- Spheroidal Buffering:
+    - When `useSpheroid` is set to true, the function returns the spheroidal buffer polygon for more accurate representation over the Earth.
+    - ST_Buffer first determines the most appropriate Spatial Reference Identifier (SRID) for a given geometry, based on its spatial extent and location, using `ST_BestSRID`.
+    - The geometry is then transformed from its original SRID to the selected SRID. If the input geometry does not have a set SRID, `ST_Buffer` defaults to using WGS 84 (SRID 4326) as its original SRID.
+    - The standard planar buffer operation is then applied in this coordinate system.
+    - Finally, the buffered geometry is transformed back to its original SRID, or to WGS 84 if the original SRID was not set.
+
 !!!note
-    `ST_Buffer` throws an `IllegalArgumentException` if the correct format, parameters, or options are not provided.
+`ST_Buffer` throws an `IllegalArgumentException` if the correct format, parameters, or options are not provided.
```
jiayuasu, line 579: Pay attention to the indent here. This will not render properly.

docs/api/flink/Function.md
```diff
@@ -563,18 +565,31 @@ The optional third parameter controls the buffer accuracy and style. Buffer accu
...
 ST_Buffer (A: Geometry, buffer: Double, bufferStyleParameters: String [Optional])
 ```
+```
+ST_Buffer (A: Geometry, buffer: Double, bufferStyleParameters: String [Optional], useSperoidal: Boolean)
+```
 
 Since: `v1.5.1`
 
-Example:
+Spark SQL Example:
```
jiayuasu, line 592: Let's stop calling any example `Spark SQL example`, even if in Spark. Let's use `SQL Example` everywhere.

jiayuasu (changes requested):

common/src/main/java/org/apache/sedona/common/FunctionsGeoTools.java
prantogg: For Check 1 - 

> If the SRID is not given, we assume it is 4326.

This happens at [line 147](https://github.com/prantogg/sedona/blob/977ee844a87d589785feb8801252622aca278160/common/src/main/java/org/apache/sedona/common/FunctionsGeoTools.java#L147).


> If the SRID is given but is not 4326, we throw an exception saying that this function requires WGS84 / EPSG:4326 coordinates, and your SRID is XXX.

Having this exception would mean we disallow bestSRID to be determined for other Lon/lat coordinates systems such as NAD83 / EPSG:4269, ED50 (European Datum 1950) / EPSG:4230 and GDA94 / EPSG:4283

Check 2 can definitely be useful.

jiayuasu: @prantogg I thought Check 1 already disallowed NAD83 / EPSG:4269, ED50 (European Datum 1950) / EPSG:4230 and GDA94 / EPSG:4283?

Is the bestSRID supposed to be used for NAD83 / EPSG:4269, ED50 (European Datum 1950) / EPSG:4230 and GDA94 / EPSG:4283?

prantogg: Users that are sure about the bestSRID for their geometries can skip ST_Buffer(useSpheroid=true) altogether by manually utilizing ST_transform with ST_Buffer like so,
```
ST_Transform(ST_Buffer(ST_Transform(geometry, <original SRID>, <desired best SRID>), radius), <desired best SRID>, <original SRID>)
```

Maybe we could point this out in the documentation?

common/src/main/java/org/apache/sedona/common/Functions.java
```diff
@@ -87,25 +88,40 @@ public static Geometry boundary(Geometry geometry) {
...
+        // Processing parameters
+        if (!params.isEmpty()) {
+            bufferParameters = parseBufferParams(params);
 
-        // convert the sign to the appropriate direction
-        // left - radius should be positive
-        // right - radius should be negative
-        if (bufferParameters.isSingleSided() &&
-                (params.toLowerCase().contains("left") && radius < 0 || params.toLowerCase().contains("right") && radius > 0)) {
+            // convert the sign to the appropriate direction
```
jiayuasu, line 105: The indention is wrong

common/src/main/java/org/apache/sedona/common/Functions.java
```diff
@@ -310,110 +349,131 @@ public boolean isGeometryChanged() {
...
         double min = Double.MAX_VALUE;
-        for(int i=0; i < points.length; i++){
+        for (int i = 0; i < points.length; i++) {
             min = Math.min(points[i].getX(), min);
         }
         return min;
     }
 
-    public static double xMax(Geometry geometry) {
+    public static double xMax (Geometry geometry){
```
jiayuasu, line 403: Why does this PR fix the format of every single function of this file?

docs/api/flink/Function.md
```diff
@@ -571,10 +590,13 @@ Format:
 ```
 ST_Buffer (A: Geometry, buffer: Double, bufferStyleParameters: String [Optional])
 ```
+```
+ST_Buffer (A: Geometry, buffer: Double, bufferStyleParameters: String [Optional], useSperoidal: Boolean)
```
jiayuasu, line 594: `UseSpheroid`  parameter should be the 3rd param.

docs/api/snowflake/vector-data/Function.md
```diff
@@ -389,15 +389,75 @@ Output: `LINESTRING Z(-1 -1 0, 10 5 5)`
...
 
-SQL example:
-```SQL
-SELECT ST_Buffer(polygondf.countyshape, 1)
-FROM polygondf
 ```
+ST_Buffer (A: Geometry, buffer: Double, bufferStyleParameters: String [Optional])
+```
+```
+ST_Buffer (A: Geometry, buffer: Double, bufferStyleParameters: String [Optional], useSperoidal: Boolean)
```
jiayuasu, line 430: Same here. `UseSpheroid`  parameter should be the 3rd param.

spark/common/src/test/scala/org/apache/sedona/sql/functionTestScala.scala
```diff
@@ -86,6 +86,15 @@ class functionTestScala extends TestBaseScala with Matchers with GeometrySample
       assert(functionDf.count() > 0);
     }
 
+    it("Passed ST_Buffer Spheroid") {
+      val polygonWktDf = sparkSession.read.format("csv").option("delimiter", "\t").option("header", "false").load(mixedWktGeometryInputLocation)
+      polygonWktDf.createOrReplaceTempView("polygontable")
+      val polygonDf = sparkSession.sql("select ST_GeomFromWKT(polygontable._c0) as countyshape from polygontable")
+      polygonDf.createOrReplaceTempView("polygondf")
+      val functionDf = sparkSession.sql("select ST_Buffer(polygondf.countyshape, 1, true) from polygondf")
+      assert(functionDf.count() > 0);
```
jiayuasu, line 95: Please add a test that uses ST_Buffer with all 4 params

jiayuasu (changes requested):

common/src/main/java/org/apache/sedona/common/Functions.java
prantogg: The indentation is because the 'convert sign' logic is placed inside another if condition to process parameters.

common/src/main/java/org/apache/sedona/common/Functions.java
jiayuasu: @prantogg I would prefer that if you can roll back the format changes.

jiayuasu (changes requested):

docs/api/sql/Function.md
```diff
@@ -551,9 +551,28 @@ Output: `LINESTRING Z(-1 -1 0, 10 5 5)`
...
-Introduction: Returns a geometry/geography that represents all points whose distance from this Geometry/geography is less than or equal to distance.
+Introduction: Returns a geometry/geography that represents all points whose distance from this Geometry/geography is less than or equal to distance. The function supports both Planar/Euclidean and Spheroidal/Geodesic buffering (Since v1.6.0). Spheroidal buffer also supports geometries crossing the International Date Line (IDL).
 
-The optional third parameter controls the buffer accuracy and style. Buffer accuracy is specified by the number of line segments approximating a quarter circle, with a default of 8 segments. Buffer style can be set by providing blank-separated key=value pairs in a list format.
+Mode of buffer calculation (Since: `v1.6.0`):
+
+The optional third parameter, `useSpheroid`, controls the mode of buffer calculation.
+
+- Planar Buffering (default): When `useSpheroid` is false, `ST_Buffer` performs standard planar buffering based on the provided parameters.
+- Spheroidal Buffering:
```
jiayuasu, line 561: One last thing. Please add that, when `useSpheroid` is set to true, the unit of the buffer is `meter`.

jiayuasu (changes requested):

docs/api/sql/Function.md
prantogg: Done

jiayuasu (approved):

PR #1258, opened 2024-02-27
[SEDONA-507] RS_AsImage and RS_AsPNG could properly handle non-integral band data

Kontinuation:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-507. The PR name follows the format `[SEDONA-XXX] my subject`.

RS_AsImage only produces PNG images for rasters with integral band data, and produces TIFF images for rasters with floating point band data. TIFF cannot be displayed on many mainstream browsers, so we should not use TIFF to visualize rasters.

This PR make RS_AsPNG work for all types of band data, so that RS_AsImage won't produce TIFF base64 output.

* If the band data type is byte, we try converting it to PNG directly. This will preserve the original color of the raster if the raster is RGB or using a color table
* If the raster has one single band, we normalize the band data to 0~255 and generate a grayscale image, with nodata pixels masked as transparent
* Otherwise we use the color model to translate pixel values to RGB value, and generate an RGB image.

## How was this patch tested?

* Tested locally on various rasters to make sure that they are displayed properly.
* Added a smoke test to make sure that it does not break when processing rasters with various band data types

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/raster/RasterOutputs.java
+++ b/common/src/main/java/org/apache/sedona/common/raster/RasterOutputs.java
@@ -31,8 +31,15 @@ import org.opengis.parameter.ParameterValueGroup;
 
 import javax.imageio.ImageIO;
 import javax.imageio.ImageWriteParam;
+import javax.media.jai.InterpolationNearest;
+import javax.media.jai.JAI;
+import javax.media.jai.RenderedOp;
+import java.awt.image.BufferedImage;
+import java.awt.image.ColorModel;
 import java.awt.image.Raster;
 import java.awt.image.RenderedImage;
+import java.awt.image.WritableRaster;
+import java.awt.image.renderable.ParameterBlock;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.FileOutputStream;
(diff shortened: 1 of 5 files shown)


jiayuasu (approved):

jiayuasu:
@Kontinuation

1. With the `normalization` in this function, does a user still need to call `RS_NormalizeAll` to normalize an image (values not within 0 - 255) before calling `AsImage`?
2. When a raster has > 3 bands, I assume this will take the first 3 bands for visualization?

Kontinuation:
> @Kontinuation
> 
> 1. With the `normalization` in this function, does a user still need to call `RS_NormalizeAll` to normalize an image (values not within 0 - 255) before calling `AsImage`?
> 2. When a raster has > 3 bands, I assume this will take the first 3 bands for visualization?

1. No. The RS_AsImage will do the normalization by itself.
2. It depends on the color model of the raster. For RGB rasters it works properly, for rasters such as EuroSAT it does not work well. We may need to revisit how to deal with multi-band rasters later.

jiayuasu:
@Kontinuation What will happen exactly when it comes to EuroSat images?

Kontinuation:
> @Kontinuation What will happen exactly when it comes to EuroSat images?



> @Kontinuation What will happen exactly when it comes to EuroSat images?

The generated PNG image is mostly black, while visualizing one single band looks OK.
image

PR #1259, opened 2024-02-28
[SEDONA-508] Add ST_CrossesDateLine

prantogg:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-508. The PR name follows the format `[SEDONA-XXX] my subject`.

Introduces a predicate function, `ST_CrossesDateLine` to determines if a given geometry crosses the International Date Line.
- Checks if the difference in longitude between any pair of consecutive points in the geometry exceeds 180 degrees. If so, returns `true`, else `false`.
- For multi-geometries (e.g., MultiPolygon, MultiLineString), this function will return true if any one of the geometries within the multi-geometry crosses the International Date Line.

## How was this patch tested?


common/src/main/java/org/apache/sedona/common/Predicates.java
```diff
@@ -13,9 +13,11 @@
  */
```
jiayuasu, line 13: Please put this one in `Functions`. `Predicates` are for functions that take two geometries as input and return true/false.

common/src/main/java/org/apache/sedona/common/Predicates.java
```diff
@@ -62,4 +64,82 @@ public static boolean dWithin(Geometry leftGeometry, Geometry rightGeometry, dou
...
+     *
+     * @param geometry The geometry to check.
+     * @return True if the geometry crosses the Date Line, false otherwise.
+     */
+    public static boolean crossesDateLine(Geometry geometry) {
+        if (geometry == null || geometry.isEmpty()) {
+            return false;
+        }
+
+        AtomicBoolean crossesDateLine = new AtomicBoolean(false);
```
jiayuasu, line 79: Is there a reason why use `AtomicBoolean`? I believe the filter functions will be called in a sequential order?

common/src/main/java/org/apache/sedona/common/Predicates.java
```diff
@@ -62,4 +64,82 @@ public static boolean dWithin(Geometry leftGeometry, Geometry rightGeometry, dou
...
+                return crossesDateLine.get();
+            }
+
+            @Override
+            public boolean isGeometryChanged() {
+                return false;
+            }
+        };
+
+        if (geometry instanceof GeometryCollection) {
```
jiayuasu, line 110: Why explicitly do this for `GeometryCollection`? Note that: `GeometryCollection` can have nested `GeometryCollection` inside. We should avoid implementing our own logic to iterate GeometryCollection. If we really want to do it, this should be done via a recursive function.

common/src/main/java/org/apache/sedona/common/Predicates.java
```diff
@@ -62,4 +64,82 @@ public static boolean dWithin(Geometry leftGeometry, Geometry rightGeometry, dou
...
+            }
+        } else {
+            geometry.apply(filter);
+        }
+
+        return crossesDateLine.get();
+    }
+
+
+//    public static boolean crossesDateLine(Geometry geometry) {
```
jiayuasu, line 127: Please remove the comments.

jiayuasu (changes requested):

common/src/main/java/org/apache/sedona/common/Predicates.java
prantogg: The `CoordinateSequenceFilter` being an anonymous inner class could not modify a local variable outside its filter scope unless its `final`. `AtomicBoolean` was a work around because final variable's cannot be modified. I have removed separate handling of `GeometryCollection` so I can define a private local variable `crossesDateLine` within the filter.

common/src/main/java/org/apache/sedona/common/Predicates.java
prantogg: I initially wanted to make sure Every geometry within the `GeometryCollection` is handled separately because `crossesDateLine` should return true if any geometry is crossing the dateline.
You're right, explicit handling for `GeometryCollection` was not required, `CoordinateSequenceFilter` is able to handle this well. Have pushed the changes.

common/src/main/java/org/apache/sedona/common/Predicates.java
jiayuasu: @prantogg Can you address this?

prantogg: Done

jiayuasu (approved):

PR #1364, opened 2024-04-24
[SEDONA-543] Fixes RS_Union_aggr throwing referenceRaster is null error when run on cluster

prantogg:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-543. The PR name follows the format `[SEDONA-XXX] my subject`.

- In RS_Union_aggr, moves class level members like `referenceRaster` to data buffer.

## How was this patch tested?

- Passes existing tests

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/raster/serde/Serde.java
+++ b/common/src/main/java/org/apache/sedona/common/raster/serde/Serde.java
@@ -32,6 +32,8 @@ import org.opengis.referencing.operation.MathTransform;
 
 import javax.media.jai.RenderedImageAdapter;
 import java.awt.image.RenderedImage;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.Serializable;
 import java.net.URI;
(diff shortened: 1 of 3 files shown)


spark/common/src/main/scala/org/apache/sedona/sql/utils/RasterSerializer.scala
```diff
@@ -42,4 +43,18 @@ object RasterSerializer {
...
+    oos.writeObject(sampleDimension)
+    oos.flush()
+    bos.toByteArray
+  }
+
+  def deserializeSampleDimension(bytes: Array[Byte]): GridSampleDimension = {
+    val bis = new ByteArrayInputStream(bytes)
+    val ois = new ObjectInputStream(bis)
+    ois.readObject().asInstanceOf[GridSampleDimension]
+  }
```
Kontinuation, line 59: We have a `GridSampleDimensionSerializer` implemented using Kryo. I recommend using that instead of the Java serializer.

spark/common/src/main/scala/org/apache/sedona/sql/utils/RasterSerializer.scala
jiayuasu, line 1: Please don't commit this kine of empty change.

prantogg: Sorry I missed this

prantogg: This is fixed now.

common/src/main/java/org/apache/sedona/common/raster/serde/Serde.java
```diff
@@ -176,4 +178,19 @@ public static GridCoverage2D deserialize(byte[] bytes) throws IOException, Class
             return state.restore();
         }
     }
+
+    public static byte[] serializeGridSampleDimension(GridSampleDimension sampleDimension) {
```
jiayuasu, line 182: This was already implemented in Sedona. Why implement it again?

jiayuasu (changes requested):

common/src/main/java/org/apache/sedona/common/raster/serde/Serde.java
prantogg: This is as per Kristin's recommendation to implement new serialization/deserialization methods for GridSampleDimension that use `kryos.get()` to retrieve `Kryo` objects. `kryos.get()` avoids repeatedly initializing new Kryo instances.
I should have used the existing `GridSampleDimensionSerializer` for read and write into output buffer. This is added now.

jiayuasu (approved):

PR #1420, opened 2024-05-21
[SEDONA-559] Make the flink example work

docete:
The sedona examples project is not managed by parent pom file. 
It should be a submodule of the parent project and the flink example should work in IDE. 
This PR fix this.

 - b24e17d enable scalastyle config
 - 84ae117 make the flink example work


docete:
@jiayuasu IMO it should be included in the parent pom if it is placed in the same repo. Otherwise, it will not be well maintained. BTW, we should implement our use cases as the formal demonstration. What do you think?

jiayuasu:
The purpose of the example projects is to demo that you can build applications on top of the released version of Sedona. This is also useful for us to discover dependency bugs which cannot be discovered when bundled together with the parent pom.

We also have CI/CD to test the correctness of the example projects.

docete:
OK, I will unbundle the example projects and fix the build/run issues.

docete:
Check the ci/cd issues
`
[INFO] ------------------------------------------------------------------------
Error: Unable to access jarfile target/sedona-spark-example-1.0.0.jar
Error: Process completed with exit code 1.
`
Should we keep the example version and the sedona version consistent?

jiayuasu:
We could but we don't have to. What is your preference?

docete:
Let it consistent with SEDONA parent to keep it up to date with every release.

jiayuasu:
sounds good!

docete:
updated the github ci/cd script.

examples/pom.xml
```diff
@@ -0,0 +1,106 @@
...
+  <properties>
+    <log4j.version>2.17.2</log4j.version>
+    <slf4j.version>1.7.36</slf4j.version>
+    <geotools.version>28.2</geotools.version>
+    <flink.version>1.19.0</flink.version>
+    <scala.compat.version>2.12</scala.compat.version>
+  </properties>
+
+  <!-- add OSGeo Repositories for examples-->
+  <repositories>
```
jiayuasu, line 42: Please don't move these dependencies to the parent pom. I think we don't even need to have a parent project for the examples. The reasons are 

1. Spark and Flink have very different dependecies. It is hard to find a common set of log4j that work for all cases.
2. Each Spark/Flink example project should be as standalone as possible. This way, a new Sedona user can just pick an example project and use it as a template to develop his/her application.

A complicated project structure makes new user hard to get started.

jiayuasu (changes requested):

examples/pom.xml
docete: sounds good. removed the parent project.

jiayuasu (approved):

PR #1520, opened 2024-07-10
[SEDONA-625] Add ST_GeneratePoints

furqaankhan:

- Add ST_GeneratePoints

## How was this patch tested?

- Add new unit tests


jiayuasu (approved):

jiayuasu:
@Kontinuation The result of this function is non-deterministic. Will this create a problem to the catalyst optimizer?

Kontinuation:
> @Kontinuation The result of this function is non-deterministic. Will this create a problem to the catalyst optimizer?

The correct way to implement non-deterministic functions is to extend the `Nondeterministic` trait and implement the `initializeInternal` method to calculate the seed for each partition. The function evaluates using the initialized seed, so the results are identical when partitions are recomputed.

Unfortunately, `RandomPointsBuilder` does not allow constructing the builder with a given seed. We have to inherit this class and override the `createRandomCoord` function to generate coordinates using the specified seed.

jiayuasu:
@furqaankhan can you solve the issue using Kristin's method?

furqaankhan:
Yes, I am working on it.

jiayuasu:
This PR should also add SRID to the generated points following the style of this PR: https://github.com/apache/sedona/pull/1521

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala
```diff
@@ -1437,6 +1439,60 @@ case class ST_ForceRHR(inputExpressions: Seq[Expression])
...
+
+  override protected def evalInternal(input: InternalRow): Any = {
+    val geom = children.head.toGeometry(input)
+    val numPoints = children(1).eval(input).asInstanceOf[Int]
+    if (nArgs == 3) {
+      val seed = children(2).eval(input).asInstanceOf[Int]
+      return GeometrySerializer.serialize(Functions.generatePoints(geom, numPoints, seed))
+    }
+    GeometrySerializer.serialize(Functions.generatePoints(geom, numPoints))
+  }
```
Kontinuation, line 1472: We should construct an instance of `RandomPointsBuilderSeed` in the `initializeInternal` function, and call    `setExtent`, `setNumPoints`, `getGeometry` methods using that instance in `evalInternal`.

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala
```diff
@@ -1437,6 +1439,60 @@ case class ST_ForceRHR(inputExpressions: Seq[Expression])
...
 }
 
+case class ST_GeneratePoints(inputExpressions: Seq[Expression], randomSeed: Option[Long] = None)
+    extends Expression
+    with Nondeterministic
+    with ExpectsInputTypes
+    with CodegenFallback
+    with ExpressionWithRandomSeed {
+
+  def this(inputExpressions: Seq[Expression]) = this(inputExpressions, Some(0L))
```
Kontinuation, line 1449: `ExpressionWithRandomSeed` has binary incompatible changes since Spark 3.0, so we cannot use it to write binary compatible code for Spark 3.0 ~ 3.5. Maintaining compatibility with old Spark versions (<= 3.2) is really a burden for us.

We can initialize `ST_GeneratePoints` in the old Spark 2.3 way:

```
case class ST_GeneratePoints(inputExpressions: Seq[Expression], randomSeed: Long) extends RDG {

  def this(inputExpressions: Seq[Expression]) = this(inputExpressions, Utils.random.nextLong())
```

The consequence is that the random seed will be fixed for multiple minibatches when running streaming jobs. However, that's the best we can do to maintain compatibility with various Spark versions.

furqaankhan: Using RDG doesn't work, as extends `UnaryExpression`.

furqaankhan:
> This PR should also add SRID to the generated points following the style of this PR: #1521

I am using the factory of the input geometry. that will preserve the SRID. Please let me know if you think otherwise.

https://github.com/furqaankhan/sedona/blob/919136e2d67e730bd864f87c466a7f8be3b84407/common/src/main/java/org/apache/sedona/common/Functions.java#L1808

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala
Kontinuation: Ah, I made a mistake when writing the example code. It should be

```
case class ST_GeneratePoints(inputExpressions: Seq[Expression], randomSeed: Long) extends Expression
    with Nondeterministic
    with ExpectsInputTypes
    with CodegenFallback  {

  def this(inputExpressions: Seq[Expression]) = this(inputExpressions, Utils.random.nextLong())
}
```

furqaankhan: I was going to push this changes by removing the `ExpressionWithRandomSeed` trait and it's overridden members, to test it.

jiayuasu:
@furqaankhan Can you create a PR on snowflake-tester?

furqaankhan:
I did create one, I closed it as it had passed. I  reopened it as I added a new function signature. [The PR.](https://github.com/wherobots/sedona-snowflake-tester/pull/27)

common/src/main/java/org/apache/sedona/common/Functions.java
```diff
@@ -1812,6 +1812,24 @@ private static Geometry[] convertGeometryToArray(Geometry geom) {
...
+    pointsBuilder.setExtent(geom);
+    pointsBuilder.setNumPoints(numPoints);
+    return pointsBuilder.getGeometry();
+  }
+
+  public static Geometry generatePoints(Geometry geom, int numPoints) {
+    return generatePoints(geom, numPoints, 0);
+  }
+
+  public static Geometry generatePoints(Geometry geom, int numPoints, Random random) {
```
jiayuasu, line 1826: Why is there another overloaded function that takes `Random` as input?

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala
```diff
@@ -1435,6 +1436,58 @@ case class ST_ForceRHR(inputExpressions: Seq[Expression])
...
+
+  override protected def evalInternal(input: InternalRow): Any = {
+    val geom = children.head.toGeometry(input)
+    val numPoints = children(1).eval(input).asInstanceOf[Int]
+    val generatedPoints = if (nArgs == 3) {
+      val seed = children(2).eval(input).asInstanceOf[Int]
+      if (seed > 0) {
+        Functions.generatePoints(geom, numPoints, seed)
+      } else {
+        Functions.generatePoints(geom, numPoints, random)
```
jiayuasu, line 1462: Why is this needed? This if condition seems to be repeating the same condition in `Functions`

jiayuasu (changes requested):

common/src/main/java/org/apache/sedona/common/Functions.java
Kontinuation: This is for using the `random` member of `ST_GeneratePoints` to generate consistent random streams

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala
Kontinuation: This is for generating consistent random streams even if seed is 0.

jiayuasu (approved):

PR #1524, opened 2024-07-17
[SEDONA-629] Change RS Functions to return struct instead of array type

zhangfengcdt:

return structs for the following RS functions using the description as the key and statistic/ metadata name as the value

- RS_SummaryStatsAll
- RS_MetaData
- RS_ZonalStatsAll
- RS_GeoTransform

## How was this patch tested?
All relevant existing tests need to be refactored and passed.

- Yes, I have updated the documentation


docs/api/sql/Raster-operators.md
```diff
@@ -452,14 +452,14 @@ SELECT RS_GeoReferrence(ST_MakeEmptyRaster(1, 3, 4, 100.0, 200.0,2.0, -3.0, 0.1,
...
-Introduction: Returns an array of parameters that represent the GeoTransformation of the raster. The array contains the following values:
+Introduction: Returns a struct of parameters that represent the GeoTransformation of the raster. The struct has the following schema:
 
-- 0: pixel width along west-east axis (x-axis)
-- 1: pixel height along north-south axis (y-axis)
-- 2: Rotation of the raster
-- 3: Angular separation between x-axis and y-axis
-- 4: X ordinate of upper-left coordinate
-- 5: Y ordinate of upper-left coordinate
+- magnitudeI: pixel width along west-east axis (x-axis)
```
jiayuasu, line 457: Why is this called `I` and `J`?

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/raster/RasterFunctions.scala
```diff
@@ -0,0 +1,259 @@
+/*
```
jiayuasu, line 1: I suggest we should change the old implementation in place, instead of creating a new .Scala file for these special functions

jiayuasu (changes requested):

docs/api/sql/Raster-operators.md
zhangfengcdt: Looks like the original (before this change) descriptions are not correct and consistent with the comments in the code for geoTransform function. 

I am updating the document to use the ones from the comments.

```
- magnitudeI: size of a pixel along the transformed i axis
- magnitudeJ: size of a pixel along the transformed j axis
- thetaI: angle by which the raster is rotated (Radians positive clockwise)
- thetaIJ: angle from transformed i axis to transformed j axis (Radians positive counter-clockwise)
- offsetX: X ordinate of the upper-left corner of the upper-left pixel
- offsetY: Y ordinate of the upper-left corner of the upper-left pixel
```

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/raster/RasterFunctions.scala
zhangfengcdt: The reason why I create new scala wrappers is that if we change the Java function (RS_Metadata), since they are under the sedona-common module, it does not have dependency for spark. However, in order to return struct, we need 
```
import org.apache.spark.sql.types.{AbstractDataType, BooleanType, DataType, DoubleType, IntegerType, StructField, StructType}
```

I think it is not appropriate to introduce spark dependencies to the sedona-common module. The wrapper is also consistent with some other RS functions in PixelFunctions for example.

jiayuasu (approved):

PR #1526, opened 2024-07-18
[SEDONA-630] Improve ST_Union_Aggr performance

zhangfengcdt:

Switch to JTS `OverlayNGRobust.union` function to perform geometry union and add geometry cache capability.
https://locationtech.github.io/jts/javadoc/org/locationtech/jts/operation/overlayng/OverlayNGRobust.html

## How was this patch tested?
All existing unit tests should pass.

### DIFF
--- a/spark/common/src/test/scala/org/apache/sedona/sql/aggregateFunctionTestScala.scala
+++ b/spark/common/src/test/scala/org/apache/sedona/sql/aggregateFunctionTestScala.scala
@@ -18,6 +18,7 @@
  */
 package org.apache.sedona.sql
 
+import org.apache.spark.sql.DataFrame
 import org.apache.spark.sql.expressions.javalang.typed
 import org.apache.spark.sql.sedona_sql.expressions.ST_Union_Aggr
 import org.locationtech.jts.geom.{Coordinate, Geometry, GeometryFactory, Polygon}
@@ -70,10 +71,12 @@ class aggregateFunctionTestScala extends TestBaseScala {
     it("Measured ST_Union_aggr wall time") {
       // number of random polygons to generate
       val numPolygons = 1000
-      createPolygonDataFrame(numPolygons)
+      val df = createPolygonDataFrame(numPolygons)
+
+      df.createOrReplaceTempView("geometry_table")
 
       // cache the table to eliminate the time of table scan
-      sparkSession.sql("cache table geometry_table")
+      df.cache()
       sparkSession.sql("select count(*) from geometry_table").take(1)(0).get(0)
 
       // measure time for optimized ST_Union_Aggr
@@ -85,6 +88,9 @@ class aggregateFunctionTestScala extends TestBaseScala {
       val durationOptimized = endTimeOptimized - startTimeOptimized
 
       assert(durationOptimized > 0, "Duration of optimized ST_Union_Aggr should be positive")
+
+      // clear cache
+      df.unpersist()
     }
 
     it("Passed ST_Intersection_aggr") {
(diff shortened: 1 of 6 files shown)


zhangfengcdt:
@jiayuasu I noticed that after switching from geo.buffer to OverlayNGRobust.union, the complex geometry representation returned from ST_Union_Aggr might change due to the reordering of polygon/polyline vertex. For example:

New: POLYGON ((1 0, 0 0, 0 1, 1 1, 2 1, 2 0, 1 0))

Old: POLYGON ((0 0, 0 1, 1 1, 2 1, 2 0, 1 0, 0 0))

They represent the same polygon, but the vertex order has changed.

jiayuasu:
> @jiayuasu I noticed that after switching from geo.buffer to OverlayNGRobust.union, the complex geometry representation returned from ST_Union_Aggr might change due to the reordering of polygon/polyline vertex. For example:
> 
> New: POLYGON ((1 0, 0 0, 0 1, 1 1, 2 1, 2 0, 1 0))
> 
> Old: POLYGON ((0 0, 0 1, 1 1, 2 1, 2 0, 1 0, 0 0))
> 
> They represent the same polygon, but the vertex order has changed.

I think this is fine.

In addition, we want to make sure the behavior of ST_Union_Aggr is similar to PostGIS ST_Union (array variant): https://postgis.net/docs/ST_Union.html

jiayuasu (commented):
@zhangfengcdt Did you see performance improvement using this implementation, compared to the previous one?

zhangfengcdt:
> @zhangfengcdt Did you see performance improvement using this implementation, compared to the previous one?

Yeah, I am adding some tests to report the performance measure and we can see the improvements for different cases there.

zhangfengcdt:
> @zhangfengcdt Did you see performance improvement using this implementation, compared to the previous one?

@jiayuasu I have used the newly added test to measure both old and new runtime for different number of geometries. Here are the results:

```
 Number of Polygons   |  OLD ST_Union_Aggr (in ms) |  NEW ST_Union_Aggr (in ms)
 -------------------------------------------------------------------------------
 100                   |             297                |              354     
 500                   |             750                |              386
 1,000                |           2,231                |              430
 5,000                |          53,870              |            1,400
 10,000               |         243,465            |            3,474
```

I think it shows clearly the new method is much efficient and scalable.

spark/common/src/test/scala/org/apache/sedona/sql/aggregateFunctionTestScala.scala
```diff
@@ -97,4 +122,24 @@ class aggregateFunctionTestScala extends TestBaseScala {
...
+    val polygons = (1 to numPolygons).map(generateRandomPolygon).toArray
+    val polygonArray = polygons.map(polygon => s"ST_GeomFromWKT('$polygon')")
+    val polygonArrayStr = polygonArray.mkString(", ")
+
+    val sqlQuery =
+      s"""
+         |SELECT explode(array($polygonArrayStr)) AS geom
+     """.stripMargin
+
+    sparkSession.sql(sqlQuery).createOrReplaceTempView("geometry_table")
```
jiayuasu, line 143: Can you return a reference of the DF as the return value of the function, instead of creating a new temp view? Otherwise this might pollute the global namespace and lead to bugs that are hard to find.

spark/common/src/test/scala/org/apache/sedona/sql/aggregateFunctionTestScala.scala
```diff
@@ -62,6 +67,26 @@ class aggregateFunctionTestScala extends TestBaseScala {
       assert(union.take(1)(0).get(0).asInstanceOf[Geometry].getArea == 10100)
     }
 
+    it("Measured ST_Union_aggr wall time") {
+      // number of random polygons to generate
+      val numPolygons = 1000
+      createPolygonDataFrame(numPolygons)
+
+      // cache the table to eliminate the time of table scan
+      sparkSession.sql("cache table geometry_table")
```
jiayuasu, line 76: Can you also unpersist this table at the end of the test case? Otherwise this will lead to memory leak.

jiayuasu (changes requested):

jiayuasu (approved):

PR #1527, opened 2024-07-19
[SEDONA-631] Add ST_Expand

furqaankhan:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-631. The PR name follows the format `[SEDONA-XXX] my subject`.

- Add ST_Expand

## How was this patch tested?

- add new unit tests


common/src/main/java/org/apache/sedona/common/utils/BBox.java
```diff
@@ -38,6 +37,64 @@ public BBox(double startLon, double endLon, double startLat, double endLat) {
...
+    startLon = Double.MAX_VALUE;
+    endLon = -Double.MAX_VALUE;
+    startLat = Double.MAX_VALUE;
+    endLat = -Double.MAX_VALUE;
+    minZ = Double.MAX_VALUE;
+    maxZ = -Double.MAX_VALUE;
+    processEnvelope(geom);
+  }
+
+  private void processEnvelope(Geometry geom) {
```
jiayuasu, line 50: Why is this so complicated? You can see how existing approach read the coordinates https://github.com/apache/sedona/blob/master/common/src/main/java/org/apache/sedona/common/Functions.java#L495

jiayuasu (changes requested):

common/src/main/java/org/apache/sedona/common/Functions.java
```diff
@@ -90,6 +93,57 @@ public static Geometry boundary(Geometry geometry) {
...
+    double minZ = Double.MAX_VALUE;
+    double maxZ = -Double.MAX_VALUE;
+
+    for (int i = 0; i < coordinates.length; i++) {
+      minX = Math.min(minX, coordinates[i].x);
+      maxX = Math.max(maxX, coordinates[i].x);
+      minY = Math.min(minY, coordinates[i].y);
+      maxY = Math.max(maxY, coordinates[i].y);
+      minZ = Math.min(minZ, coordinates[i].z);
+      maxZ = Math.max(maxZ, coordinates[i].z);
```
jiayuasu, line 123: When a Geometry does not have Z value, JTS will give NaN (https://locationtech.github.io/jts/javadoc/org/locationtech/jts/geom/Coordinate.html#getZ--). Does the `Math` comparison still work in this case?

docs/api/sql/Function.md
```diff
@@ -1216,6 +1216,38 @@ Output:
 POLYGON ((0 0, 0 3, 1 3, 1 0, 0 0))
 ```
 
+## ST_Expand
+
+Introduction: Returns a geometry expanded from the bounding box of the input. The expansion can be specified in two ways:
```
jiayuasu, line 1221: Please say this only deal with X, Y, Z value.

If `uniformDelta` is used and the geometry has `Z`, all X, Y, Z will be expanded. If only `deltaX` and `deltaY` are specified, only X and Y will be expanded even if the geometry has `Z`

jiayuasu (changes requested):

common/src/main/java/org/apache/sedona/common/Functions.java
furqaankhan: Yeah, it just returns a NaN, and I am skipping the Z values when Geometry doesn't have it. I don't see it being a problem.

jiayuasu (changes requested):
Can you add a test here? https://github.com/furqaankhan/sedona/blob/st-expand/spark/common/src/test/scala/org/apache/sedona/sql/PreserveSRIDSuite.scala

jiayuasu (approved):

PR #1543, opened 2024-08-07
[SEDONA-636] Support GEOMETRY in Spark SQL table creation DDL

zhangfengcdt:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-XXX. The PR name follows the format `[SEDONA-636] my subject`.

This PR add the geometry type in the table creation DDL. For example:
```
CREATE TABLE T_TEST_EXPLICIT_GEOMETRY (GEO_COL GEOMETRY)

CREATE TABLE T_TEST_EXPLICIT_GEOMETRY_2 (INT_COL INT, GEO_COL GEOMETRY)
```
Before the change, it throws error below:

DataType geometryee is not supported.(line 1, pos 62)

== SQL ==
CREATE TABLE T_TEST_EXPLICIT_GEOMETRY_2 (INT_COL INT, GEO_COL GEOMETRY)
--------------------------------------------------------------^^^

## How was this patch tested?
new unit test: SQLSyntaxTestScala


zhangfengcdt:
There are breaking changes to the SparkSqlParser and SparkSqlAstBuilder, that removed the configuration from the constructor (committed on Nov 14, 2020 ). The spark 3.0.3 thus failed.

https://github.com/apache/spark/commit/156704ba0dfcae39a80b8f0ce778b73913db03b2

jiayuasu:
What happens to the `spark-3.0` profile?

I understand that it could create some redundant code but I believe we should still replicate the tests to all profiles. What do you think? @zhangfengcdt

zhangfengcdt:
> What happens to the `spark-3.0` profile?
> 
> I understand that it could create some redundant code but I believe we should still replicate the tests to all profiles. What do you think? @zhangfengcdt

@jiayuasu I've just added the unit tests to all spark versions we currently support. For spark-3.0, I also refactor the custom sedona sql parser constructor to take the extra `conf` as the legacy spark codebase requires.

jiayuasu (approved):

PR #1589, opened 2024-09-17
[SEDONA-655] DBSCAN

james-willis:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-655. The PR name follows the format `[SEDONA-XXX] my subject`.

This PR add a DBSCAN function in the scala and python APIs of the spark implementation of sedona.

## How was this patch tested?
unit tests


jiayuasu:
@james-willis The memory consumption of new tests seem to be very high?

james-willis:
> @james-willis The memory consumption of new tests seem to be very high?

This is something I had worked with Kristin on in the past. We had thought it was related to cached dataframes holding references to broadcast relations. But this committed version of the code does not contain any caching, just checkpoints. 

There is some persisting inside of the connected component implementation, but those should be getting cleaned up. I'll check if something here is leaking persisted dataframes

james-willis:
Theres some dataframes inside of the connected components algo that dont get unpersisted. I will look into raising a PR in that repo tomorrow to fix that.

james-willis:
I submitted a fix to graphframes for this. I can clear the spark catalog if the disable broadcast fix doesnt resolve

https://github.com/graphframes/graphframes/pull/459

jiayuasu:
what are the remaining tasks for this PR?

james-willis:
just documentation

zwu-net:
> just documentation

After you guys (@james-willis @jiayuasu)  finish this, I'll write an article on this.

python/tests/stats/test_dbscan.py
```diff
@@ -0,0 +1,252 @@
...
+        try:
+            dbscan(df, -0.1, 5, "arealandmark")
+            assert False
+        except Exception:
+            assert True
+
+    def test_dbscan_invalid_min_pts(self):
+        df = self.create_sample_dataframe()
+
+        try:
```
Imbruced, line 174: I think that's equivalent 
```python
with pytest.raises(Exception):
    dbscan(df, 0.1, -5, "arealandmark")
```

spark/common/src/main/scala/org/apache/sedona/stats/Util.scala
```diff
@@ -0,0 +1,41 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.sedona.stats
+
+import org.apache.spark.sql.DataFrame
+import org.apache.spark.sql.sedona_sql.UDT.GeometryUDT
+
+private[stats] object Util {
```
Imbruced, line 24: what do you think about keeping code to the left ? 

```scala
def getGeometryColumnName(dataframe: DataFrame): String = {
    val geomFields = dataframe.schema.fields.filter(_.dataType == GeometryUDT)
    
    if (geomFields.length == 1) {
      geomFields.head.name
    }
    
    if (geomFields.isEmpty) {
      throw new IllegalArgumentException("No GeometryType column found. Provide a dataframe containing a geometry column.")
    }

    if (geomFields.length > 1 && !geomFields.exists(_.name == "geometry")) {
      throw new IllegalArgumentException(
        "Multiple GeometryType columns found. Provide the column name as an argument.")
    }

    "geometry"
  }
```

I think that way is more readable

python/tests/stats/test_dbscan.py
```diff
@@ -0,0 +1,252 @@
...
+        result.unpersist()
+
+        clusters = {
+            frozenset([y[0] for y in clusters_members if y[1] == x])
+            for x in set([y[1] for y in clusters_members])
+        }
+
+        return clusters
+
+    def test_dbscan_valid_parameters(self):
```
Imbruced, line 93: I think it might be a good use case for parametrized tests 
https://docs.pytest.org/en/7.1.x/how-to/parametrize.html

I mean you can try to create cases based on those loops 
      for epsilon in [0.6, 0.7, 0.8]:
            for min_pts in [3, 4, 5]:

python/tests/stats/test_dbscan.py
```diff
@@ -0,0 +1,252 @@
...
+from sedona.sql.st_functions import ST_Buffer
+from sklearn.cluster import DBSCAN as sklearnDBSCAN
+from sedona.stats.clustering.dbscan import dbscan
+
+from tests.test_base import TestBase
+
+
+class TestDBScan(TestBase):
+
+    def get_data(self):
```
Imbruced, line 30: that might be fixture 
https://docs.pytest.org/en/6.2.x/fixture.html

sth like 

```python
@pytest.fixture
def sample_db_scan_data():
    ...
```
and then for each function
```python
def test_sth(sample_db_scan_data):
```

python/sedona/stats/utils/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+#  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+
+from pyspark.sql import DataFrame, Column, SparkSession
+from sedona.sql.types import GeometryType
+
+
+def get_geometry_column_name(df: DataFrame) -> Column:
```
Imbruced, line 22: I think here you can also keep happy path to the left as in the scala :)

spark/common/src/main/scala/org/apache/sedona/stats/clustering/DBSCAN.scala
```diff
@@ -0,0 +1,153 @@
...
+  def dbscan(
+      dataframe: DataFrame,
+      epsilon: Double,
+      min_pts: Int,
+      geometry: String = null,
+      includeOutliers: Boolean = true,
+      useSpheroid: Boolean = false): DataFrame = {
+
+    // We want to disable broadcast joins because the broadcast reference were using too much driver memory
+    val spark = SparkSession.getActiveSession.get
```
Imbruced, line 64: is here possible to have an empty session ?

spark/common/src/main/scala/org/apache/sedona/stats/clustering/DBSCAN.scala
```diff
@@ -0,0 +1,153 @@
...
+    val distanceFunction: (Column, Column) => Column =
+      if (useSpheroid) ST_DistanceSpheroid else ST_Distance
+
+    val hasIdColumn = dataframe.columns.contains("id")
+    val idDataframe = if (hasIdColumn) {
+      dataframe
+        .withColumnRenamed("id", ID_COLUMN)
+        .withColumn("id", sha2(to_json(struct("*")), 256))
+    } else {
+      dataframe.withColumn("id", sha2(to_json(struct("*")), 256))
```
Imbruced, line 81: duplicated records should be equal and aggregated or not ? I am wondering if here `monotonically_increasing_id ` function is enough ? I think that might by costly `sha2(to_json(struct("*"))` also as key for join might not be super efficient compared to bigints ?

james-willis: monotonically increasing id is not deterministic so it can give incorrect results in joins unless you immediately checkpoint after generating the ids. This blog does a decent job of describing the issue: https://xebia.com/blog/spark-surprises-for-the-uninitiated/ I've been bitten by this issue before, espcially when an executor crashes.

> duplicated records should be equal and aggregated or not ?

This implementation does not deal with duplicated records well. There is only a comment in the doc string saying dont provide duplicates. monotonically_increasing_id + checkpoint would handle duplicates.

spark/common/src/main/scala/org/apache/sedona/stats/clustering/DBSCAN.scala
james-willis: Only if the session crashed in some other test. The parent class initializes the spark session.

docs/api/stats/sql.md
```diff
@@ -0,0 +1,23 @@
+## Overview
+The stats module of Sedona implements Scala and Python functions that can be called on dataframes with spatial columns to perform geospatial statistical analysis. The stats module is built on top of the core module and provides a set of functions that can be used to perform spatial analysis on dataframes. The stats module is designed to be used with the core module and the viz module to provide a complete set of geospatial analysis tools.
```
kadolor, line 2:
```suggestion
Sedona's stats module provides Scala and Python functions for conducting geospatial
statistical analysis on dataframes with spatial columns. 
The stats module is built on top of the core module and provides a set of functions
that can be used to perform spatial analysis on these dataframes. The stats module
is designed to be used with the core module and the viz module to provide a
complete set of geospatial analysis tools.
```

docs/api/stats/sql.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Using DBSCAN
+The DBSCAN function is provided at `org.apache.sedona.stats.DBSCAN.dbscan` in scala/java and `sedona.stats.dbscan.dbscan` in python.
+
+The function annotates a dataframe with a cluster label for each data record using the DBSCAN algorithm.
+The dataframe should contain at least one GeometryType column. Rows must be unique. If one
```
kadolor, line 8:
```suggestion
The dataframe should contain at least one `GeometryType` column. Rows must be unique. If one
```

docs/api/stats/sql.md
jiayuasu, line 1: This file is not put in the `mkdocs.yml` hence it will show up on the website navigation bar.

jiayuasu (changes requested):

docs/api/stats/sql.md
james-willis: on it. and fixing test failures.

jiayuasu (approved):

PR #1706, opened 2024-11-29
[SEDONA-636] fix some bugs when sedona parser running with other parser

freamdx:
* fix some blocking problem when running sql concurrently
* fix some error when setting "spark.sql.extensions" with different order
* improve performance


spark/common/src/main/scala/org/apache/sedona/sql/ParserRegistrator.scala
```diff
@@ -38,21 +38,6 @@ object ParserRegistrator {
...
-      val delegate: ParserInterface = sparkSession.sessionState.sqlParser
-
-      val parser =
-        ParserFactory.getParser(parserClassName, sparkSession.sessionState.conf, delegate)
-      val field = sparkSession.sessionState.getClass.getDeclaredField("sqlParser")
-      field.setAccessible(true)
-      field.set(sparkSession.sessionState, parser)
     } catch {
       case _: Exception =>
     }
```
Kontinuation, line 43: ParserRegistrator.scala is not needed anymore, as for now we register parsers using the extension interface instead of using the reflection trick. We can delete this file.

spark/common/src/main/scala/org/apache/sedona/spark/SedonaContext.scala
```diff
@@ -65,9 +65,6 @@ object SedonaContext {
     RasterRegistrator.registerAll(sparkSession)
     UdtRegistrator.registerAll()
     UdfRegistrator.registerAll(sparkSession)
-    if (sparkSession.conf.get("spark.sedona.enableParserExtension", "true").toBoolean) {
-      ParserRegistrator.register(sparkSession)
-    }
```
Kontinuation, line 70: NOTE: This introduces a behavior change. Now setting up `spark.sql.extensions` is required for enabling the parser extension. I think this is a reasonable change and I'm OK with it.

jiayuasu:
@freamdx can you address the comment from @Kontinuation ? Thanks! Then I will merge the PR.

freamdx:
> @freamdx can you address the comment from @Kontinuation ? Thanks! Then I will merge the PR.

done

jiayuasu:
@freamdx 

Can you do this? Thanks!
> ParserRegistrator.scala is not needed anymore, as for now we register parsers using the extension interface instead of using the reflection trick. We can delete this file.

freamdx:
> @freamdx
> 
> Can you do this? Thanks!
> 
> > ParserRegistrator.scala is not needed anymore, as for now we register parsers using the extension interface instead of using the reflection trick. We can delete this file.

done

jiayuasu (approved):

freamdx:
@jiayuasu 
can you release a new version before 2024/12/31? such as 1.7.1, i need this patch.

PR #1735, opened 2024-12-22
[WIP] add canonical links to docs

### LINKED ISSUE #1730: Add canonical links to the documentation pages
## Expected behavior

For all the versioned documentation pages to have canonical links to the latest documentation pages.

## Actual behavior

Canonical links are missing, which causes SEO issues.

## Example of problem

When I do a google search like "sedona ST_DistanceSphere", the 1.4.1 docs are the first result: https://sedona.apache.org/1.4.1/api/sql/Function/

We can fix this by adding canonical links to the pages.

We can just add this to the page: 

```
<link rel="canonical" href="https://sedona.apache.org/latest/api/sql/Function/" />
```

MrPowers:
Work in progress!

Fixes #1730.

Opened up an issue with mike cause it's potentially an issue on their end: https://github.com/jimporter/mike/issues/235

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -174,6 +174,8 @@ theme:
 extra:
   version:
     provider: mike
+    default:
+      - latest
   social:
     - icon: fontawesome/brands/github-alt
       link: 'https://github.com/apache/sedona'
@@ -225,3 +227,5 @@ plugins:
       type: datetime
   - mkdocs-jupyter:
       include_source: True
+  - mike:
+      canonical_version: "latest"


MrPowers:
I added the `site_url` so I think this should be working now, but someone should double check!

I only think this will fix the docs on a go-forward basis.  We'd need to roll out some sort of hotfix if we wanted to add canonical links to all the existing docs sites.

jiayuasu:
@MrPowers is this ready to go?

PR #1751, opened 2025-01-10
[WIP] Expose spatial partitioning from SpatialRDD

paleolimbot:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-695. The PR name follows the format `[SEDONA-695] my subject`.

Closes https://github.com/apache/sedona/issues/1268.

This PR exposes spatial partitioning information from the SpatialRDD API. Sedona is exceptionally good at this and the spatial community would love to have access to this information!

There are two pieces of information that would be helpful:

- The actual boundaries
- A partitioned RDD that remembers the partition identifier (i.e., partitioned results).

There are a few ideas in this PR...the boundaries seem straightforward but I'm a little new to the RDD API to know what the options are for returning these things.

## How was this patch tested?

Working on it!

- Yes, I have updated the documentation. (Or will when the API is settled)


spark/common/src/main/java/org/apache/sedona/core/spatialRDD/SpatialRDD.java
```diff
@@ -326,6 +327,25 @@ public void remove() {
...
+        .flatMapToPair(
+            new PairFlatMapFunction<T, Integer, T>() {
+              @Override
+              public Iterator<Tuple2<Integer, T>> call(T spatialObject) throws Exception {
+                return partitioner.placeObject(spatialObject);
+              }
+            })
+        .partitionBy(partitioner);
+  }
+
```
paleolimbot, line 348: This is probably not needed (the changes in the `Adapter` to preserve the partitioning of `spatialPartitionedRDD` into the output data frame should eliminate the need to keep any identifier alongside the partition).

spark/common/src/main/scala/org/apache/sedona/sql/utils/Adapter.scala
```diff
@@ -162,10 +162,14 @@ object Adapter {
       spatialRDD: SpatialRDD[T],
       schema: StructType,
       sparkSession: SparkSession): DataFrame = {
-    val rdd = spatialRDD.rawSpatialRDD.rdd.map[Row](geom => {
-      val stringRow = extractUserData(geom)
-      castRowToSchema(stringRow = stringRow, schema = schema)
-    })
+    val rdd = spatialRDD.rawSpatialRDD.rdd.mapPartitions(
```
paleolimbot, line 165: (But moved to a different overload since most of the time this will introduce duplicates)

```suggestion
    val rdd = spatialRDD.spatialPartitionedRDD.rdd.mapPartitions(
```

paleolimbot (commented):
This also would benefit from a `SpatialPartitioner` that removes duplicates (perhaps by wrapping a SpatialPartitioner, consuming the result of `placeObject` and deterministically choosing one of the results), since most of the time having duplicates when partitioning is not really desired.

james-willis:
Would it be better to add a preservePartitions arguments to the toDf method?

paleolimbot (commented):
For this to be universally useful (and to remove the warning in code and in the docs), we'll need partitioners that don't introduce duplicates. I'm happy to do that here or in another PR (with my preference being another PR since the changes are largely orthogonal).

> Would it be better to add a preservePartitions arguments to the toDf method?

@james-willis I didn't forget about this! I'm happy to defer to anything here...I added it as a separate function because there were already a lot of `toDf` overloads and I wasn't sure if there would have to be any more options for `toDfPartitioned` (there aren't yet, so maybe it wasn't needed!).

jiayuasu:
Close in favor of https://github.com/apache/sedona/pull/1780

PR #1767, opened 2025-01-22
[SEDONA-660] Add GeoArrow export from Spark DataFrame

### LINKED ISSUE #1756: Add GeoArrow IO to Sedona/Python
I'm still getting to know how the input/output works for Spark/Sedona, but I noticed that there's a `_collect_as_arrow()` method on data frames (I think this is exposed as `.toArrow()` in pyspark 4.0.0) and I'm wondering if there's an opportunity to implement something like `Adapter.to_geoarrow()` to provide compatibility with tools that import/export it (e.g., geopandas, geoarrow-rs).

```python
import os
import geopandas
import pyspark
from sedona.spark import SedonaContext
config = SedonaContext.builder().getOrCreate()
sedona = SedonaContext.create(config)

gs = geopandas.GeoSeries.from_wkt(["POINT (1 2)"])
gdf = geopandas.GeoDataFrame({"geometry": gs})

sedona.createDataFrame(gdf)._collect_as_arrow()
#> [pyarrow.RecordBatch
#>  geometry: binary
#>  ----
#>  geometry: [1200000001000000000000000000F03F0000000000000040]]
```

I am assuming that the binary here is the same binary that is being serialized/deserialized in https://github.com/apache/sedona/tree/52b6ae8e71601cdf36a6176198839bc3daf5547c/python/src .

A simple case might be converting that serialization to WKB and exporting `geoarrow.wkb`, which has the widest compatibility. With some information about which geometry types are present, it would be possible to generate "native" geoarrow (i.e., all coordinates together with separate buffers for part/ring offsets).

paleolimbot:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-660. The PR name follows the format `[SEDONA-660] my subject`.

Closes #1756.

Added `Adapter.toArrow()` to collect an RDD or DataFrame to an Arrow Table where Geometry columns are represented as `geoarrow.wkb` extension types

## How was this patch tested?

Tests will beadded to `python/tests/utils/test_geoarrow.py`.

- A new API is added to the Python bindings and will be documented


python/sedona/utils/geoarrow.py
```diff
@@ -0,0 +1,127 @@
...
+from sedona.sql.types import GeometryType
+
+
+def dataframe_to_arrow(df):
+    spark_schema = df.schema
+    table = dataframe_to_arrow_raw(df, spark_schema)
+
+    col_is_geometry = [
+        isinstance(f.dataType, GeometryType) for f in spark_schema.fields
+    ]
```
james-willis, line 32: This doesnt seem to handle geometries in complex types (arrays, structs)

james-willis:
Why not do all this in scala?

python/sedona/utils/geoarrow.py
paleolimbot: The tools for dealing with this on the pyarrow side are not great and might be quite a lot of work. Is this an important use case or is there a way to structure a `select()` (or push something into Scala) that I'm missing?

jiayuasu:
Maybe relevant: can this solve the toPandas() issue of PySpark DataFrame (https://spark.apache.org/docs/latest/api/python/user_guide/sql/arrow_pandas.html#enabling-for-conversion-to-from-pandas)?

In Sedona, when you call `toPandas()`, it gives you a GeoPandas DF but this will throw an exception if `spark.sql.execution.arrow.pyspark.enabled` set to `true` because Sedona Python does not work with GeoArrow.

CC @Kontinuation

python/sedona/utils/geoarrow.py
james-willis: I think it is important. I use geometries in complex types frequently. At least arrays of geometries is something that sedona itself outputs.

paleolimbot: Got it! I may defer to a follow-up since that's a bit of a learning curve for me. I think I can figure out how to walk the PySpark schema to get the nested paths corresponding to the Geometry nodes, and I think I can figure out how to do the node search/replace from the pyarrow side. For conversion, we can either:

- Go back to the version here where we transform the result on the Python side in C
- Figure out how to replace the `FieldSerializer`/`ArrowSerializer` on the Scala side to do what we need it to do there (basically: write WKB and add field metadata)
- Figure out how to issue a `select()` call in pyspark that does the transformation using the pyspark dataframe API

jiayuasu:
@paleolimbot do you have a benchmark number to show the performance gain? Will this benefit our integration with [lonboard](https://developmentseed.org/lonboard/latest/)?

jiayuasu:
@paleolimbot 

This is awesome.

Let me know if this works. Then I will merge this PR.

> I'll double check, but this should work out of the box (viz() accepts anything that implements __arrow_c_stream__, like a pyarrow.Table).

jiayuasu (changes requested):
can you also add a short paragraph at the bottom of this page? https://sedona.apache.org/latest/tutorial/sql/#convert-between-dataframe-and-spatialrdd

jiayuasu (approved):

PR #1769, opened 2025-01-23
[SEDONA-702] Create IndexedGridPartitioner, Support preserveUncontain…

james-willis:
…edGeometries in *GridPartitioners

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-702. The PR name follows the format `[SEDONA-XXX] my subject`.

* Adding an IndexedGridPartitioner class which should perform better than the FlatGridPartitioner in most cases for `placeObject` calls
* adding a `preserveUncontainedGeometries` flag into the GridPartitioner classes to allow callers to configure if geometries that are not contained within any partition should be placed in the overflow partition. 

## How was this patch tested?
Unit Tests added

While part of the public Java API, these partitioner classes do not have documentation strings. I have followed this norm by not adding doc strings to the classes.

### DIFF
--- a/spark/common/src/main/java/org/apache/sedona/core/spatialPartitioning/IndexedGridPartitioner.java
+++ b/spark/common/src/main/java/org/apache/sedona/core/spatialPartitioning/IndexedGridPartitioner.java
@@ -1,0 +1,88 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.sedona.core.spatialPartitioning;
+
+import java.util.*;
+import org.apache.sedona.core.enums.GridType;
+import org.locationtech.jts.geom.Envelope;
+import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.index.strtree.STRtree;
+import scala.Tuple2;
+
+public class IndexedGridPartitioner extends FlatGridPartitioner {
+  private final STRtree index;
+
+  public IndexedGridPartitioner(
+      GridType gridType, List<Envelope> grids, Boolean preserveUncontainedGeometries) {
+    super(gridType, grids, preserveUncontainedGeometries);
+    this.index = new STRtree();
+    for (int i = 0; i < grids.size(); i++) {
+      final Envelope grid = grids.get(i);
+      index.insert(grid, i);
+    }
+    index.build();
+  }
+
+  public IndexedGridPartitioner(GridType gridType, List<Envelope> grids) {
+    this(gridType, grids, false);
+  }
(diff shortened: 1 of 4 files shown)


spark/common/src/main/java/org/apache/sedona/core/spatialPartitioning/IndexedGridPartitioner.java
```diff
@@ -0,0 +1,88 @@
...
+      // borrowed from EqualPartitioning.placeObject
+      final int overflowContainerID = grids.size();
+      final Envelope envelope = spatialObject.getEnvelopeInternal();
+
+      Set<Tuple2<Integer, Geometry>> result = new HashSet();
+      boolean containFlag = false;
+      for (Object queryResult : results) {
+        Integer i = (Integer) queryResult;
+        final Envelope grid = grids.get(i);
+        if (grid.covers(envelope)) {
```
jiayuasu, line 71: What is the purpose of doing containFlag?

spark/common/src/main/java/org/apache/sedona/core/spatialPartitioning/IndexedGridPartitioner.java
```diff
@@ -0,0 +1,88 @@
...
+package org.apache.sedona.core.spatialPartitioning;
+
+import java.util.*;
+import org.apache.sedona.core.enums.GridType;
+import org.locationtech.jts.geom.Envelope;
+import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.index.strtree.STRtree;
+import scala.Tuple2;
+
+public class IndexedGridPartitioner extends FlatGridPartitioner {
```
jiayuasu, line 28: Can you add some explanation here to explain the usage of this Partitioner? It is used to index grid cells that are not  (1) from a equal partitioning (2) from space-partitioning tree.

james-willis: This should serve the same use cases as the FlatGridPartitioner, but provide better performance for `placeObject ` calls.

This serves use cases where you have specific grid cells you want to use for partitioning.

james-willis: added docstrings to both classes I changed here.

I'll note that equalPartitioning is not a Partitioner and there are no uses of the `public EqualPartitioning(Envelope boundary, int partitions)` constructor. EqualPartitioning is only use in the FlatGridPartitioner class for placing objects.

spark/common/src/main/java/org/apache/sedona/core/spatialPartitioning/IndexedGridPartitioner.java
```diff
@@ -0,0 +1,96 @@
...
+
+      Set<Tuple2<Integer, Geometry>> result = new HashSet();
+      boolean containFlag = false;
+      for (Object queryResult : results) {
+        Integer i = (Integer) queryResult;
+        final Envelope grid = grids.get(i);
+        if (grid.covers(envelope)) {
+          result.add(new Tuple2(i, spatialObject));
+          containFlag = true;
+        } else if (grid.intersects(envelope) || envelope.covers(grid)) {
```
jiayuasu, line 82: intersects is a superset of covers. Why check twice?

james-willis: not sure tbh; i took this logic from EqualParitioning, I assumed there was some edge case or odd behavior I wasnt privy to. You wrote this logic in commit `f90cfe82` 😂

jiayuasu: @james-willis OK. I think it was done by the old me 10 years ago. Let's remove it.

james-willis: done

jiayuasu (approved):

PR #1774, opened 2025-01-25
[SEDONA-703] Add utils for converting between RDD[Row] and SpatialRdd

james-willis:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-703. The PR name follows the format `[SEDONA-XXX] my subject`.

Added utils for converting between RDD[Row] and SpatialRDD. These are helpful for usecases that back physical nodes or functions that take in and return RDDs.

## How was this patch tested?
Unit Tests

Docstrings are available.

### DIFF
--- a/spark/common/src/main/scala/org/apache/sedona/sql/utils/Adapter.scala
+++ b/spark/common/src/main/scala/org/apache/sedona/sql/utils/Adapter.scala
@@ -76,7 +76,7 @@ object Adapter {
     val fieldList = dataFrame.schema.toList.map(f => f.name)
     val geomColId = fieldList.indexOf(geometryFieldName)
     assert(geomColId >= 0)
-    toRdd(dataFrame, geomColId)
+    toStringEncodedRdd(dataFrame.rdd, geomColId)
   }
 
   /**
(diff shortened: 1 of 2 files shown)


james-willis:
@Kontinuation @jiayuasu is this logic duplicated anywhere else? Maybe somewhere in the Join logic?

jiayuasu:
@james-willis This is where Sedona join does the conversion: 

DF to SpatialRDD:
https://github.com/apache/sedona/blob/3151f7d0c643daf915426435504ba5b88c7d975d/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/TraitJoinQueryBase.scala#L46

RDD to DF:
https://github.com/apache/sedona/blob/3151f7d0c643daf915426435504ba5b88c7d975d/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/TraitJoinQueryExec.scala#L125

Please consider reusing the logic if possible,

james-willis:
> Please consider reusing the logic if possible

Thanks Jia.

I was thinking that these functions I created would be useful for Physical Nodes but the linked code reminds me that InternalRows/UnsafeRows are used in these Physical Nodes, whereas the functions in this PR are for Rows. While you can convert between Rows and InternalRows, its an unnecessary conversion. It is probably best to keep distinct implementations for these two interfaces.

The fact that the implementations here are only useful for RDD[Row] but not RDD[InternalRow] makes me wonder if these are useful enough to introduce into the Sedona API.

PR #1778, opened 2025-01-29
[SEDONA-705] Add unique partitioner wrapper to enable partitioned writes with Sedona

paleolimbot:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-705. The PR name follows the format `[SEDONA-705] my subject`.

After SEDONA-695 (https://github.com/apache/sedona/pull/1751), we will have the ability to do partitioned writes! To be useful in most contexts, we also need the ability to create those partitions assigning a unique partition to each feature (i.e., don't introduce duplicates).

- Added a set of `spatialPartitioningWithoutDuplicates()` functions to match `spatialPartitioning()`
- Added a `GenericUniqueSpatialPartitioner` that wraps an existing partitioner producing a (deterministic) single result on `placeObject()`.
- Wired up the functions to Python

## How was this patch tested?

Tests were added in Java and Python.

- New API, will add once the approach is validated in principle!

### DIFF
--- a/docs/tutorial/sql.md
+++ b/docs/tutorial/sql.md
@@ -1651,7 +1651,9 @@ You can use `StructuredAdapter` and the `spatialRDD.spatialPartitioningWithoutDu
 === "Scala"
 
 	```scala
-	spatialRDD.spatialPartitioning(GridType.KDBTREE)
+	spatialRDD.spatialParitioningWithoutDuplicates(GridType.KDBTREE)
+	// Specify the desired number of partitions as 10, though the actual number may vary
+	// spatialRDD.spatialParitioningWithoutDuplicates(GridType.KDBTREE, 10)
 	var spatialDf = StructuredAdapter.toSpatialPartitionedDf(spatialRDD, sedona)
 	```
 
--- a/docs/tutorial/sql.md
+++ b/docs/tutorial/sql.md
@@ -1638,6 +1638,37 @@ Use SedonaSQL DataFrame-RDD Adapter to convert a DataFrame to an SpatialRDD. Ple
 	spatialDf = StructuredAdapter.toDf(spatialRDD, sedona)
 	```
 
+### SpatialRDD to DataFrame with spatial partitioning
+
+By default, `StructuredAdapter.toDf()` does not preserve spatial partitions because doing so
+may introduce duplicate features for most types of spatial data. These duplicates
+are introduced on purpose to ensure correctness when performing a spatial join;
+however, when using Sedona to prepare a dataset for distribution this is not typically
+desired.
+
+=== "Scala"
+
+	```scala
+	spatialRDD.spatialPartitioning(GridType.KDBTREE)
+	var spatialDf = StructuredAdapter.toSpatialPartitionedDf(spatialRDD, sedona)
+	```
+
+=== "Java"
+
+	```java
+	spatialRDD.spatialPartitioning(GridType.KDBTREE)
+	Dataset<Row> spatialDf = StructuredAdapter.toSpatialPartitionedDf(spatialRDD, sedona)
+	```
+
+=== "Python"
+
+	```python
+	from sedona.utils.structured_adapter import StructuredAdapter
+
+	spatialRDD.spatialPartitioning(GridType.KDBTREE)
+	spatialDf = StructuredAdapter.toSpatialPartitionedDf(spatialRDD, sedona)
+	```
+
 ### SpatialPairRDD to DataFrame
 
 PairRDD is the result of a spatial join query or distance join query. SedonaSQL DataFrame-RDD Adapter can convert the result to a DataFrame. But you need to provide the schema of the left and right RDDs.
--- a/docs/tutorial/sql.md
+++ b/docs/tutorial/sql.md
@@ -1662,7 +1662,7 @@ You can use `StructuredAdapter` and the `spatialRDD.spatialPartitioningWithoutDu
 	```java
 	spatialRDD.spatialParitioningWithoutDuplicates(GridType.KDBTREE)
 	// Specify the desired number of partitions as 10, though the actual number may vary
-	// spatialRDD.spatialParitioningWithoutDuplicates(GridType.KDBTREE, 10)	
+	// spatialRDD.spatialParitioningWithoutDuplicates(GridType.KDBTREE, 10)
 	Dataset<Row> spatialDf = StructuredAdapter.toSpatialPartitionedDf(spatialRDD, sedona)
 	```
 
--- a/docs/tutorial/sql.md
+++ b/docs/tutorial/sql.md
@@ -1667,7 +1667,9 @@ You can use `StructuredAdapter` and the `spatialRDD.spatialPartitioningWithoutDu
 	```python
 	from sedona.utils.structured_adapter import StructuredAdapter
 
-	spatialRDD.spatialPartitioning(GridType.KDBTREE)
+	spatialRDD.spatialPartitioningWithoutDuplicates(GridType.KDBTREE)
+	# Specify the desired number of partitions as 10, though the actual number may vary
+	# spatialRDD.spatialParitioningWithoutDuplicates(GridType.KDBTREE, 10)
 	spatialDf = StructuredAdapter.toSpatialPartitionedDf(spatialRDD, sedona)
 	```
 
--- a/docs/tutorial/sql.md
+++ b/docs/tutorial/sql.md
@@ -1646,6 +1646,8 @@ are introduced on purpose to ensure correctness when performing a spatial join;
 however, when using Sedona to prepare a dataset for distribution this is not typically
 desired.
 
+You can use `StructuredAdapter` and the `spatialRDD.spatialPartitioningWithoutDuplicates` function to obtain a Sedona DataFrame that is spatially partitioned without duplicates. This is especially useful for generating balanced GeoParquet files while preserving spatial proximity within files, which is crucial for optimizing filter pushdown performance in GeoParquet files.
+
 === "Scala"
 
 	```scala
--- a/docs/tutorial/sql.md
+++ b/docs/tutorial/sql.md
@@ -1660,7 +1660,9 @@ You can use `StructuredAdapter` and the `spatialRDD.spatialPartitioningWithoutDu
 === "Java"
 
 	```java
-	spatialRDD.spatialPartitioning(GridType.KDBTREE)
+	spatialRDD.spatialParitioningWithoutDuplicates(GridType.KDBTREE)
+	// Specify the desired number of partitions as 10, though the actual number may vary
+	// spatialRDD.spatialParitioningWithoutDuplicates(GridType.KDBTREE, 10)	
 	Dataset<Row> spatialDf = StructuredAdapter.toSpatialPartitionedDf(spatialRDD, sedona)
 	```
 
--- a/python/tests/test_base.py
+++ b/python/tests/test_base.py
@@ -26,6 +26,7 @@ from sedona.spark import *
 from sedona.utils.decorators import classproperty
 
 SPARK_REMOTE = os.getenv("SPARK_REMOTE")
+EXTRA_JARS = os.getenv("SEDONA_PYTHON_EXTRA_JARS")
 
 from shapely import wkt
 from shapely.geometry.base import BaseGeometry
@@ -36,6 +37,10 @@ class TestBase:
     @classproperty
     def spark(self):
         if not hasattr(self, "__spark"):
+            # This lets a caller override the value of SPARK_HOME to just use whatever
+            # is provided by pyspark. Otherwise, export SPARK_HOME="" has no effect.
+            if "SPARK_HOME" in os.environ and not os.environ["SPARK_HOME"]:
+                del os.environ["SPARK_HOME"]
 
             builder = SedonaContext.builder()
             if SPARK_REMOTE:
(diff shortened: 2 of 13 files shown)


spark/common/src/main/java/org/apache/sedona/core/spatialPartitioning/GenericUniquePartitioner.java
```diff
@@ -0,0 +1,67 @@
...
+import javax.annotation.Nullable;
+import org.apache.sedona.core.joinJudgement.DedupParams;
+import org.locationtech.jts.geom.Geometry;
+import scala.Tuple2;
+
+public class GenericUniquePartitioner extends SpatialPartitioner {
+  private SpatialPartitioner parent;
+
+  public GenericUniquePartitioner(SpatialPartitioner parent) {
+    super(parent.getGridType(), parent.getGrids());
```
jiayuasu, line 32: I recommend using partitionTree instead of grids. You can assume we will only support QuadTree and KDBTree partitioning for non-duplicate partitioner. All leaf nodes in a partition tree are actual grids.

Checking 1 geometry against the partitionTree is way faster than iterating over a list of grids.

jiayuasu, line 32: I recommend using partitionTree instead of grids. You can assume we will only support QuadTree and KDBTree partitioning for non-duplicate partitioner. All leaf nodes in a partition tree are actual grids.

Checking 1 geometry against the partitionTree is way faster than iterating over a list of grids.

python/tests/test_base.py
paleolimbot, line 1: I'm happy to move this to another PR...it's basically what it takes to get VSCode's Python testing integration to enable run/debug/discover.

spark/common/src/main/java/org/apache/sedona/core/spatialPartitioning/SpatialPartitioner.java
```diff
@@ -35,6 +35,11 @@ public abstract class SpatialPartitioner extends Partitioner implements Serializ
   protected final GridType gridType;
   protected final List<Envelope> grids;
 
+  protected SpatialPartitioner() {
+    gridType = null;
+    grids = null;
+  }
```
paleolimbot, line 41: A cleaner way to do this would probably be to remove the `gridType` and `grids` field and make `getGridType()` and `getGrids()` abstract? I could also remove this constructror and just pass the `GenericUniquePartitioner`'s parent grids/grid type through here.

spark/common/src/main/java/org/apache/sedona/core/spatialPartitioning/GenericUniquePartitioner.java
```diff
@@ -0,0 +1,77 @@
...
+        minGeometry = value._2();
+      }
+    }
+
+    HashSet<Tuple2<Integer, Geometry>> out = new HashSet<Tuple2<Integer, Geometry>>();
+    if (minGeometry != null) {
+      out.add(new Tuple2<Integer, Geometry>(minParitionId, minGeometry));
+    }
+
+    return out.iterator();
```
paleolimbot, line 64: I could also just take `it.Next()` here (i.e., always return the first one) and modify the `placeObject` implementations to not return an iterator off of a `HashSet` (i.e., write our own implementation of `Iterator`, which might be better anyway).

The motivation here is to ensure that the output is deterministic (i.e., if you set grids and ask Sedona to partition, you'll get the same result if you run your pipeline today or tomorrow).

spark/common/src/test/resources/.gitignore
```diff
@@ -1,2 +1,3 @@
 *.DS_Store
 real-*
+wkb/testSaveAs*
```
paleolimbot, line 3: Again, I can move this to another PR (running the Python tests creates quite a lot of files here!)

spark/common/src/main/java/org/apache/sedona/core/spatialRDD/SpatialRDD.java
```diff
@@ -278,7 +304,7 @@ public void spatialPartitioning(SpatialPartitioner partitioner) {
 
   /** @deprecated Use spatialPartitioning(SpatialPartitioner partitioner) */
   public boolean spatialPartitioning(final List<Envelope> otherGrids) throws Exception {
-    this.partitioner = new FlatGridPartitioner(otherGrids);
+    this.partitioner = new IndexedGridPartitioner(otherGrids);
```
paleolimbot, line 307: I think this probably is a better default (log complexity vs. linear complexity for placement). It's important to support this one for the partitioning use case (enables composability between the "generate my grids" and "partition my data" steps).

spark/common/src/main/java/org/apache/sedona/core/spatialPartitioning/IndexedGridPartitioner.java
```diff
@@ -48,7 +48,7 @@ public IndexedGridPartitioner(
   }
 
   public IndexedGridPartitioner(GridType gridType, List<Envelope> grids) {
-    this(gridType, grids, false);
+    this(gridType, grids, true);
```
paleolimbot, line 51: This aligns the default between the `IndexedGridPartitioner` and the `FlatGridPartitioner`

spark/common/src/main/java/org/apache/sedona/core/spatialRDD/SpatialRDD.java
james-willis: if we think this is important should we un-deprecate it? I have a similar pattern in isochrones, but I create the IndexedGridParitioner myself because this API is deprecated.

james-willis: Is it right to include the overflow partition here? perhaps this is why the API is deprecated, so you can configure the spatial partitioner on its own.

spark/common/src/main/java/org/apache/sedona/core/spatialRDD/SpatialRDD.java
```diff
@@ -159,6 +159,32 @@ public boolean spatialPartitioning(GridType gridType) throws Exception {
...
+    this.spatialPartitionedRDD = partition(partitioner);
+  }
+
+  public void spatialPartitioningWithoutDuplicates(SpatialPartitioner partitioner) {
+    partitioner = new GenericUniquePartitioner(partitioner);
+    this.spatialPartitionedRDD = partition(partitioner);
+  }
+
+  /** @deprecated Use spatialPartitioningWithoutDuplicates(SpatialPartitioner partitioner) */
+  public boolean spatialPartitioningWithoutDuplicates(final List<Envelope> otherGrids)
```
james-willis, line 181: why write a new deprecated API?

spark/common/src/main/java/org/apache/sedona/core/spatialRDD/SpatialRDD.java
```diff
@@ -159,6 +159,32 @@ public boolean spatialPartitioning(GridType gridType) throws Exception {
     return true;
   }
 
+  public boolean spatialParitioningWithoutDuplicates(GridType gridType) throws Exception {
```
james-willis, line 162: is withoutDuplicates better as its own set of methods or as a bool flag? No strong opinion but something to consider.

james-willis (changes requested):

spark/common/src/main/java/org/apache/sedona/core/spatialRDD/SpatialRDD.java
paleolimbot: That's a great point! I added it here because it's invoked by Python for the non-unique case, and my test uses it.

> Is it right to include the overflow partition here?

I think so...if it's not included will rows be removed from the output? I don't think there's a performance issue with including it.

spark/common/src/main/java/org/apache/sedona/core/spatialRDD/SpatialRDD.java
paleolimbot: Your idea of un-deprecating this is a good one...I had to add this because it's called from Python if you pass a list of envelopes (and that's how I test this). I'll turn this into a docstring with some useful content on the appropriate uses of this.

spark/common/src/main/java/org/apache/sedona/core/spatialRDD/SpatialRDD.java
paleolimbot: I do tend to have a preference towards methods (but also happy to change this to the prevailing opinion!). I know most Java IDEs let you toggle inlays that make this mute, but my theory was that `spatialParitioningWithoutDuplicates(GridType.KDBTREE)` is slightly more informative to read than `spatialPartitioning(GridType.KDBTREE, false)`.

Kontinuation: The python API has `introduce_duplicates` added as an optional parameter. I prefer making python API consistent with the Java API by adding a new `spatialParitioningWithoutDuplicates` method.

spark/common/src/main/java/org/apache/sedona/core/spatialPartitioning/GenericUniquePartitioner.java
Kontinuation: I'm OK with your current approach. We'd better add comment to the code selecting the partitioning result with the minimum partition id to signify that this is for producing consistent result.

spark/common/src/main/java/org/apache/sedona/core/spatialRDD/SpatialRDD.java
james-willis: > if it's not included will rows be removed from the output? 

Correct. best to leave it in

spark/common/src/main/java/org/apache/sedona/core/spatialRDD/SpatialRDD.java
paleolimbot: Done!

docs/tutorial/sql.md
```diff
@@ -1638,6 +1638,37 @@ Use SedonaSQL DataFrame-RDD Adapter to convert a DataFrame to an SpatialRDD. Ple
...
+By default, `StructuredAdapter.toDf()` does not preserve spatial partitions because doing so
+may introduce duplicate features for most types of spatial data. These duplicates
+are introduced on purpose to ensure correctness when performing a spatial join;
+however, when using Sedona to prepare a dataset for distribution this is not typically
+desired.
+
+=== "Scala"
+
+	```scala
+	spatialRDD.spatialPartitioning(GridType.KDBTREE)
```
jiayuasu, line 1652:
```suggestion
	spatialRDD.spatialParitioningWithoutDuplicates(GridType.KDBTREE)
	// Specify the desired number of partitions as 10, though the actual number may vary
	// spatialRDD.spatialParitioningWithoutDuplicates(GridType.KDBTREE, 10)
```

docs/tutorial/sql.md
```diff
@@ -1638,6 +1638,37 @@ Use SedonaSQL DataFrame-RDD Adapter to convert a DataFrame to an SpatialRDD. Ple
...
+	var spatialDf = StructuredAdapter.toSpatialPartitionedDf(spatialRDD, sedona)
+	```
+
+=== "Java"
+
+	```java
+	spatialRDD.spatialPartitioning(GridType.KDBTREE)
```
jiayuasu, line 1659:
```suggestion
	spatialRDD.spatialParitioningWithoutDuplicates(GridType.KDBTREE)
	// Specify the desired number of partitions as 10, though the actual number may vary
	// spatialRDD.spatialParitioningWithoutDuplicates(GridType.KDBTREE, 10)	
```

docs/tutorial/sql.md
```diff
@@ -1638,6 +1638,37 @@ Use SedonaSQL DataFrame-RDD Adapter to convert a DataFrame to an SpatialRDD. Ple
...
+	Dataset<Row> spatialDf = StructuredAdapter.toSpatialPartitionedDf(spatialRDD, sedona)
+	```
+
+=== "Python"
+
+	```python
+	from sedona.utils.structured_adapter import StructuredAdapter
+
+	spatialRDD.spatialPartitioning(GridType.KDBTREE)
```
jiayuasu, line 1668:
```suggestion
	spatialRDD.spatialPartitioningWithoutDuplicates(GridType.KDBTREE)
	# Specify the desired number of partitions as 10, though the actual number may vary
	# spatialRDD.spatialParitioningWithoutDuplicates(GridType.KDBTREE, 10)
```

docs/tutorial/sql.md
```diff
@@ -1638,6 +1638,37 @@ Use SedonaSQL DataFrame-RDD Adapter to convert a DataFrame to an SpatialRDD. Ple
 	spatialDf = StructuredAdapter.toDf(spatialRDD, sedona)
 	```
 
+### SpatialRDD to DataFrame with spatial partitioning
+
+By default, `StructuredAdapter.toDf()` does not preserve spatial partitions because doing so
+may introduce duplicate features for most types of spatial data. These duplicates
+are introduced on purpose to ensure correctness when performing a spatial join;
+however, when using Sedona to prepare a dataset for distribution this is not typically
+desired.
```
jiayuasu, line 1647:
```suggestion
desired.

You can use `StructuredAdapter` and the `spatialRDD.spatialPartitioningWithoutDuplicates` function to obtain a Sedona DataFrame that is spatially partitioned without duplicates. This is especially useful for generating balanced GeoParquet files while preserving spatial proximity within files, which is crucial for optimizing filter pushdown performance in GeoParquet files.
```

jiayuasu (changes requested):

PR #1788, opened 2025-02-04
[SEDONA-707] Refactor rasterization process and add allTouched parameter for rasterization functions

### LINKED ISSUE #1715: Add all touched parameter to RS_ZonalStats
I want to use ZonalStats (https://sedona.apache.org/latest/api/sql/Raster-operators/#rs_zonalstats) for my use case of integrating polygons of buildings with information from a tif file.

In the past I used the rasterstats library to get the info in python, but i have a lot of buildings and wanted to use the power of Sedona. When I look at rasterstats it has the property all_touched which determines if the pixel should be included in the rasterstats calculation see; https://pythonhosted.org/rasterstats/manual.html#zonal-statistics. 

## Actual behavior
The current RS_ZonalStats function only returns the value of pixels that fall within my polygon. This causes a lot of issues when the houses / buildings are quite small (it will return a NaN because non of the pixels are fully encapsulated by the polygon. 

```
Example of the usage
results = sedona.sql('''
select 
id
, geometry
, RS_ZonalStats(r.raster, a.geometry, 1, 'max', false, true) as max_elevation 
from buildings a, tifmap_with_elevation r
where 
RS_intersects(geometry, r.raster) 
'''
```

It would be great to have an additional parameter which would indicate if RS_ZonalStats should only summarize the statistics of pixels within the geometry or all pixel that are being touched by the geometry 

I thought that adding a buffer with the size of my pixels would solve my issue, but i think it is then still possible to grab pixels which should not be part of the calculation. 

## Settings

Sedona version = 1.7.0

Apache Spark version = 3.5.0

Apache Flink version = ?

API type = Python / SQL 

Scala version = 2.12

JRE version = ?

Python version = 3.11

Environment = Databricks

prantogg:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-707. The PR name follows the format `[SEDONA-XXX] my subject`.

Fixes #1715

Affects API of downstream RS_Functions, introduces `allTouched` mode to:
- `RS_ZonalStatsAll` & `RS_ZonalStats`
- `RS_Clip`
- `RS_AsRaster`
- `RS_SetValues`

### Algorithm Overview

The algorithm is inspired by GDAL’s `GDALdllImageFilledPolygon` but is implemented in Java with optimizations for duplicate handling, centroid-based inclusion, and allTouched behavior.

The rasterization process follows two main strategies:

1. Polygon Rasterization (Scanline Algorithm)
    - Used for Polygon and MultiPolygon geometries.
    - Implements scanline-based filling with centroid-based inclusion.
    - Supports `allTouched` mode for including all pixels touched by the polygon.
2. Line Rasterization (Modified Bresenham Algorithm)
    - Used for LineString and MultiLineString geometries.
    - Implements a modified Bresenham’s line-drawing algorithm with fractional steps.
    - Ensures sub-pixel accuracy using stepSize control.
    - Clips lines that extend beyond raster bounds.
3. Optimized support for GeometryCollections and MultiPolygons 
    - Decomposes into sub-geoms and computes a separate search gird for each.

## How was this patch tested?

- Benchmarked against previous rasterization process
    - Up to **31.4% faster** rasterization for `Polygons`
    - Up to **194% faster** for tests including all geometry types.


common/src/main/java/org/apache/sedona/common/raster/Rasterization.java
```diff
@@ -0,0 +1,598 @@
...
+
+  // Modified Bresenham with Fractional Steps
+  private static void drawLineBresenham(
+      RasterizationParams params,
+      double x0,
+      double y0,
+      double x1,
+      double y1,
+      double value,
+      double stepSize) {
```
Kontinuation, line 196: Why not use a traditional Bresenham's line algorithm using integer arithmetics? Using a `stepSize = 0.2` here introduces lots of repeated setSample operations.

common/src/main/java/org/apache/sedona/common/raster/Rasterization.java
```diff
@@ -0,0 +1,598 @@
...
+
+        boolean intersects =
+            geom.intersects(JTS.toGeometry(new Envelope(worldX, cellMaxX, worldY, cellMaxY)));
+
+        if (intersects) {
+          params.writableRaster.setSample(x, yIndex, 0, value);
+        }
+      }
+    }
+  }
```
Kontinuation, line 155: I think we can apply some arithmetic to figure out which pixels to render, instead of iterating all the pixels and test each of them by calling `intersects`.

Kontinuation (commented):
I like the overall idea of implementing rasterization on our own instead of relying on GeoTools' `VectorToRasterProcess` as well as Java AWT's Graphics2D. I've left some comments regarding the implementation of various rasterization functions.

common/src/main/java/org/apache/sedona/common/raster/Rasterization.java
jiayuasu: Are you sure there is a better way to do this? I think the only alternative is scanline but I don't think that is worth trying. Each polygon could have holes too, which is more complex.

prantogg: Here, the loop does not iterate over all points in each scan line. The loop iterates over a search grid in raster space of the minimum spanning bbox of the point geom. This search grid is calculated by `rasterizeGeomExtent()` and is passed to `rasterizePoint` as `geomExtent`. 

There are 3 possibilities for the point to be located in raster space -
- If the point is fully contained in a pixel - `geomExtent` covers thats pixel only - and `intersects` gets called once.
- If the point is on an edge between 2 pixels - `geomExtent` covers 2 pixels touching the edge - and `intersects` gets called twice.
- If point is at the vertex of a pixel where it touches 4 total pixels - `geomExtent` will cover those 4 pixels - in this case `intersects` gets called 4 times.

Kontinuation: OK. If the searched space is the bbox of the point then this should be fine.

common/src/main/java/org/apache/sedona/common/raster/Rasterization.java
```diff
@@ -0,0 +1,598 @@
...
+        // Contain y range within geomExtent; Use centroid y line as scan line
+        yStart = Math.max(0.5, Math.abs(yStart) - 0.5);
+        yEnd = Math.min((params.writableRaster.getHeight() - 0.5), Math.abs(yEnd) + 0.5);
+
+        double slope = (worldP2.y - worldP1.y) / (worldP2.x - worldP1.x);
+        double p1X = (worldP1.x - params.upperLeftX) / params.scaleX;
+        double p1Y = (params.upperLeftY - worldP1.y) / params.scaleY;
+
+        for (double y = yStart; y >= yEnd; y--) {
+          double xIntercept = p1X + ((p1Y - y) / slope);
```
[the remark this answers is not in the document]
prantogg: For each segment, yStart and yEnd are calculated separately (shown below). This ensures that xIntercepts are calculated only for scanLines that may intersect the respective segment.

```
// Calculate scan line limits to iterate between for each segment
// Using BigDecimal to avoid floating point errors
double yStart =
    Math.ceil(
        (valueOf(params.upperLeftY)
                .subtract(valueOf(worldP1.y))
                .divide(valueOf(params.scaleY), RoundingMode.CEILING))
            .doubleValue());

double yEnd =
    Math.floor(
        (valueOf(params.upperLeftY)
                .subtract(valueOf(worldP2.y))
                .divide(valueOf(params.scaleY), RoundingMode.FLOOR))
            .doubleValue());

// Contain y range within geomExtent; Use centroid y line as scan line
yEnd = Math.max(0.5, Math.abs(yEnd) + 0.5);
yStart = Math.min((params.writableRaster.getHeight() - 0.5), Math.abs(yStart) - 0.5);
```

common/src/main/java/org/apache/sedona/common/raster/Rasterization.java
```diff
@@ -0,0 +1,598 @@
...
+        }
+
+        double yStart = Math.floor((worldP1.y - params.upperLeftY) / params.scaleY);
+        double yEnd = Math.round((worldP2.y - params.upperLeftY) / params.scaleY);
+
+        // Contain y range within geomExtent; Use centroid y line as scan line
+        yStart = Math.max(0.5, Math.abs(yStart) - 0.5);
+        yEnd = Math.min((params.writableRaster.getHeight() - 0.5), Math.abs(yEnd) + 0.5);
+
+        double slope = (worldP2.y - worldP1.y) / (worldP2.x - worldP1.x);
```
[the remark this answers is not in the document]
prantogg: Fixed

common/src/main/java/org/apache/sedona/common/raster/Rasterization.java
```diff
@@ -0,0 +1,598 @@
...
+        // Ensure worldP1.y ≤ worldP2.y
+        if (worldP1.y > worldP2.y) {
+          Coordinate temp = worldP1;
+          worldP1 = worldP2;
+          worldP2 = temp;
+        }
+
+        if (worldP1.y == worldP2.y) {
+          continue; // Skip horizontal edges
+        }
```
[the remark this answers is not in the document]
prantogg: Pixels between `worldP1.x` and `worldP2.x` get burned when working on non-horizontal lines that meet `worldP1` and `worldP2`.
The xIntercepts near `worldP1.x` and `worldP2.x` are recorded when iterating over those non-horizontal lines that meet `worldP1` and `worldP2`.

common/src/main/java/org/apache/sedona/common/raster/Rasterization.java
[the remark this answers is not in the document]
Kontinuation: Nit: Don't import the `valueOf` static method. use `BigDecimal.valueOf` instead of `valueOf` for better clarity.

common/src/main/java/org/apache/sedona/common/raster/Rasterization.java
Kontinuation: I found that GDAL uses a [more complex approach](https://github.com/OSGeo/gdal/blob/v3.10.2/alg/llrasterize.cpp#L651-L710) for allTouched mode: it figures out which step to take (right or up/down) to step into the next pixel. This could be more reliable than using a fractional step of 0.2.

I am OK with the current approach, but I think we can switch to GDAL's approach in the future.

PR #1808, opened 2025-02-12
[DOCS] Improve Makefile by Using requirements-docs.txt for Documentation Dependencies

BaseMax:
This update refactors the docsinstall target in the Makefile to use a predefined requirements-docs.txt file instead of dynamically creating it. This ensures better dependency management, simplifies maintenance, and aligns with best practices for handling Python package installations.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,29 +15,72 @@
 # specific language governing permissions and limitations
 # under the License.
 
-check :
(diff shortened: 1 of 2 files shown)


jbampton (approved):
Tested locally.  Ran the Makefile phony targets.  Seems to work. 

Thanks :+1:

jiayuasu (approved):

PR #1811, opened 2025-02-13
[WIP] Add Geography user-defined type

paleolimbot:

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-711. The PR name follows the format `[SEDONA-711] my subject`.

This PR adds a `GeographyUDT` to match the `GeometryUDT`. This matches how the types are defined in Iceberg and Parquet, and maps to the concepts of `"edges": "spherical"` in GeoParquet and GeoArrow.

I'm happy to discuss the scope of the initial implementation in this PR...I'd propose starting with a thin wrapper around a JTS geometry (After all: coordinates are coordinates! And many structural operations are identical in a geometry/geography context). Things like serializers and WKB/WKT IO are also nicely defined for a Geometry already and I don't think there's a strong case for reinventing those in Java land (but feel free to correct me!).

## How was this patch tested?

Work in progress!

Work in progress!


common/src/main/java/org/apache/sedona/common/geometryObjects/Geography.java
```diff
@@ -0,0 +1,37 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.sedona.common.geometryObjects;
+
+import org.locationtech.jts.geom.Geometry;
+
+public class Geography {
```
jiayuasu, line 23: I think `extends Geometry` might be better? CC @Kontinuation

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/UDT/GeographyUDT.scala
```diff
@@ -0,0 +1,61 @@
...
+ * under the License.
+ */
+package org.apache.spark.sql.sedona_sql.UDT
+
+import org.apache.sedona.common.geometrySerde.GeometrySerializer;
+import org.apache.spark.sql.catalyst.util.{ArrayData, GenericArrayData}
+import org.apache.spark.sql.types._
+import org.json4s.JsonDSL._
+import org.json4s.JsonAST.JValue
+import org.apache.sedona.common.geometryObjects.Geography;
```
jiayuasu, line 26: Honestly we probably don't even need to have a new Java Geography in sedona-common because the storage model of Geography is identical to Geometry (unless we want to annotate on the edge interpolation algorithm?). So I would say we just have a GeographyUDT and it uses JTS Geometry out of the box.

paleolimbot: Sorry for taking a while to circle back here!

I am worried that a strict subclass will lead to functions that accept a `Geometry` being able to also implicitly accept a `Geography`, which in general is a footgun for getting accidentally invalid results (`Geometry` should only ever be compared with `Geometry`; `Geography` should only ever be compared with `Geography`). A subclass would also tie the implementation of the coordinate storage to JTS, which would mean a copy if we ever use something like s2geometry to do actual computations.

Apologies if I'm missing something obvious here!

common/src/main/java/org/apache/sedona/common/geometryObjects/Geography.java
```diff
@@ -0,0 +1,37 @@
...
+  private Geometry geometry;
```
james-willis, line 24: is there any extra overhead to using a Geometry internally? like time or space cost in the constructor.

python/sedona/register/java_libs.py
```diff
@@ -26,6 +26,7 @@ class SedonaJvmLib(Enum):
     KNNQuery = "org.apache.sedona.core.spatialOperator.KNNQuery"
     RangeQuery = "org.apache.sedona.core.spatialOperator.RangeQuery"
     Envelope = "org.locationtech.jts.geom.Envelope"
+    Geography = "import org.apache.sedona.common.geometryObjects.Geography"
```
james-willis, line 29: should the word import be here?

common/src/main/java/org/apache/sedona/common/Functions.java
```diff
@@ -784,6 +785,10 @@ public static byte[] asEWKB(Geometry geometry) {
     return GeomUtils.getEWKB(geometry);
   }
 
+  public static byte[] geogAsEWKB(Geography geography) {
```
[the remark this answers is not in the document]
paleolimbot: If I add that it fails to compile with:

```
[ERROR] /workspaces/sedona/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala:509: error: ambiguous reference to overloaded definition,
[ERROR] both method asEWKB in class Functions of type (x$1: org.apache.sedona.common.geometryObjects.Geography)Array[Byte]
[ERROR] and  method asEWKB in class Functions of type (x$1: org.locationtech.jts.geom.Geometry)Array[Byte]
[ERROR] match expected type org.apache.spark.sql.sedona_sql.expressions.InferrableFunction
[ERROR]     extends InferredExpression(Functions.asEWKB _) {
[ERROR]                                          ^
[ERROR] one error found
```

(That said, I can't any call to `ST_AsWKB()` on a geography or a geometry to work after this change, so maybe there is something else wrong!)

python/sedona/register/java_libs.py
paleolimbot: Good catch!

common/src/main/java/org/apache/sedona/common/Functions.java
```diff
@@ -29,6 +29,7 @@
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.sedona.common.geometryObjects.Circle;
+import org.apache.sedona.common.geometryObjects.Geography;
```
jiayuasu, line 32: Let's put the Geography related functions / constructors ... to separate files, instead of mixing with the Geometry function.

Can you also put Geography functions into individual files? The old `Functions.java` / `Constructors.java` are too large so it is probably better to put them into individual files such as "GeogFromWKB"?

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/InferredExpression.scala
```diff
@@ -23,7 +23,7 @@ import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.{Expression, ImplicitCastInputTypes}
 import org.apache.spark.sql.catalyst.expressions.codegen.CodegenFallback
 import org.apache.spark.sql.catalyst.util.ArrayData
-import org.apache.spark.sql.sedona_sql.UDT.GeometryUDT
+import org.apache.spark.sql.sedona_sql.UDT.{GeometryUDT, GeographyUDT}
```
jiayuasu, line 26: For functions that support both Geometry and Geography, I believe we should overload the functions with Geometry and Geography function. This is not currently supported by `InferredExpression`? @Kontinuation @zhangfengcdt

common/src/main/java/org/apache/sedona/common/Functions.java
paleolimbot: Totally! I would like to solve the runtime overload problem first...I can try to look harder at how one registers a UDF with more than one signature (maybe it is not possible in Spark?), since that is the part that is currently causing an issue.

common/src/main/java/org/apache/sedona/common/geometryObjects/Geography.java
paleolimbot: There might be (e.g., we use some other library to do an overlay operation like intersection and have to convert from that library's representation back to JTS). Keeping the field private I think should at least provide a route to changing the internal implementation if there are performance issues in the future (but also happy to hear suggestions otherwise!)

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/InferredExpression.scala
paleolimbot: (That is how PostGIS does it, although we can also punt on that and use a different prefix if we have to do that to get started)

common/src/main/java/org/apache/sedona/common/Functions.java
Kontinuation: It is possible in Spark, the way to do this is quite flexible but awkward.

The implementation of `inputTypes` can inspect what is the types of the actual expressions passed into it, and return a suitable function signature. Examples are

* Summary stats function: https://github.com/apache/sedona/blob/a8da3cdb0953ca06d1eb5a64e799f2da75d0e712/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/raster/RasterFunctions.scala#L140-L150
* Inferred expression: https://github.com/apache/sedona/blob/sedona-1.7.0/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/InferredExpression.scala#L61-L79

We can also inspect the types of inputs and run different code depending on input types in eval function:

* Summary stats function: https://github.com/apache/sedona/blob/a8da3cdb0953ca06d1eb5a64e799f2da75d0e712/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/raster/RasterFunctions.scala#L105-L131

Inferred expression encapsulates the above function overloading mechanism of Spark and supports delegating the Spark expression to Java functions according to their arity. It is possible to extend it to support more complex function overloading rules.

common/src/main/java/org/apache/sedona/common/Functions.java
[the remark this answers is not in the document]
Kontinuation: There's a (quite complex syntax) for referring to a specific overload of functions in Scala. Even if we support argument type based overloading in Inferred expression, we still have to list all overloads we want to delegate to. For instance

```scala
InferredExpression(
  (g: Geography) => Functions.asEWKB(g),
  (g: Geometry) => Functions.asEWKB(g))
```

common/src/main/java/org/apache/sedona/common/geometryObjects/Geography.java
Kontinuation: I agree with Dewey that we'd better define Geography differently to avoid possible misuse. The internal representation of Geography may change as we integrate with libraries that supports spherical geometry, then the JTS Geometry representation of Geography will become optional. This design gives us the flexibility of opting out JTS Geometry when it is not needed.

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/InferredExpression.scala
zhangfengcdt: I think creating a base class for GeographyUDT and GeometryUDT can promote code reuse and simplify the implementation of common operations. By doing this, we can ensure a clear hierarchy and facilitate maintainability, especially when dealing with operations applicable to both geography and geometry types. 

Also, for the Geography and Geometry, we can define a common base GeoObject as well, and have a Geometry implementation in Sedona that extends this base class but use JTS Geometry internally to implement algorithms.

zhangfengcdt: Also our RasterUDT can also potentially rebased on this as well.

common/src/main/java/org/apache/sedona/common/Functions.java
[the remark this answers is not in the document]
zhangfengcdt: Having Geometry and Geography from different packages / repositories looks a bit strange to me, though I understand the tradeoff here. Is it a good opportunity to formally introduce the sedona geo types for this in this effort?

zhangfengcdt (commented):
just some thoughts on the types

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/InferredExpression.scala
paleolimbot: I quite like the idea of having base classes that we own (e.g., it would enable things like `SpatialRDD<T extends SedonaAbstractGeo>` without making Geography/Raster a subclass of Geometry). I wonder if that level of refactoring we might want to defer slightly (and keep our initial efforts focused on the minimum requirement for iceberg?)

Kontinuation:
> (@Kontinuation if you end up getting here before I have time to implement that, feel free to let me know and start with the right thing!).

I have taken over this work and will continue working on https://github.com/apache/sedona/pull/1828

PR #1860, opened 2025-03-17
[SEDONA-722] Fix precision loss problems caused by casting world coordinates from double to float

Kontinuation:

- Yes, and the PR name follows the format `[SEDONA-XXX] my subject`.

The world coordinates were silently casted from double to float when getting world coordinates from grid coordinates. This affects the precision of `RS_RasterToWorldCoord*`, `RS_PixelAsPolygon`, etc.

## How was this patch tested?

Updated the expected results of tests to more precise coordinates.

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/utils/RasterUtils.java
+++ b/common/src/main/java/org/apache/sedona/common/utils/RasterUtils.java
@@ -45,7 +45,6 @@ import org.geotools.coverage.Category;
 import org.geotools.coverage.CoverageFactoryFinder;
 import org.geotools.coverage.GridSampleDimension;
 import org.geotools.coverage.TypeMap;
-import org.geotools.coverage.grid.GridCoordinates2D;
 import org.geotools.coverage.grid.GridCoverage2D;
 import org.geotools.coverage.grid.GridCoverageFactory;
 import org.geotools.coverage.grid.GridEnvelope2D;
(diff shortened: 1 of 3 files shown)


common/src/main/java/org/apache/sedona/common/utils/RasterUtils.java
```diff
@@ -467,10 +466,11 @@ public static AffineTransform2D translateAffineTransform(
 
   public static Point2D getWorldCornerCoordinates(GridCoverage2D raster, int colX, int rowY)
       throws TransformException {
+    Point2D.Double gridCoord = new DirectPosition2D(colX - 1, rowY - 1);
     return raster
         .getGridGeometry()
         .getGridToCRS2D(PixelOrientation.UPPER_LEFT)
-        .transform(new GridCoordinates2D(colX - 1, rowY - 1), null);
+        .transform(gridCoord, null);
```
Kontinuation, line 473: `GridCoordinates2D` is not a subclass of `Point2D.Double`, so the result of transform will be `Point2D.Float`: https://github.com/openjdk/jdk/blob/jdk-17%2B35/src/java.desktop/share/classes/java/awt/geom/AffineTransform.java#L2899-L2906. We explicitly use `Point2D.Double` here to avoid loss of precision.

jiayuasu (approved):

PR #1871, opened 2025-03-20
[SEDONA-724] Fix RS_ZonalStats and RS_ZonalStatsAll edge case bug

furqaankhan:

- Yes, and the PR name follows the format `[SEDONA-724] my subject`.

When the ROI barely intersects with the input raster's extent and the `allTouched` argument is false then this causes the pixel data to be empty. That breaks the zonalMode method which assumes there will be pixel data.

Add better error handling and a shortcut for RS_ZonalStatsAll to skip unnecessary stats compute (saves ~6% time).

## How was this patch tested?

- add unit tests

### DIFF
--- a/common/src/test/java/org/apache/sedona/common/raster/RasterBandAccessorsTest.java
+++ b/common/src/test/java/org/apache/sedona/common/raster/RasterBandAccessorsTest.java
@@ -28,7 +28,6 @@ import org.junit.Test;
 import org.locationtech.jts.geom.Geometry;
 import org.locationtech.jts.io.ParseException;
 import org.opengis.referencing.FactoryException;
-import org.opengis.referencing.operation.TransformException;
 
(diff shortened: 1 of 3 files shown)


common/src/test/java/org/apache/sedona/common/raster/RasterBandAccessorsTest.java
```diff
@@ -84,6 +84,41 @@ public void testBandNoDataValueIllegalBand() throws FactoryException, IOExceptio
     assertEquals("Provided band index 2 is not present in the raster", exception.getMessage());
   }
 
+  @Test
```
jiayuasu, line 87: Please add a test to show how this works in a real Sedona RS function. I am not sure if it can successfully convert Double.NaN to Spark null.

jiayuasu (changes requested):

jiayuasu (approved):

PR #1885, opened 2025-03-27
[GH-1886] Create EnvelopeAccumulator class

james-willis:

https://github.com/apache/sedona/issues/1886

Add an EnvelopeAccumulator to allow developers to expose the spatial extent of the data in each stage/task to the user through the spark ui.

## How was this patch tested?
Manual - Unit Tests to come.

### DIFF
--- a/spark/common/src/main/java/org/apache/sedona/core/monitoring/EnvelopeAccumulator.java
+++ b/spark/common/src/main/java/org/apache/sedona/core/monitoring/EnvelopeAccumulator.java
@@ -1,0 +1,66 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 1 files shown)


jiayuasu:
Good idea!

james-willis:
> are there any performance implications here?

Yes but I dont have specific numbers. merging n envelopes (ie `expandToInclude`) takes O(n) time. 

> are we computing an aggregate bbox even for queries that would normally not require one?
The way it is used is kind of up to the implementor of the RDD code that backs the task/stage.

>  is this only computed when viewed?
It is always computed if the RDD operation uses one. it is persisted into the spark history file. 

An A/B test could be a good thing here for perf. Generally I think the observability benefits will outweigh the performance cost which I expect to be <1s wall time for almost any query.

Kontinuation (approved):

PR #1917, opened 2025-04-12
[SEDONA-720] Refactor GeoPandas Series implementation

zhangfengcdt:

- Yes, and the PR name follows the format `[SEDONA-XXX] my subject`.

This PR refactor the GeoSeries python wrapper so that the code can be simplified, e.g.,

```
    def area(self) -> "GeoSeries":
           return self._process_geometry_column("ST_Area", rename="area")

    def buffer(
        self,
        distance,
        resolution=16,
        cap_style="round",
        join_style="round",
        mitre_limit=5.0,
        single_sided=False,
        **kwargs,
    ):
        return self._process_geometry_column(
            "ST_Buffer", rename="buffer", distance=distance
        )

```

## How was this patch tested?
test_geoseries.py

### DIFF
--- a/python/sedona/geopandas/base.py
+++ b/python/sedona/geopandas/base.py
@@ -49,7 +49,6 @@ class GeoFrame(metaclass=ABCMeta):
     def __getitem__(self, key: Any) -> Any:
         raise NotImplementedError("This method is not implemented yet.")
 
-    @abstractmethod
     def _reduce_for_geostat_function(
         self,
         sfun: Callable[["GeoSeries"], Column],
--- a/python/sedona/geopandas/geodataframe.py
+++ b/python/sedona/geopandas/geodataframe.py
@@ -16,21 +16,19 @@
 #  under the License.
 from __future__ import annotations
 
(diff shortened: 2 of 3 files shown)


furqaankhan (approved):
LGTM! 

One QQ: What if the user wants to use any other geometry column than the first one in the schema? Because I see that we take the very first geometry column as input to the functions. Should we add a field that lets users pick the geometry column? This may not be the scope for this PR.

zhangfengcdt:
> LGTM!
> 
> One QQ: What if the user wants to use any other geometry column than the first one in the schema? Because I see that we take the very first geometry column as input to the functions. Should we add a field that lets users pick the geometry column? This may not be the scope for this PR.

The GeoSeries (and Pandas Series) should only have one "column" that in geo should be geometry type. Though for GeoDataFrame (and Pandas DataFrame) they can have multiple geometry columns. We can use set_geometry on the dataframe, but it won't be necessary (needed) for GeoSeries.

furqaankhan:
Understood, thank you!

jiayuasu (approved):

Issue #1910, opened 2025-04-04
Hidden requirement for geopandas in apache-sedona version 1.7.1


ivanthewebber:
Newest python package version gives following error:
```shell
    import sedona.spark as ss
  File "/opt/hostedtoolcache/Python/3.12.9/x64/lib/python3.12/site-packages/sedona/spark/__init__.py", line 36, in <module>
    from sedona.maps.SedonaKepler import SedonaKepler
  File "/opt/hostedtoolcache/Python/3.12.9/x64/lib/python3.12/site-packages/sedona/maps/SedonaKepler.py", line 18, in <module>
    from sedona.maps.SedonaMapUtils import SedonaMapUtils
  File "/opt/hostedtoolcache/Python/3.12.9/x64/lib/python3.12/site-packages/sedona/maps/SedonaMapUtils.py", line 21, in <module>
    from sedona.utils.geoarrow import dataframe_to_arrow
  File "/opt/hostedtoolcache/Python/3.12.9/x64/lib/python3.12/site-packages/sedona/utils/geoarrow.py", line 30, in <module>
    import geopandas as gpd
ModuleNotFoundError: No module named 'geopandas'
```
I think this is a re-emergence of #1404

jiayuasu:
@ivanthewebber thank you for reporting this!

PR #1929, opened 2025-04-29
[GH-1910] Import geopandas only when type checking

oliverbeagley:

Move geopandas import to a type checking block to prevent a hidden dependency on it.

## How was this patch tested?

`import sedona.spark` now works

### DIFF
--- a/python/sedona/utils/geoarrow.py
+++ b/python/sedona/utils/geoarrow.py
@@ -15,7 +15,7 @@
 #  specific language governing permissions and limitations
 #  under the License.
 import itertools
-from typing import List, Callable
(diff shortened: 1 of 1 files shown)


oliverbeagley:
As an aside [this file](https://github.com/apache/sedona/blob/master/python/sedona/maps/SedonaMapUtils.py) makes an import from `packaging` which is also not listed as a dependency

jiayuasu:
@oliverbeagley great work. I will merge this one first. Then you can create another PR to fix the `packaging` dependency issue

jiayuasu (approved):

PR #1932, opened 2025-04-30
[GH-1931] Move packaging module import to geopandas try-except block

### LINKED ISSUE #1931: Hidden requirement for `packaging` in apache-sedona version 1.7.1
Discovered in #1929.

When moving the geopandas import to a type checking block there is still an import failure for `sedona.spark` due to and import of the packaging module: https://github.com/apache/sedona/blob/01c67e9e4b3407e64173c7af9d450243112dc9d4/python/sedona/maps/SedonaMapUtils.py#L22

oliverbeagley:

Fixes #1931 

I have moved the import of `packaging.version.parse` to the try-except block that `geopandas` is imported in, as `packaging` is already a dependency of `geopandas`

## How was this patch tested?

Local test by successfully running

```py
from sedona.spark import SedonaContext
import pyspark.sql.functions as sf


SedonaContext.create(spark)
spark.range(10).withColumn("geom", sf.expr("ST_POINT(id, id)")).show()
```

### DIFF
--- a/python/sedona/maps/SedonaMapUtils.py
+++ b/python/sedona/maps/SedonaMapUtils.py
@@ -19,7 +19,6 @@ import json
 
 from sedona.sql.types import GeometryType
 from sedona.utils.geoarrow import dataframe_to_arrow
-from packaging.version import parse
 
 
 class SedonaMapUtils:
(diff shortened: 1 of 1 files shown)


oliverbeagley:
I am in two minds about whether `packaging` should be added to the requirements.

I haven't as it stands, and other areas of the code make similar decisions e.g. `geoarrow`, `pyarrow`, `pyproj` are not listed as optional dependencies anywhere but are all used in the code in `python/sedona/utils/geoarrow.py`.

jiayuasu:
@oliverbeagley I don't think we need to add `packaging` to optional dependencies because it is already a required dependency of GeoPandas. And currently it is only used for checking geopandas version. If geopandas is there, then packaging should be there too

jiayuasu (approved):

PR #1942, opened 2025-05-09
[GH-1943] Upgrade to Java 11

jesspav:

- Yes, GH-1943, and the PR name follows the format `[GH-XXX] my subject`.

- Minimum changes for upgrading to Java 11:
  - Configure pom.xml files for Java 11
  - Upgrade spotless to the version recommended for Java 11
  - Update workflows to use Java 11

## How was this patch tested?

- Built and run tests
- Running workflows as part of this PR

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Compile JavaDoc
         run: mvn -q clean install -DskipTests && mkdir -p docs/api/javadoc/spark && cp -r spark/common/target/apidocs/* docs/api/javadoc/spark/
       - name: Compile ScalaDoc
-        run: mvn scala:doc && mkdir -p docs/api/scaladoc/spark && cp -r spark/common/target/site/scaladocs/* docs/api/scaladoc/spark
+        run: mvn scala:doc --pl spark && mkdir -p docs/api/scaladoc/spark && cp -r spark/common/target/site/scaladocs/* docs/api/scaladoc/spark
       - uses: actions/setup-python@v5
         with:
           python-version: 3.x
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Compile JavaDoc
         run: mvn -q clean install -DskipTests && mkdir -p docs/api/javadoc/spark && cp -r spark/common/target/apidocs/* docs/api/javadoc/spark/
       - name: Compile ScalaDoc
-        run: mvn scala:doc --pl spark && mkdir -p docs/api/scaladoc/spark && cp -r spark/common/target/site/scaladocs/* docs/api/scaladoc/spark
+        run: mvn scala:doc -pl !common,!snowflake,!flink && mkdir -p docs/api/scaladoc/spark && cp -r spark/common/target/site/scaladocs/* docs/api/scaladoc/spark
       - uses: actions/setup-python@v5
         with:
           python-version: 3.x
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -44,7 +44,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'zulu'
-          java-version: '8'
+          java-version: '11'
       - name: Compile JavaDoc
         run: mvn -q clean install -DskipTests && mkdir -p docs/api/javadoc/spark && cp -r spark/common/target/apidocs/* docs/api/javadoc/spark/
       - name: Compile ScalaDoc
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -46,6 +46,7 @@ on:
 env:
   MAVEN_OPTS: -Dmaven.wagon.httpconnectionManager.ttlSeconds=60
   DO_NOT_TRACK: true
+  SPARK_LOCAL_IP: 127.0.0.1
 
 permissions:
   contents: read
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -69,16 +69,8 @@ jobs:
             scala: 2.12.15
             jdk: '11'
             skipTests: ''
-          - spark: 3.5.0
-            scala: 2.12.15
-            jdk: '11'
-            skipTests: ''
           - spark: 3.4.0
             scala: 2.13.8
-            jdk: '11'
-            skipTests: ''
-          - spark: 3.4.0
-            scala: 2.12.15
             jdk: '11'
             skipTests: ''
           - spark: 3.4.0
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -63,11 +63,11 @@ jobs:
         include:
           - spark: 3.5.0
             scala: 2.13.8
-            jdk: '8'
+            jdk: '11'
             skipTests: ''
           - spark: 3.5.0
             scala: 2.12.15
-            jdk: '8'
+            jdk: '11'
             skipTests: ''
           - spark: 3.5.0
             scala: 2.12.15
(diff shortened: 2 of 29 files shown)


.github/workflows/java.yml
```diff
@@ -63,35 +63,35 @@ jobs:
...
             jdk: '11'
             skipTests: ''
           - spark: 3.4.0
             scala: 2.13.8
-            jdk: '8'
+            jdk: '11'
             skipTests: ''
           - spark: 3.4.0
             scala: 2.12.15
-            jdk: '8'
```
jiayuasu, line 82: There are Java 11 tests for Spark 3.4 and 3.5 already. Please remove those duplicates.

jiayuasu (changes requested):

.github/workflows/java.yml
jesspav: good catch. thanks!

zhangfengcdt:
> @Kontinuation @zhangfengcdt I don't understand why this is happening (complain about the port address). Is there anything specific in this PR caused this problem?

This might be an issue that JDK 11’s implementation of InetAddress.getLocalHost() may return an address tied to a network interface that isn’t up or even an IPv6 address. The Spark 3.3 changed to use spark.driver.host (defaulting to the IP from getLocalHost()) as its bind address. 

I'd suggest to set it explicitly, so the hostname maps do not impact it on the build machine. 

For example,

SparkSession.builder()
    .config("spark.driver.bindAddress", "127.0.0.1")
    .config("spark.driver.host",        "127.0.0.1")
    .getOrCreate();

or 

export SPARK_LOCAL_IP=127.0.0.1


Note that a later Spark version might fix this already.

common/pom.xml
```diff
@@ -122,6 +122,11 @@
             <groupId>edu.ucar</groupId>
             <artifactId>cdm-core</artifactId>
         </dependency>
+            <dependency>
```
jiayuasu, line 125: Does this have to be here? I think this should be in the `pom.xml` of `spark` module, not the `common` module.

jiayuasu (changes requested):

spark/common/src/test/java/org/apache/sedona/sql/adapterTestJava.java
```diff
@@ -55,7 +55,12 @@ public class adapterTestJava implements Serializable {
   public static void onceExecutedBeforeAll() {
     sparkSession =
         SedonaContext.create(
-            SedonaContext.builder().master("local[*]").appName("adapterTestJava").getOrCreate());
+            SedonaContext.builder()
+                .master("local[*]")
+                .appName("adapterTestJava")
+                .config("spark.driver.bindAddress", "127.0.0.1")
```
zhangfengcdt, line 61: Did you test exposing "export SPARK_LOCAL_IP=127.0.0.1" as a solution to this? I think it might be a cleaner solution if it works.

common/pom.xml
jesspav: I'll try moving it to pom.xml. 
Also - sorry for the alignment, I thought spotless would fix that.

spark/common/src/test/java/org/apache/sedona/sql/adapterTestJava.java
jesspav: Yeah, that would be a much smaller/nicer change.  I'll try it out.

common/pom.xml
jiayuasu: The reason why I want to remove it from `common`'s pom.xml is that the `common` module is a pure Java module that is shared among all Sedona modules including snowflake, spark, flink. Sedona Snowflake, and Sedona Flink don't even use Scala. So I don't think we should have it bundled in the `common` module.

spark/common/src/test/java/org/apache/sedona/sql/adapterTestJava.java
jesspav: Tests fail locally again with spark fail to connect errors when I remove the configs and run `mvn clean install  -Dspark=3.5 -Dscala=2.12 -Dgeotools -DSPARK_LOCAL_IP=127.0.0.1` .

common/pom.xml
jesspav: oh! That is an excellent reason, I didn't know that. :)
In that case, how about I just change the yaml script that calls scala:doc to build docs only for the `spark` module?
Can use:
```
-pl,--projects <arg>            Comma-delimited list of specified reactor projects to build instead
                                              of all projects. A project can be specified by [groupId]:artifactId or by its relative path
```

.github/workflows/docs.yml
```diff
@@ -44,11 +44,11 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'zulu'
-          java-version: '8'
+          java-version: '11'
       - name: Compile JavaDoc
         run: mvn -q clean install -DskipTests && mkdir -p docs/api/javadoc/spark && cp -r spark/common/target/apidocs/* docs/api/javadoc/spark/
       - name: Compile ScalaDoc
-        run: mvn scala:doc && mkdir -p docs/api/scaladoc/spark && cp -r spark/common/target/site/scaladocs/* docs/api/scaladoc/spark
+        run: mvn scala:doc --pl spark && mkdir -p docs/api/scaladoc/spark && cp -r spark/common/target/site/scaladocs/* docs/api/scaladoc/spark
```
zhangfengcdt, line 51: The --pl spark could change the doc generation behavior that only the submodule doc is generated. So the document i believe is only at: spark/target/site/scaladocs/*

jiayuasu: @zhangfengcdt I think this is expected? See my conversation above with Jess about the scala doc.

zhangfengcdt: Yeah, undersood. The issue is that we need to change the cp command to use the correct source path after adding --pl spark.

jesspav: Yup, will fix. one sec...

jesspav: Looks like it isn't quite the same thing.

After running `mvn scala:doc -pl spark` there are only these two directories:
```
% find ./ -name "scaladocs"
./spark/spark-3.4/target/site/scaladocs
./spark/spark-3.5/target/site/scaladocs
```

vs `mvn scala:doc` has a third directory
```
% find ./ -name "scaladocs"  
...
./spark/common/target/site/scaladocs
./spark/spark-3.4/target/site/scaladocs
./spark/spark-3.5/target/site/scaladocs
```

Which contains docs that are not in the other two and is the one that we are currently uploading
```
% ls spark/common/target/site/scaladocs/org/apache/sedona/sql 
datasources			RasterRegistrator$.html		UDF				utils
index.html			SedonaSqlExtensions.html	UDT
% ls spark/spark-3.4/target/site/scaladocs/org/apache/sedona/sql 
datasources	index.html	parser
% ls spark/spark-3.5/target/site/scaladocs/org/apache/sedona/sql
datasources	index.html	parser
```

The common doc returns if I exclude the modules that are failing: `mvn scala:doc -X -pl !common,!snowflake,!flink`

Note that the mission related docs are all in the spark module:

```
% find ./ -name "*RasterRegistrator*"
./spark/common/target/classes/org/apache/sedona/sql/RasterRegistrator$.class
./spark/common/target/classes/org/apache/sedona/sql/RasterRegistrator.class
./spark/common/target/classes/org/apache/sedona/sql/RasterRegistrator$$typecreator1$1.class
./spark/common/target/site/scaladocs/org/apache/sedona/sql/RasterRegistrator$.html
./spark/common/src/main/scala/org/apache/sedona/sql/RasterRegistrator.scala
```

jiayuasu (approved):
Nice work. LGTM!

zhangfengcdt (approved):

jiayuasu:
@jesspav My bad. It looks like the Snowflake test failed: https://github.com/wherobots/sedona-snowflake-tester

Note that: snowflake tests only runs under the `[sedona-snowflake-tester](https://github.com/wherobots/sedona-snowflake-tester)` github repo because it needs a snowflake account. Please create a new PR against both `apache/sedona` and [sedona-snowflake-tester](https://github.com/wherobots/sedona-snowflake-tester)

jiayuasu:
There are some example PRs here: https://github.com/wherobots/sedona-snowflake-tester/pulls?q=sort%3Aupdated-desc+is%3Apr+is%3Aclosed

jesspav:
will do

PR #1975, opened 2025-06-04
[SEDONA-738] Add moran i autocorrelation.

Imbruced:

- Yes

SEDONA-738

MoranI index

## How was this patch tested?
Unit tests

Yes

### DIFF
--- a/docs/api/stats/sql.md
+++ b/docs/api/stats/sql.md
@@ -136,7 +136,7 @@ names in parentheses are python variable names
 
 In both cases the output is the input DataFrame with the weights column added to each row.
 
-# MoranI
+## Moran I
 
 Moran I is the spatial autocorrelation algorithm, which is using spatial
 location and non-spatial attribute. When the value is close to the 1 it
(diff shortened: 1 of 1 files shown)


Imbruced:
I need to add docs

Imbruced:
and update the description of the MR

james-willis:
Thanks for this Pawel. 

I think we need an ST Function for this. Currently it is asymmetric with the other stats functions in this regard.

I also believe that we should deprecate these function call interfaces for geostats functions and tell users to use the ST functions instead. If we go this direction I would advise against making python and public scala functions for morans I (other than ST functions for the df interfaces of course)

jiayuasu:
@james-willis Is your SQL function framework able to support Pawel's Moran'I case as well? Maybe you can point him to the right place for this?

james-willis:
Sure. We should be able to wrap the function in this PR with an implementation of a [PhysicalFunction](https://github.com/apache/sedona/blob/master/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/PhysicalFunction.scala#L98).

[Here](https://github.com/apache/sedona/blob/master/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/GeoStatsFunctions.scala) are the existing ones.

[Here](https://github.com/apache/sedona/pull/1736) is the PR that added SQL support to geostats

jiayuasu:
@Imbruced I will merge this for now but the doc needs significant improvement. I believe we need some comprehensive end-to-end examples of each Stat function

james-willis:
Filed a ticket for ST function: https://github.com/apache/sedona/issues/2115

PR #1992, opened 2025-06-20
[GH-1996] Create object of S2Geography, and implement Point/Polyline/PolygonGeography with its encoder/decoder

ZhuochengShang:

The first step of porting https://github.com/paleolimbot/s2geography, ported S2Geography base class and 3 sub types.

## How was this patch tested?

Added unit tests.


jiayuasu:
@ZhuochengShang great work. Please create a GitHub issue and link the PR to that issue using [GH-XXXX]

jiayuasu:
Great work @ZhuochengShang . It is great to have a working PR that shows the progress. But when it is ready, we'd like to break it into a sequence of small PRs. This will make the review process easier.

jiayuasu:
Please also update the title of this PR to `[GH-XXXX]` that links to the GitHub issue.

common/src/main/java/org/apache/sedona/common/S2Geography/S2Geography.java
```diff
@@ -0,0 +1,220 @@
...
+  /**
+   * Returns an S2Region that represents the object. The caller retains ownership of the S2Region
+   * but the data pointed to by the object requires that the underlying Geography outlives the
+   * returned object.
+   *
+   * @return S2Region
+   */
+  public S2Region region() {
+    return null;
+  }
```
Kontinuation: We can define `S2Geography` as an abstract class and don't implement these methods. Subclasses are required to implement these methods.

common/src/main/java/org/apache/sedona/common/S2Geography/PointShape.java
```diff
@@ -0,0 +1,117 @@
...
+import com.google.common.geometry.S2Edge;
+import com.google.common.geometry.S2Point;
+import com.google.common.geometry.S2Shape;
+import java.util.List;
+
+/**
+ * Java equivalent of the C++ S2PointVectorShape: each point is a degenerate edge (start==end), one
+ * chain per point.
+ */
+/** A degenerate S2Shape that represents exactly one point. */
```
Kontinuation: `PointShape` does not represent exactly one point. It represents a list of points. Please refer to [`S2PointVectorShape`](https://github.com/google/s2geometry/blob/v0.12.0/src/s2/s2point_vector_shape.h#L32-L37) for details.

common/src/main/java/org/apache/sedona/common/S2Geography/PointShape.java
```diff
@@ -0,0 +1,117 @@
...
+    result.set(point.get(chainId), point.get(chainId));
+  }
+
+  @Override
+  public S2Point getChainVertex(int chainId, int offset) {
+    if (chainId != 0 || offset != 0) {
+      throw new IndexOutOfBoundsException("PointShape chainId and offset must both be 0");
+    }
+    return point.get(chainId);
+  }
```
Kontinuation: numChains should return the number of points, all other getChain* methods should be adjusted accordingly.

common/src/main/java/org/apache/sedona/common/S2Geography/S2Geography.java
```diff
@@ -0,0 +1,220 @@
...
+    tag.setKind(kind);
+    if (numShapes() == 0) {
+      tag.setFlags((byte) (tag.getFlags() | EncodeTag.FLAG_EMPTY));
+    }
+    // compute covering if requested
+    List<S2CellId> cover = new ArrayList<>();
+    if (options.isIncludeCovering()) {
+      getCellUnionBound(cover);
+    }
+    tag.setCoveringSize((byte) cover.size());
```
Kontinuation, line 173: We should skip encoding the covering when `cover.size() > 256`, since it does not fit in the 8-bit `coveringSize`.

common/src/main/java/org/apache/sedona/common/S2Geography/S2Geography.java
```diff
@@ -0,0 +1,220 @@
...
+    // 3) encode by GeographyKind
+    switch (kind) {
+      case POINT:
+        ((PointGeography) this).encodeTagged(out, options);
+        break;
+
+        // TODO: handle POLYLINE, POLYGON, etc.
+      default:
+        throw new IllegalArgumentException("encodeTagged not implemented for kind=" + kind);
+    }
```
Kontinuation: We can directly call `this.encode(out)` here. Subclasses will implement the `encode` method to encode itself without tag. Explicit dispatching is not needed.

common/src/main/java/org/apache/sedona/common/S2Geography/PointGeography.java
```diff
@@ -0,0 +1,168 @@
...
+    tag.encode(out);
+    for (var c2 : cover) out.writeLong(c2.id());
+
+    // payload
+    S2Point.Shape shp = S2Point.Shape.fromList(points);
+    if (opts.getCodingHint() == EncodeOptions.CodingHint.FAST) {
+      CountingPointVectorCoder.INSTANCE.encode(shp, out);
+    } else {
+      PointShapeCoders.COMPACT.encode(shp, out);
+    }
```
Kontinuation: The empty points or other cases can fallback to parent class's implementation (the C++ implementation uses `Geography::EncodeTagged(encoder, options)` for doing this). We just need to implement `encode()` here. The parent class's implementation will call our `encode` in `encodeTagged`.

common/src/main/java/org/apache/sedona/common/S2Geography/S2Geography.java
```diff
@@ -0,0 +1,220 @@
...
+        return PointGeography.decodeTagged(in, tag);
+      case POLYLINE:
+        return PolylineGeography.decodeTagged(in, tag);
+      case POLYGON:
+        return PolygonGeography.decodeTagged(in, tag);
+      default:
+        throw new IllegalArgumentException("Unsupported kind " + kind);
+    }
+  }
+}
```
Kontinuation, line 214: We should declare `encode` and `decode` functions for encoding/decoding geography objects without tags.

common/src/main/java/org/apache/sedona/common/S2Geography/PointShapeCoders.java
```diff
@@ -0,0 +1,48 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.sedona.common.S2Geography;
+
+import com.google.common.geometry.S2Coder;
+import com.google.common.geometry.S2Point;
+import java.lang.reflect.Field;
+
+public class PointShapeCoders {
```
Kontinuation: The latest `org.datasyslab:s2-geometry-library` has `S2Point.Shape.FAST_CODER` and `COMPACT_CODER` as public. This utility class for extracting non-public static fields is no longer needed.

`CountingPointVectorCoder` can be removed as well.

common/src/main/java/org/apache/sedona/common/S2Geography/S2Geography.java
```diff
@@ -0,0 +1,141 @@
...
+   * (e.g., which geography type or flags). Encode this geography into a stream as: 1) a 5-byte
+   * EncodeTag header (see EncodeTag encode / decode) 2) coveringSize × 8-byte cell-ids 3) the raw
+   * shape payload (point/polyline/polygon) via the built-in coder
+   *
+   * @param opts CodingHint.FAST / CodingHint.COMPACT / Include or omit the cell‐union covering
+   *     prefix
+   */
+  public abstract void encode(OutputStream os, EncodeOptions opts) throws IOException;
+
+  // public abstract S2Geography decode(DataInputStream in) throws IOException;
```
Kontinuation: This still does not look correct to me. We should also implement `encodeTagged` here. `encodeTagged` constructs the `EncodeTag` object and write it into the output stream, then calls the abstract `encode` method.

`encode` method are implemented by subclasses, they don't need to deal with `EncodeTag` in their `encode` implementations.

`PointGeography` has an optimized representation when it encodes only one point in compact mode, so it overrides the default `encodeTagged` implementation provided by `S2Geography` parent class to to provide a more optimized implementation. Other S2Geography subclasses don't do this. They just implement `encode`, and let the parent class encode tags.

The `decodeTagged` implementations I saw in the previous round of review look good, we should not remove them.

Kontinuation: The overall structure will be identical to the original C++ implementation.

In S2Geography, we implement `encodeTagged` and `decodeTagged`, and defines `encode` abstract methods for subclasses to implement:

```java
public class S2Geography {
  public void encodeTagged(OutputStream os, EncodeOptions opts) throws IOException {
    DataOutputStream out = new DataOutputStream(os);

    // 1) Get covering if needed
    // ...

    // 2) Write tag header
    // ...

    // 3) Write the geography
    this.encode(out, opts);
    out.flush();
  }

  public static S2Geography decodeTagged(InputStream is) throws IOException {
    DataInputStream in = new DataInputStream(is);
    
    // 1) decode the tag
    EncodeTag tag = ...

    // 2) dispatch to subclass's decode method according to tag.kind
    switch (tag.kind) {
      case GeographyKind::POINT: return PointGeography::decode(in, tag);
      case ...:
    }
  }

  public abstract void encode(DataOutputStream out, EncodeOptions opts) throws IOException;
}
```

The C++ version defines `decode` as a non-static function, while the implementation I reviewed the last time defines it as a static method. Both approaches are fine.

PointGeography implements `encode` and `decode`, it also overrides `encodeTagged` to implement an optimized encoding by omitting the points when possible.

```java
public class PointGeography {
  @Override
  public void encodeTagged(OutputStream os, EncodeOptions opts) throws IOException {
    if (points.size() == 1 && opts.getCodingHint() == EncodeOptions.CodingHint.COMPACT) {
      // Optimized encoding which only uses covering to represent the point
      return;
    }

    // In other cases, fallback to the default encodeTagged implementation:
    super.encodeTagged(os, opts);
  }

  @Override
  public void encode(DataOutputStream out, EncodeOptions opts) throws IOException {
    // encode point list
    S2Point.Shape shp = S2Point.Shape.fromList(points);
    switch (opts.getCodingHint()) {
      case FAST: ...
      case COMPACT: ...
    }
  }

  public static PointGeography decode(DataInputStream in, EncodeTag tag) throws IOException {
     ...
  }
}
```

For any other geography types without an optimized encoded representation, only `encode` and `decode` needs to be implemented:

```java
public class PolygonGeography {
  @Override
  public void encode(DataOutputStream out, EncodeOptions opts) throws IOException {
    for (S2Polygon poly : polygons) {
      poly.encode(out);
    }
  }

  public static PolygonGeography decode(DataInputStream in, EncodeTag tag) throws IOException {
     ...
  }
}
```

common/src/main/java/org/apache/sedona/common/S2Geography/PolylineGeography.java
```diff
@@ -0,0 +1,155 @@
...
+    EncodeTag tag = new EncodeTag(opts);
+    tag.setKind(GeographyKind.POLYLINE);
+    tag.setCoveringSize((byte) cover.size());
+    tag.encode(out);
+    for (S2CellId c2 : cover) {
+      out.writeLong(c2.id());
+    }
+
+    // 3) **Critical**: write the number of polylines in little-endian
+    out.writeInt(polylines.size());
```
Kontinuation: `DataOutputStream` always write multi-byte numbers as big-endian.

common/src/main/java/org/apache/sedona/common/S2Geography/PointGeography.java
```diff
@@ -0,0 +1,215 @@
...
+        out.writeByte(0); // coveringSize
+        out.writeByte(2); // COMPACT encode type
+        out.writeLong(cid.id());
+        out.flush();
+        return;
+      }
+      super.encodeTagged(os, opts); // Not exactly encodable as a cell center
+    }
+    // In other cases, fallback to the default encodeTagged implementation:
+    super.encodeTagged(os, opts);
```
Kontinuation: `super.encodeTagged` may be called twice. We can remove the super.encodeTagged call on line 136.

common/src/main/java/org/apache/sedona/common/S2Geography/PointGeography.java
```diff
@@ -0,0 +1,215 @@
...
+    switch (opts.getCodingHint()) {
+      case FAST:
+        S2Point.Shape.FAST_CODER.encode(shp, out);
+        break;
+      case COMPACT:
+        S2Point.Shape.COMPACT_CODER.encode(shp, out);
+    }
+  }
+
+  public static PointGeography decode(Input in, EncodeTag tag) throws IOException {
```
Kontinuation, line 171: I suggest that we change the input parameter type of `decode` methods from `Input` to `UnsafeInput`, and also change `skipCovering` and `decodeCovering` methods of `EncodeTag` to take `UnsafeInput` as argument. This is because UnsafeInput may over-consume the wrapped input stream, we'd better only use one UnsafeInput object throughout the decoding of an object.

common/src/main/java/org/apache/sedona/common/S2Geography/PointGeography.java
```diff
@@ -0,0 +1,215 @@
...
+      // representation, we will have a very loose covering if the level is low.
+      // Level 23 has a cell size of ~1 meter
+      // (http://s2geometry.io/resources/s2cell_statistics)
+      if (cid.level() >= 23) {
+        out.writeByte(GeographyKind.CELL_CENTER.getKind());
+        out.writeByte(0); // POINT kind
+        out.writeByte(1); // flag
+        out.writeByte(0); // coveringSize
+        out.writeByte(2); // COMPACT encode type
+        out.writeLong(cid.id());
```
Kontinuation: This looks strange to me. The bytes written here does not conform to the format of `EncodeTag`. We should write an EncodeTag header with kind = CELL_CENTER, flag = 0, coveringSize = 1, reserved = 0.

You can construct a new EncodeTag object and set the fields to correct values, then call `tag.encode(out)`. This will ensure that we always write the encode tag header in consistent format.

common/src/main/java/org/apache/sedona/common/S2Geography/PointGeography.java
```diff
@@ -0,0 +1,215 @@
...
+    switch (tag.getEncodeType()) {
+      case 1:
+        points = S2Point.Shape.FAST_CODER.decode(bytes, cursor);
+        break;
+      case 2:
+        points = S2Point.Shape.COMPACT_CODER.decode(bytes, cursor);
+        break;
+      default:
+        throw new IllegalArgumentException("Unknown coding hint");
+    }
```
Kontinuation: The internal buffer of UnsafeInput may not contain the entire encoded point list. This will be problematic when we have a PointGeography with lots of points.

The S2 Coder interface of Java makes it hard to decode data using streams, we can write an integer indicating the total length of the encoded point before the actual payload in `encode`. We can read the length and read the entire payload into a byte array, then call the decode function of S2 Coder. This results in in-compatible encoding format with the C++ implementation, but we can do this for now until we need to exchange data with some native components.

common/src/main/java/org/apache/sedona/common/S2Geography/PointGeography.java
```diff
@@ -0,0 +1,215 @@
...
+
+    PrimitiveArrays.Cursor cursor = bytes.cursor();
+    List<S2Point> points;
+    switch (tag.getEncodeType()) {
+      case 1:
+        points = S2Point.Shape.FAST_CODER.decode(bytes, cursor);
+        break;
+      case 2:
+        points = S2Point.Shape.COMPACT_CODER.decode(bytes, cursor);
+        break;
```
Kontinuation: Note:

Actually we don't need to distinguish between FAST_CODER and COMPACT_CODER when decoding. The payload already has the encoding in the first byte, we can always use FAST_CODER and it will decode the payload correctly regardless of the coding type of the payload.

We can omit the `encodeType` field in `EncodeTag` to make `EncodeTag` consistent with the C++ implementation, this will help a lot when we decide to exchange encoded data with some native components in the future.

I'm not sure if we can do the same thing when decoding shape index, so we'd better leave it as is for now.

common/src/main/java/org/apache/sedona/common/S2Geography/EncodeTag.java
```diff
@@ -0,0 +1,150 @@
...
+  public EncodeTag() {}
+
+  public EncodeTag(EncodeOptions opts) {
+    this.encodeType = (byte) (opts.getCodingHint() == EncodeOptions.CodingHint.FAST ? 1 : 2);
+  }
+
+  // ——— Write the 4-byte tag header ——————————————————————————————————————
+
+  /** Write exactly 4 bytes: [kind|flags|coveringSize|reserved]. */
+  public void encode(Output out) throws IOException {
```
Kontinuation: Now our header is 5 bytes. I suggest we remove the reserved byte to keep it aligning to 4-bytes boundary.

Kontinuation (commented):
The overall structure looks good. There are some additional comments mostly for the decoding part.

I also suggest that we add more comprehensive tests for encoding and decoding of geography objects. For instance, `includeCovering` option has not been tested for now. Adding a utility test function for round-trip encoding/decoding any geography objects and validate their equality would help writing such tests a lot.

common/src/main/java/org/apache/sedona/common/S2Geography/PointGeography.java
ZhuochengShang: use flag byte to check encode type, avoiding mismatch 4 bytes of header

common/src/main/java/org/apache/sedona/common/S2Geography/EncodeTag.java
```diff
@@ -0,0 +1,162 @@
...
+public class EncodeTag {
+  /**
+   * Subclass of S2Geography whose decode() method will be invoked. Encoded using a single unsigned
+   * byte (represented as an int in Java, range 0–255).
+   */
+  private GeographyKind kind = GeographyKind.UNINITIALIZED;
+  /**
+   * Flags for encoding metadata. Currently, only {@code kFlagEmpty} is supported, which is set if
+   * and only if the geography contains zero shapes.
+   */
```
Kontinuation: This comment needs to be updated as we have 2 supported flags now.

common/src/main/java/org/apache/sedona/common/S2Geography/PointGeography.java
```diff
@@ -0,0 +1,241 @@
...
+      S2CellId cid = S2CellId.fromPoint(points.get(0));
+      // Only encode this for very high levels: because the covering *is* the
+      // representation, we will have a very loose covering if the level is low.
+      // Level 23 has a cell size of ~1 meter
+      // (http://s2geometry.io/resources/s2cell_statistics)
+      if (cid.level() >= 23) {
+        out.writeByte(GeographyKind.CELL_CENTER.getKind());
+        out.writeByte(0); // POINT kind
+        out.writeByte(1); // flag
+        out.writeByte(0); // coveringSize
```
Kontinuation:
```suggestion
        out.writeByte(0); // flag
        out.writeByte(1); // coveringSize
        out.writeByte(0); // reserved
```

common/pom.xml
```diff
@@ -177,6 +189,10 @@
                                     <pattern>org.codehaus</pattern>
                                     <shadedPattern>org.apache.sedona.shaded.codehaus</shadedPattern>
                                 </relocation>
+                                <relocation>
+                                    <pattern>com.google.common.geometry;</pattern>
```
Kontinuation:
```suggestion
                                    <pattern>com.google.common.geometry</pattern>
```

common/pom.xml
```diff
@@ -109,6 +114,12 @@
             <groupId>edu.ucar</groupId>
             <artifactId>cdm-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-collections4</artifactId>
+            <version>4.4</version>
+            <scope>compile</scope>
+        </dependency>
```
Kontinuation: I think we don't need to add this dependency. The newly added code does not depend on `commons-collections4`.

common/pom.xml
```diff
@@ -81,9 +81,14 @@
             <groupId>org.locationtech.spatial4j</groupId>
             <artifactId>spatial4j</artifactId>
         </dependency>
+<!--        <dependency>-->
+<!--            <groupId>com.google.geometry</groupId>-->
+<!--            <artifactId>s2-geometry</artifactId>-->
+<!--        </dependency>-->
```
Kontinuation: We can remove this commented dependency, but add some comment to clarify that `org.datasyslab:s2-geometry-library` is a fork of `com.google.geometry:s2-geometry-library`, and probably add a link to the issue [GH-1996](https://github.com/apache/sedona/issues/1996).

common/src/main/java/org/apache/sedona/common/S2Geography/PolylineGeography.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.logging.Logger;
+
+/** A Geography representing zero or more polylines using S2Polyline. */
+public class PolylineGeography extends S2Geography {
+  private static final Logger logger = Logger.getLogger(PolylineGeography.class.getName());
+
+  private final List<S2Polyline> polylines;
```
Kontinuation: We define `PolylineGeography` to contain a list of `S2Polyline`s, which is different from the [C++ version](https://github.com/paleolimbot/s2geography/blob/0.2.0/src/s2geography/geography.h#L179-L203). I think this is a reasonable change, as it makes distinguishing between POLYGON and MULTIPOLYGON much easier as long as we ensure that each `S2Polyline` only contains one polygon. WDYT @paleolimbot

common/src/main/java/org/apache/sedona/common/S2Geography/PointGeography.java
Kontinuation: Here is a better way of doing this:

```java
EncodeTag tag = new EncodeTag();
tag.setKind(GeographyKind.CELL_CENTER);
tag.setCompact(true);
tag.setCoveringSize((byte) 1);
tag.encode(out);
```

This ensure that the encoding of tag is always handled by `EncodeTag.encode`.

common/src/test/java/org/apache/sedona/common/S2Geography/PointGeographyTest.java
```diff
@@ -0,0 +1,182 @@
...
+    TestHelper.assertRoundTrip(geog, new EncodeOptions());
+  }
+
+  @Test
+  public void testEncodedSnappedPoint() throws IOException {
+    // 1) Build the original point and its snapped-to-cell-center version
+    S2Point pt = S2LatLng.fromDegrees(45, -64).toPoint();
+    S2CellId cellId = S2CellId.fromPoint(pt);
+    S2Point ptSnapped = cellId.toPoint();
+    System.out.println(ptSnapped.toString());
```
jiayuasu: Please remove `print` command.

common/src/main/java/org/apache/sedona/common/S2Geography/PolylineGeography.java
paleolimbot: I believe that Polyline in C++ is also a `std::vector<S2Polyline>`, but you're correct that the polygon is defined this way.

Paul (PostGIS) complained about this too, mostly because some multipolygons don't roundtrip through WKB/T/JTS. I've personally found it easier to keep the S2-centric structure since a lot of the time internally it doesn't matter. Either way is fine!

common/src/main/java/org/apache/sedona/common/S2Geography/PolygonGeography.java
```diff
@@ -32,42 +31,39 @@
 public class PolygonGeography extends S2Geography {
   private static final Logger logger = Logger.getLogger(PolygonGeography.class.getName());
 
-  private final List<S2Polygon> polygons;
+  public final S2Polygon polygon;
 
   public PolygonGeography() {
     super(GeographyKind.POLYGON);
-    this.polygons = new ArrayList<>();
+    this.polygon = null;
```
Kontinuation: We can initialize `this.polygon` with an empty `S2Polygon` object created using the default constructor. This will eliminate possible NPEs in `PolygonGeography.encode` and any other methods.

common/src/main/java/org/apache/sedona/common/S2Geography/PolygonGeography.java
```diff
@@ -32,42 +31,39 @@
...
+    return polygon.shape();
   }
 
   @Override
   public S2Region region() {
-    Collection<S2Region> regionCollection = new ArrayList<>(polygons);
+    S2RegionWrapper s2RegionWrapper = new S2RegionWrapper(polygon);
+    Collection<S2Region> regionCollection = new ArrayList<>();
+    regionCollection.add(s2RegionWrapper);
     return new S2RegionUnion(regionCollection);
```
Kontinuation: We can simply return `this.polygon` here. The `S2RegionWrapper` used in C++ code is only for following the lifetime expectation conventions, we don't have to worry about this in Java.

jiayuasu:
What is the status of this PR?

common/src/main/java/org/apache/sedona/common/S2Geography/PolygonGeography.java
```diff
@@ -0,0 +1,118 @@
...
+  @Override
+  public int numShapes() {
+    return polygon == null ? 0 : 1;
+  }
+
+  @Override
+  public S2Shape shape(int id) {
+    assert polygon != null;
+    return polygon.shape();
+  }
```
Kontinuation: `this.polygon` should never be null (but can be empty). If we replicate the C++ implementation:

* `dimension` should always return 2.
* `numShapes` should be `return polygon.isEmpty()? 0: 1;`.

common/src/main/java/org/apache/sedona/common/S2Geography/PolygonGeography.java
```diff
@@ -0,0 +1,118 @@
...
+  @Override
+  public void getCellUnionBound(List<S2CellId> cellIds) {
+    super.getCellUnionBound(cellIds);
+  }
+
+  @Override
+  public void encode(UnsafeOutput out, EncodeOptions opts) throws IOException {
+    // 3) Write number of polygon
+    out.writeInt(numShapes());
+    // 4) Encode polygon
```
Kontinuation: There's no need to write the number of polygon, since the polygon will not be encoded when it is empty (handled by `encodeTagged` in S2Geography). When we reach here, there must be only 1 polygon to be encoded.

common/src/main/java/org/apache/sedona/common/S2Geography/PolygonGeography.java
```diff
@@ -0,0 +1,118 @@
...
+      throw new IOException("PolygonGeography.decodeTagged error: insufficient header bytes");
+    }
+
+    // 4) Read the number of polylines (4-byte)
+    int count = in.readInt();
+    // Decode each polygon
+    if (count != 1) {
+      throw new IOException(
+          "PolygonGeography.decode error: expected exactly one polygon, but found " + count);
+    }
```
Kontinuation: These steps are not needed as well.

Kontinuation (commented):
There are some minor issues related to the revised PolygonGeography. I think this PR is close to completion.

Kontinuation (approved):
Now it looks good to me. Thank you for your great work @ZhuochengShang

PR #2014, opened 2025-06-26
[GH-2013] Geopandas.Series: Implement crs property and set_crs

### LINKED ISSUE #2013: Geopandas.GeoSeries: Implement crs property and set_crs

petern48:

- Closes #2013

Implement `crs` property and `set_crs()`

## How was this patch tested?
Added new tests


petern48:
`inplace` is not yet working. I tried fixing it a few ways (e.g modifying the internal frame or anchor), but it didn't work. I prefer to just move on for now since it's not the most important feature. Implementing `inplace` for all functions should be the same.

EDIT: nvm i got it

petern48:
Ready for review @zhangfengcdt. I'll also start properly doing draft PRs in the future and open for review when they're ready.

python/sedona/geopandas/geoseries.py
```diff
@@ -49,13 +51,8 @@ def __repr__(self) -> str:
         """
         Return a string representation of the GeoSeries in WKT format.
         """
-        try:
```
zhangfengcdt, line 54: Any reason why the try/except in __repr__, is removed here? Are we assuming to_geopandas() always succeeds. If there are scenarios where conversion might fail, I think we need to deal with it as well.

zhangfengcdt (commented):
looks good, just one inline comment

python/sedona/geopandas/geoseries.py
petern48: I originally added the old except logic thinking we need to handle the case where GeoSeries doesn't contain geometric types, but that shouldn't happen actually the so except logic isn't quite right, since GeoSeries must contain geometric types. The except code instead silently caught errors that I didn't want it to so I removed it.

I'd expect to_geopandas to work for any case where the data is small enough for geopandas to handle. This doesn't yet handle big data, but the old except code wasn't correct either since it doesn't handle the transformation from wkb to wkt (the reason behind calling to_geopandas). I think any __repr__ change is out of scope for this PR, but we can file an issue if there is a need.

zhangfengcdt (approved):

python/sedona/geopandas/geoseries.py
```diff
@@ -155,6 +162,176 @@ def __init__(
...
+        - Ellipsoid: WGS 84
+        - Prime Meridian: Greenwich
+
+        See Also
+        --------
+        GeoSeries.set_crs : assign CRS
+        GeoSeries.to_crs : re-project to another CRS
+        """
+        tmp_df = self._process_geometry_column("ST_SRID", rename="crs")
+        srid = tmp_df.to_pandas().iloc[0]
```
jiayuasu: Note that `to_pandas()` function will collect the distributed Sedona DF to the master node. This will likely cause OOM on the master. The best way to do this is via Spark's `take()` which only takes the first value of the distributed DF without collecting the entire DF.

In addition, this function assumes all records in the DF has the same CRS/SRID which is not true for Sedona in many cases. It is worth mentioning this in the docs.

jiayuasu (changes requested):

python/sedona/geopandas/geoseries.py
```diff
@@ -17,10 +17,11 @@
 
 import os
 import typing
-from typing import Any, Union
+from typing import Any, Union, Literal
 
 import geopandas as gpd
```
jiayuasu, line 22: I also saw that `geopandas` is now a required dependency of `geoseries.py`. I think it is probably fair to assume this is true?

If so, I guess `pyroj` dependency can exist here.

python/sedona/geopandas/geoseries.py
zhangfengcdt: You can also do tmp_df._to_spark_pandas_df() that returns the pspd.DataFrame (pyspark).

python/sedona/geopandas/geoseries.py
petern48: In the original geopandas implementation, they conditionally use/import pyproj, so I followed the same logic. It's not needed in anything other than the CRS functions.
```
def set_crs()
       ....
       from pyproj import CRS
```
https://github.com/geopandas/geopandas/blob/e9b58ce57a238b28ebf5eddd83d437db92c314b5/geopandas/geoseries.py#L1153

python/sedona/geopandas/geoseries.py
petern48: Thanks. Yeah I remember realizing that, but I forgot to circle back to figure out how to remove the to_pandas.

Also added the doc change. Slight nit about your wording: all CRS have to be the same in the Series, not the DF. Different series are still allowed to have different CRS in geopandas.

zhangfengcdt: I think we should overwrite the two functions `to_pandas()` and `to_geopandas` in GeoSeries to disable end users from performing these operations directly. We can still keep internal versions of them _to_pandas and _to_geopandas if needed by the implementation itself. 

In PySpark, it keeps the to_pandas() that warn users that 

```
        log_advice(
            "`to_pandas` loads all data into the driver's memory. "
            "It should only be used if the resulting pandas Series is expected to be small."
        )
```

And we should at least do this in to_geopandas() as well if we don't like the idea of disabling it.

python/sedona/geopandas/geoseries.py
```diff
@@ -17,10 +17,11 @@
...
 import pandas as pd
+from pyproj import CRS
```
jiayuasu: @petern48 can you import pyproj only when it is needed?

python/sedona/geopandas/geoseries.py
petern48: I personally prefer keeping the public method, so I added the warning message to `to_geopandas()`.

python/sedona/geopandas/geoseries.py
petern48: Yep, just did. Didn't realize I had a top level import

PR #2027, opened 2025-06-30
[GH-2028] Implementations of GeographyCollection & ShapeIndexGeography on S2Geography Apache Sedona

ZhuochengShang:

1. **GeographyCollection** class, which  takes **List<S2Geography>** 
2. Encode / decode functions of GeographyCollection object
3. **ShapeIndexGeography** class, which creates by taking empty, S2Geography or maxEdgesPerCell; 
4. ShapeIndexGeography supports **addIndex** which assigns input S2Geography with index based on S2ShapeIndex add function. 
5. ShapeIndexGeography only contains **encode** function, using VectorCoder encoding. 
6. EncodedShapeIndexGeography class supports only **decode** function, using VectorCoder decoding. It automatically decode EncodedShapeIndexGeography based on corresponding geography type. This class also supports **addIndex** same as  ShapeIndexGeography.


jiayuasu:
@ZhuochengShang please resolve the conflicts

common/src/main/java/org/apache/sedona/common/S2Geography/ShapeIndexGeography.java
```diff
@@ -0,0 +1,179 @@
...
+      if (opts.getCodingHint() == EncodeOptions.CodingHint.FAST) {
+        throw new IOException("Lazy output only supported with COMPACT hint");
+      }
+      // 1) Encode each shape in the index, using our “compact” path when possible
+      for (S2Shape rawShape : shapeIndex.getShapes()) {
+        // If it's a polygon or 1-chain polyline, do the custom lax encoding…
+        if (!customCompactTaggedShapeEncoder(rawShape, tmpOut)) {
+          // …otherwise fall back to the usual S2Geography encode
+          // (we know every shape in this index is actually an S2Geography)
+          ((S2Geography) rawShape).encode((UnsafeOutput) tmpOut, opts);
```
Kontinuation: > "(we know every shape in this index is actually an S2Geography)"

This is not true. For instance, when we call `addIndex(polylineGeography)`, we iterate over the shapes of `polylineGeography` and add them into `shapeIndex`. Each shape is an `S2Polyline`, which is not an S2Geography.

common/src/main/java/org/apache/sedona/common/S2Geography/ShapeIndexGeography.java
```diff
@@ -0,0 +1,179 @@
...
+      Iterable<S2Point> verts = shape.chain(0);
+      S2LaxPolylineShape laxPolyline = S2LaxPolylineShape.create(verts);
+      S2LaxPolylineShape.COMPACT_CODER.encode(laxPolyline, out);
+      return true;
+    }
+    // Fallback: encode the original shape directly
+    else {
+      return false;
+    }
+  }
```
Kontinuation: The C++ S2Geography does this because the C++ S2 library has `EncodedS2LaxPolygonShape` and `EncodedS2LaxPolylineShape`, these types supports lazy decoding. Java S2 library does not have these features, so I think there's no point implementing `customCompactTaggedShapeEncoder` or lazy mode in general.

Simply using `VectorCoder` and `S2ShapeIndexCoder` for encoding and decoding S2ShapeIndex should be good enough. You can refer to this test case for details: https://github.com/google/s2-geometry-library-java/blob/master/tests/tests/com/google/common/geometry/S2ShapeIndexCoderTest.java#L372-L387

common/src/main/java/org/apache/sedona/common/S2Geography/GeographyIndex.java
```diff
@@ -0,0 +1,120 @@
...
+  public GeographyIndex(S2ShapeIndex.Options options) {
+    this.index = new S2ShapeIndex(options);
+    this.values = new int[0];
+  }
+
+  public void add(S2Geography geog, int value) {
+    // Determine total slots needed in the array
+    int currentLength = (values == null ? 0 : values.length);
+    int needed = currentLength + geog.numShapes();
+    values = expand(values, needed);
```
Kontinuation: I suggest defining `values` as an `ArrayList`, so that it grows its space exponentially.

[The original C++ implementation](https://github.com/paleolimbot/s2geography/blob/main/src/s2geography/index.h#L30-L37) does not exhibit the exponential growth behavior and may incur lots of reallocation overhead, I think it could be a problem. @paleolimbot

paleolimbot: Yes, I should probably be just calling `value_.push_back()`?

paleolimbot: While this particular bit of code has done the trick for the sizes of geography vectors that we usually see in R, I would say that it is a particular weak point in the implementation because it indexes every single edge for ever single member. Ideally it would only store a covering per item.

common/src/main/java/org/apache/sedona/common/S2Geography/ShapeIndexGeography.java
```diff
@@ -0,0 +1,138 @@
...
+    //  // avoid unexpected client behavior, this method continues to add shapes
+    //  // even once the specified S2MemoryTracker limit has been exceeded.
+    //  const int id = shapes_.size();
+    //  shape->id_ = id;
+    //  mem_tracker_.AddSpace(&shapes_, 1);
+    //  shapes_.push_back(std::move(shape));
+    //  MarkIndexStale();
+    //  return id;
+    return lastId;
+  }
```
Kontinuation: I don't find the result of `ShapeIndexGeography::Add` being used anywhere in C++ code. We can change the return type to void.

Kontinuation (approved):
It looks good to me. Can you update the description of this PR and remove unnecessary contents came from the PR template?

jiayuasu:
@ZhuochengShang Please check if the copilot comments make sense or not

jiayuasu (approved):

PR #2046, opened 2025-07-02
[GH-2045] Geopandas.GeoSeries: Implement fillna

### LINKED ISSUE #2045: Geopandas.GeoSeries: Implement fillna

petern48:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2045

Implement fillna

## How was this patch tested?
Added tests


python/sedona/geopandas/geoseries.py
```diff
@@ -1374,8 +1374,128 @@ def notnull(self) -> pspd.Series:
...
+        dtype: geometry
+
+        See Also
+        --------
+        GeoSeries.isna : detect missing values
+        """
+        from shapely.geometry.base import BaseGeometry
+        from geopandas.array import GeometryArray
+
+        # TODO: Implement limit
```
zhangfengcdt: Can you create a github issue and link the todo to that issue?

python/sedona/geopandas/geoseries.py
```diff
@@ -1374,8 +1374,128 @@ def notnull(self) -> pspd.Series:
...
+            result = self._query_geometry_column(select, col, "")
+        elif isinstance(value, (GeoSeries, GeometryArray, gpd.GeoSeries)):
+
+            if isinstance(value, (gpd.GeoSeries, GeometryArray)):
+                value = GeoSeries(value)
+
+            # Replace all None's with empty geometries
+            value = value.fillna(None)
+
+            select = f"COALESCE(`L`, `R`)"
```
zhangfengcdt, line 2125: Can we comment on why using coalesce here?

python/sedona/geopandas/geoseries.py
```diff
@@ -1374,8 +1374,128 @@ def notnull(self) -> pspd.Series:
...
+        if pd.isna(value) == True or isinstance(value, BaseGeometry):
+            if (
+                value is not None and pd.isna(value) == True
+            ):  # ie. value is np.nan or pd.NA:
+                value = "NULL"
+            else:
+                if value is None:
+                    from shapely.geometry import GeometryCollection
+
+                    value = GeometryCollection()
```
zhangfengcdt, line 2110: Why do we use GeometryCollection here?

petern48: "If None is passed, NA values will be filled with GEOMETRYCOLLECTION EMPTY." This is already in the docstring, so I don't see the need for an additional comment

petern48:
Tests are now passing again after merging in the recent changes

python/sedona/geopandas/geoseries.py
```diff
@@ -537,7 +537,9 @@ def _to_geopandas(self) -> gpd.GeoSeries:
         try:
             return gpd.GeoSeries(
                 pd_series.map(
-                    lambda wkb: shapely.wkb.loads(bytes(wkb)) if wkb else None
+                    lambda wkb: (
+                        shapely.wkb.loads(bytes(wkb)) if not pd.isna(wkb) else None
```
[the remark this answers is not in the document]
petern48: Empty byte strings are unexpected behavior, so I'd rather not check for them here and let them error in the event that they occur.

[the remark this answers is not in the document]
zhangfengcdt: I think the issue is that if user get that error, e.g., WKBReadingError because of empty byte strings are not valid WKB representations, they won't be able to know what to do and where that wkb come from - unless if they pass that as a parameter explicitly. We would either:

- add a document to tell user what the error WKBReadingError meant to them or
- process / skip them as NA values

Either way, this should be documented in the public method, i.e., to_geopandas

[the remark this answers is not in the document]
petern48: I agree that it would be nice to tell the user what the error indicates, but as I mentioned, this isn't expected behavior. In the case where an empty byte string is returned, I think this would be our error (the Sedona developer's error) and not the user's error, so I don't know what to tell them. Do you know what might cause an empty byte string to be returned here? What would you recommend specifying in the docstring?

zhangfengcdt (approved):
LGTM

PR #2051, opened 2025-07-02
[GH-2052] Implement geospatial functions of Distance, Predicates, Projection

ZhuochengShang:

- Distance function:  min/max distance between two geographies, minimumClearanceLineBetween two geographies
- Projection function: MercatorProjection, OrthographicProjection
- Predicates function: intersects, equal, contains, intersectsBox


common/src/main/java/org/apache/sedona/common/S2Geography/Accessors.java
```diff
@@ -0,0 +1,294 @@
...
+      if (!shape.isEmpty()) return false;
+    }
+    return true;
+  }
+
+  public boolean S2_isCollection(PolygonGeography polygonGeography) {
+    int numOuterLoops = 0;
+    for (int i = 0; i < polygonGeography.polygon.numLoops(); i++) {
+      S2Loop loop = polygonGeography.polygon.loop(i);
+      if (loop.depth() == 0 && ++numOuterLoops > 10) {
```
Kontinuation: I guess the correct code should be:

```suggestion
      if (loop.depth() == 0 && ++numOuterLoops > 1) {
```

common/src/main/java/org/apache/sedona/common/S2Geography/Accessors.java
```diff
@@ -0,0 +1,294 @@
...
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Returns true if `geog` is a “collection” (i.e. multi‐point, multi‐linestring, or multi‐polygon)
+   * rather than a single simple feature.
+   */
+  public boolean s2IsCollection(S2Geography geog) {
```
Kontinuation: Why don't we name it as `S2_isCollection` as well? Function overloading with sub type as parameter is allowed in Java.

common/src/main/java/org/apache/sedona/common/S2Geography/Accessors.java
```diff
@@ -0,0 +1,294 @@
...
+    if (S2_dimension(geog) != 2) return 0;
+
+    switch (geog.kind) {
+      case POLYGON:
+        if (geog != null) return S2_area((PolygonGeography) geog);
+      case GEOGRAPHY_COLLECTION:
+        if (geog != null) return S2_area((GeographyCollection) geog);
+    }
+    // TODO: build
+    return 0;
```
Kontinuation: I cannot figure out why we need to build a polygon geography to compute its area here: https://github.com/paleolimbot/s2geography/blob/main/src/s2geography/accessors.cc#L127-L128
@paleolimbot

common/src/main/java/org/apache/sedona/common/S2Geography/Distance.java
```diff
@@ -0,0 +1,192 @@
...
+    S2ShapeIndex index2 = geo2.shapeIndex;
+
+    S2ClosestEdgeQuery query =
+        S2ClosestEdgeQuery.builder().setIncludeInteriors(false).build(index1);
+    S2ClosestEdgeQuery.ShapeIndexTarget queryTarget =
+        S2ClosestEdgeQuery.createShapeIndexTarget(index2);
+
+    Optional<S2BestEdgesQueryBase.Result> resultVisitor = query.findClosestEdge(queryTarget);
+
+    if (resultVisitor.get().edgeId() == -1) {
```
Kontinuation: Should test if `resultVisitor` is present since `resultVisitor` is an Optional.

common/src/main/java/org/apache/sedona/common/S2Geography/Distance.java
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  /**
+   * The Java equivalent of:
+   *
+   * <p>std::pair<S2Point,S2Point> GetEdgePairClosestPoints(a0,a1,b0,b1) # returns the shortest
+   * possible line between x and y
+   */
+  public static Pair<S2Point, S2Point> getEdgePairClosestPoints(
+      S2Point a0, S2Point a1, S2Point b0, S2Point b1) {
```
Kontinuation, line 133: Why not directly use `S2EdgeUtil.getEdgePairClosestPoints`?

common/src/main/java/org/apache/sedona/common/S2Geography/Predicates.java
```diff
@@ -0,0 +1,87 @@
...
+    PolygonGeography polygonGeography = new PolygonGeography(polygon);
+    // c++ is using MutableS2ShapeIndex which is missing in Java
+    // MutableS2ShapeIndex is a class for in-memory indexing of polygonal geometry.
+    // The objects in the index are known as "shapes", and may consist of points,
+    // polylines, and/or polygons, possibly overlapping.  The index makes it very
+    // fast to answer queries such as finding nearby shapes, measuring distances,
+    // testing for intersection and containment, etc.  It is one of several
+    // implementations of the S2ShapeIndex interface (see EncodedS2ShapeIndex).
+    GeographyIndex geographyIndex = new GeographyIndex();
+    geographyIndex.add(polygonGeography, 0);
```
Kontinuation: `S2ShapeIndex` is already mutable. Does this simplified version work?
```java
S2ShapeIndex index = new S2ShapeIndex();
index.add(loop);
```

common/src/main/java/org/apache/sedona/common/S2Geography/Accessors.java
paleolimbot: The Area operation is one of the things that we absolutely need an `S2Polygon` instance for (i.e., not just shapes), so if somebody calls `s2_area()` on a (e.g.) ShapeIndexGeography the downcasts would fail. We can also throw here since we can ensure this doesn't happen when you implement st_area.

Kontinuation: OK. I think we can throw an exception here. ST_Area is only expected to work with simple feature instances.

common/src/main/java/org/apache/sedona/common/S2Geography/Predicates.java
ZhuochengShang: the main difference is c++ code is using S2LaxLoopShape, which is more simplified construction of loop shape. Rather than the S2Loop requires more steps when initializing, such as building bounding box.  Double checked the Java library and notice that ShapeIndex in Java is same as MutableS2ShapeIndex in c++.

Kontinuation: `S2Loop` does not allow duplicated nodes, and we don't have `S2LaxLoopShape`, does this mean that we need to deduplicate the vertices by ourselves before constructing a `S2Loop`? The input `rect` may degenerate as a single point, all four vertices will be the same value in this case.

paleolimbot: Deduplicating vertices is probably a good idea (we should do it in C++, too). We've hit some annoying/difficult to work around bugs in R and Python where "valid" GEOS geometries errored when imported into S2.

jiayuasu (approved):

PR #2057, opened 2025-07-03
[GH-2050] Geopandas.GeoSeries: Implement align=False for _row_wise_operation + support indexes

### LINKED ISSUE #2050: Geopandas.GeoSeries: Implement `align=False` behavior for _row_wise_operation
As mentioned in [#2038](https://github.com/apache/sedona/pull/2038), `align=False` is not yet supported. It seems like Pandas on Pyspark maintains an artificial "natural order column" that's based on `F.monotonically_increasing_id()`, so we should try to use that to perform the align=False behavior. It's not immediately clear whether this will match the geopandas ordering.

```
NATURAL_ORDER_COLUMN_NAME = "__natural_order__"

HIDDEN_COLUMNS = {NATURAL_ORDER_COLUMN_NAME}
```
https://github.com/apache/spark/blob/a1e628574b7d9cdf89472fa550ecc41f8a871b98/python/pyspark/pandas/internal.py#L77-L79

petern48:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2050

- Fix the `__init__` to support traditional pandas indexes
- Implement `align=False` behavior using existing `NATURAL_ORDER_COLUMN_NAME`

## How was this patch tested?
Added tests

### DIFF
--- a/python/sedona/geopandas/geoseries.py
+++ b/python/sedona/geopandas/geoseries.py
@@ -39,7 +39,10 @@ from sedona.geopandas.base import GeoFrame
 from sedona.geopandas.geodataframe import GeoDataFrame
 from sedona.geopandas.geoindex import GeoIndex
 
-from pyspark.pandas.internal import SPARK_DEFAULT_INDEX_NAME  # __index_level_0__
+from pyspark.pandas.internal import (
+    SPARK_DEFAULT_INDEX_NAME,  # __index_level_0__
+    NATURAL_ORDER_COLUMN_NAME,
+)
 
 
 class GeoSeries(GeoFrame, pspd.Series):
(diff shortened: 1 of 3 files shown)


python/sedona/geopandas/geoseries.py
```diff
@@ -158,14 +161,7 @@ def __init__(
...
             )
             # initialize the parent class pyspark Series with the pandas Series
-            super().__init__(
-                data=pdf,
-                index=index,
-                dtype=dtype,
-                name=name,
-                copy=copy,
-                fastpath=fastpath,
-            )
```
petern48, line 168: This was the fix for index support. It was a lot simpler than I expected because I forgot pandas-on-pyspark does it for us.

python/sedona/geopandas/geoseries.py
```diff
@@ -158,14 +161,7 @@ def __init__(
...
+            super().__init__(data=pdf)
```
petern48, line 164: This was the fix for index support. It was a lot simpler than I expected because I forgot pandas-on-pyspark does it for us.

petern48:
The doc changes are simply to replicate the same example in the original geopandas doc string for simplicity. The original one I wrote didn't have an example of using `index=`

jiayuasu (approved):

Issue #2066, opened 2025-07-06
Geopandas.SpatialIndex _init_ and refactor


zhangfengcdt:
GeoPandas exposes the Sort-Tile-Recursive R-tree from shapely on any GeoDataFrame and GeoSeries using the GeoSeries.sindex property. This page outlines its options and common usage patterns.

This issue is to implement the Spatial Index under the sedona.geopandas project.

Reference:
https://geopandas.org/en/stable/docs/reference/sindex.html
https://geopandas.org/en/stable/docs/user_guide/spatial_indexing.html

PR #2067, opened 2025-07-06
[GH-2066] Implement spatial index sindex query() and unit tests

zhangfengcdt:

This PR is to implement GeoPandas's spatial index using the GeoSeries.sindex property. 

## How was this patch tested?
test_sindex


python/sedona/geopandas/sindex.py
```diff
@@ -0,0 +1,137 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import numpy as np
+from pyspark.sql import DataFrame as PySparkDataFrame
+
+
+class SpatialIndex:
```
jiayuasu, line 26: So looks like this PR is to provide a template for the spatial index, not the actual implementation?

zhangfengcdt: Yes, this is to correct the previous incorrect interface (GeoIndex vs SpatialIndex) and implement the init, test, and size, is_empty functions. The rest of the functions in the SpatialIndex (i.e., query, nearest, and intersection) will be implemented separately.

python/sedona/geopandas/sindex.py
```diff
@@ -0,0 +1,137 @@
...
+        if isinstance(geometry, np.ndarray):
+            self.geometry = geometry
+        elif isinstance(geometry, PySparkDataFrame):
+            if column_name is None:
+                raise ValueError("column_name must be specified for PySpark DataFrame.")
+            if column_name not in geometry.columns:
+                raise ValueError(
+                    f"Column '{column_name}' does not exist in the DataFrame."
+                )
+            self.geometry = geometry[column_name].values
```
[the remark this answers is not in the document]
zhangfengcdt: It now takes PySparkDataFrame with indexing column_name instead. We don't want to collect the column to local array in init but keep it as distributed dataframe / column.

python/sedona/geopandas/sindex.py
```diff
@@ -0,0 +1,224 @@
...
+                result_rdd = RangeQuery.SpatialRangeQuery(
+                    self._indexed_rdd, geometry, True, True
+                )
+            else:  # Default to intersects
+                result_rdd = RangeQuery.SpatialRangeQuery(
+                    self._indexed_rdd, geometry, False, True
+                )
+
+            # Convert results to indices
+            results = list(range(result_rdd.count()))
```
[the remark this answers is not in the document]
zhangfengcdt: The results already are indices, so we don't need (cannot) use lambda to do this.

python/sedona/geopandas/sindex.py
```diff
@@ -0,0 +1,223 @@
...
```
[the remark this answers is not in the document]
zhangfengcdt: result_rdd.collect() should be sufficient according to the SpatialRangeQuery.

python/sedona/geopandas/sindex.py
zhangfengcdt: @jiayuasu I have just implemented the build index and query() method for the SpatialIndex. I'd prefer implementing other methods with following PRs because this PR is already large. We'd added more specific tests with their functions added for nearest and intersection on the spatial index.

python/sedona/geopandas/sindex.py
```diff
@@ -0,0 +1,225 @@
...
+
+        This uses Sedona's built-in indexing functionality.
+        """
+
+        # Convert index_type string to Sedona IndexType enum
+        index_type_map = {"strtree": IndexType.RTREE, "quadtree": IndexType.QUADTREE}
+        sedona_index_type = index_type_map.get(self.index_type.lower(), IndexType.RTREE)
+
+        # Create a SpatialRDD from the DataFrame
+        spatial_rdd = Adapter.toSpatialRdd(self._dataframe, column_name)
```
jiayuasu: We are deprecating the Adapter. Please consider using StructuredAdapter. See: https://sedona.apache.org/latest/tutorial/sql/#convert-between-dataframe-and-spatialrdd

zhangfengcdt: Yes, i have tried using:

```
StructuredAdapter.toSpatialRdd(self._dataframe, column_name)
```
And it looks like it only works with Point data in the RangeQuery on the Rdd converted this way.  For other types, it throws errors.

Here is the unit test that used to test StructuredAdapter with Range Query:

```

    def test_build_index_and_range_query(self):
        from sedona.spark.core.SpatialRDD import PointRDD
        from sedona.spark.core.enums import IndexType
        from sedona.spark.core.spatialOperator import RangeQuery
        from shapely.geometry import Polygon

        # # Create a dataframe with points
        # xys = [(i, i // 100, i % 100) for i in range(1000)]
        # spatial_df = self.spark.createDataFrame(xys, ["id", "x", "y"]).selectExpr(
        #     "id", "ST_Point(x, y) AS geometry"
        # )

        # Create a spatial DataFrame with polygons
        polygons_data = [
            (1, "POLYGON((0 0, 1 0, 1 1, 0 1, 0 0))"),
            (2, "POLYGON((1 1, 2 1, 2 2, 1 2, 1 1))"),
            (3, "POLYGON((2 2, 3 2, 3 3, 2 3, 2 2))"),
            (4, "POLYGON((3 3, 4 3, 4 4, 3 4, 3 3))"),
            (5, "POLYGON((4 4, 5 4, 5 5, 4 5, 4 4))"),
        ]

        df = self.spark.createDataFrame(polygons_data, ["id", "wkt"])
        spatial_df = df.withColumn("geometry", expr("ST_GeomFromWKT(wkt)"))

        # Convert to SpatialRDD
        spatial_rdd = StructuredAdapter.toSpatialRdd(spatial_df, "geometry")

        # Analyze and build index on the spatial RDD
        spatial_rdd.analyze()
        spatial_rdd.buildIndex(IndexType.RTREE, False)

        # Create a range query window (a bounding box)
        boundary = spatial_rdd.boundaryEnvelope
        # Create a query window within the boundary
        query_window = Polygon([
            (25, 25),
            (75, 25),
            (75, 75),
            (25, 75),
            (25, 25)
        ])

        query_point = Point(2.2, 2.2)

        # Perform range query
        query_result = RangeQuery.SpatialRangeQuery(
            spatial_rdd, query_point, True, True
        )

        # Assertions
        result_count = query_result.count()

        assert result_count >= 0, f"Expected at least one result, got {result_count}"
```

It works with Point (commented out) but not not Polygon.


--> 
```
scala.MatchError: [0,3,2000000022,4800000060,284e4f47594c4f50,2033202c32203228,202c332033202c32,322032202c332032,2929,500000032,4000000000000000,4000000000000000,4008000000000000,4000000000000000,4008000000000000,4008000000000000,4000000000000000,4008000000000000,4000000000000000,4000000000000000,500000001] (of class org.apache.spark.sql.catalyst.expressions.UnsafeRow)
	at org.apache.sedona.python.wrapper.utils.implicits$GeometryEnhancer.userDataToUtf8ByteArray(implicits.scala:52)
```

jiayuasu: Oh, then this is a bug I guess. Can you help fix it? The problem with the old Adapter is that the data types of non-spatial columns are lost unless during the conversion.

zhangfengcdt: Sure, i will investigate how to fix the StructuredAdapter before changing this PR then.

zhangfengcdt: Now it uses StructuredAdapter after the bug fix - https://github.com/apache/sedona/pull/2073

jiayuasu (approved):

PR #2071, opened 2025-07-08
[GH-2070] Geopandas.GeoSeries: Implement `get_geometry`

### LINKED ISSUE #2070: Geopandas.GeoSeries: Implmement `get_geometry`

petern48:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2070

Implement get_geometry

## How was this patch tested?
Added tests


python/sedona/geopandas/geoseries.py
```diff
@@ -829,9 +829,78 @@ def get_precision(self):
...
+        Notes
+        -----
+        Simple geometries act as collections of length 1. Any out-of-range index value
+        returns None.
+
+        Examples
+        --------
+        >>> from shapely.geometry import Point, MultiPoint, GeometryCollection
+        >>> s = geopandas.GeoSeries(
+        ...     [
```
jiayuasu, line 968: How about POLYGON EMPTY?

petern48: Polygon Empty isn't special, but MultiPolygon Empty is interesting, since it's a collection. Our `ST_GeometryN` errors when we use a negative index out of scope of the length of the Multi geometry object.

e.g
```sql
SELECT ST_GeometryN(MultiPolygon(Point(1, 1)), -2);  -- errors

SELECT ST_GeometryN(MultiPolygon(), -1);  -- errors
```

Geopandas returns null in this case. PostGIS also returns null in this case (the PostGIS implementation is already very different than ours since they use 1-indexing).

We can either fix this on the java Sedona level, Python dataframe level or just the geopandas level (case statement). Which do you think @jiayuasu ?

petern48: Actually, I thought about it some more and decided to implement on geopandas level. I forgot the existing negative index handling logic was already implemented there.

jiayuasu (approved):

zhangfengcdt (commented):
looks like the sheply 1 still has issue with one of the tests

jiayuasu:
@petern48 I am fine dropping Shapely 1.x support in the next Sedona major release. What do you think? @Kontinuation

petern48:
We don't necessarily have to drop support for shapely 1 in all of sedona if you don't want to. We could also just drop it for geopandas specifically, or skip this test for shapely 1 and accept the difference. Of course, updating documentation however necessary. Dropping support would make my life easier, but it's probably not necessarily at this point.

Looking at the error output, it looks like shapely 1, automatically interprets `Polygon()` as `GEOMETRY COLLECTION`.

```python
        expected = gpd.GeoSeries(
            [
                Point(0, 0),
                Point(0, 0),
                MultiPoint([(0, 0), (1, 1), (0, 1), (1, 0)]),
                Polygon(),
                None,
            ]
        )
```
From the logs: (idk why it's formatted this way, but GeometryCollection Empty corresponds to the Polygon() object)
```
expected = 0                              POINT (0.00000 0.00000)
1                              POINT (0.00000 0.00000)
2    MUL...                        GEOMETRYCOLLECTION EMPTY
4                                                 None
```

petern48:
I personally think it's reasonable to just skip the test for shapely < 2.0.0 and update the docs about the behavior. Pushed a change. What do you think?

Issue #2072, opened 2025-07-08
StructuredAdapter Does Not Work With RangeQuery on Geometries Other Than Point


zhangfengcdt:
It looks like StructuredAdapter toSpatialRdd only works with Point data in the RangeQuery on the Rdd converted this way. For other types, it throws errors.

scala.MatchError: [0,3,2000000022,4800000060,284e4f47594c4f50,2033202c32203228,202c332033202c32,322032202c332032,2929,500000032,4000000000000000,4000000000000000,4008000000000000,4000000000000000,4008000000000000,4008000000000000,4000000000000000,4008000000000000,4000000000000000,4000000000000000,500000001] (of class org.apache.spark.sql.catalyst.expressions.UnsafeRow)
	at org.apache.sedona.python.wrapper.utils.implicits$GeometryEnhancer.userDataToUtf8ByteArray(implicits.scala:52)

This issue was found during work on https://github.com/apache/sedona/pull/2067

PR #2073, opened 2025-07-08
[GH-2072] Fix Structured Adapter Issue with Spatial Queries

zhangfengcdt:

This PR fixed the issue of StructuredAdapter that does not work with RangeQuery on geometries.

## How was this patch tested?
TestStructuredAdapter

### DIFF
--- a/python/tests/sql/test_structured_adapter.py
+++ b/python/tests/sql/test_structured_adapter.py
@@ -18,6 +18,11 @@ import glob
 import tempfile
 
 from pyspark.sql import DataFrame
+from pyspark.sql.functions import expr
+from shapely.geometry.point import Point
+
+from sedona.spark.core.enums import IndexType
+from sedona.spark.core.spatialOperator import RangeQuery
 
 from sedona.spark.core.SpatialRDD import CircleRDD
 from sedona.spark.core.enums import GridType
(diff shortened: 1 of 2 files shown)


spark/common/src/main/scala/org/apache/sedona/python/wrapper/utils/implicits.scala
```diff
@@ -50,10 +50,13 @@ object implicits {
 
     def userDataToUtf8ByteArray: Array[Byte] = {
       geometry.getUserData match {
+        // Case when user data is null: return an empty UTF-8 byte array
         case null => EMPTY_STRING.getBytes(StandardCharsets.UTF_8)
+        // Case when user data is a String: convert the string to a UTF-8 byte array
         case data: String => data.getBytes(StandardCharsets.UTF_8)
+        // Case for any other type: convert to string, then to a UTF-8 byte array
+        case data => data.toString.getBytes(StandardCharsets.UTF_8)
```
jiayuasu, line 63: Why convert it to string? I believe this will lose the types of data in `UserData` field.

python/tests/sql/test_structured_adapter.py
```diff
@@ -80,3 +85,124 @@ def test_spatial_partitioned_write(self):
...
+            5,
+            "MULTIPOLYGON(((4 4, 5 4, 5 5, 4 5, 4 4)), ((4.1 4.1, 4.2 4.1, 4.2 4.2, 4.1 4.2, 4.1 4.1)))",
+        ),
+    ]
+
+    df = self.spark.createDataFrame(mixed_data, ["id", "wkt"])
+    spatial_df = df.withColumn("geometry", expr("ST_GeomFromWKT(wkt)"))
+
+    # Convert to SpatialRDD
+    spatial_rdd = StructuredAdapter.toSpatialRdd(spatial_df, "geometry")
```
jiayuasu: Please add tests to make sure the StructuredAdapter.toDf can get the exact same DF back including the non-spatial columns with correct values and types.

spark/common/src/main/scala/org/apache/sedona/python/wrapper/utils/implicits.scala
zhangfengcdt: The issue is that we cannot directly convert a generic Object (data) to bytes using getBytes(StandardCharsets.UTF_8) unless the object is a String or implements such a method. The issue here is caused by UnsafeRow object, so I am adding that specifically but keep the rest (unknown types) cast to string and getBytes.

python/tests/sql/test_structured_adapter.py
zhangfengcdt: Sure, I have added a test (test_toDf_preserves_columns_with_proper_types) that specifically tests the toDf method to preserve the type and values.

Issue #2074, opened 2025-07-09
Feature Request: Integrated with libpostal to provide ExpandAddress and ParseAddress functions


james-willis:
Address data is closely related to geospatial type data. adding function to enable spark/sedona users to work with libpostal to support the parsing and normalization of addresses would be great. I just released a fork of jpostal with the focus on making spark integration easier. I think now is a great time to take on this work.

https://github.com/wherobots/jpostal

PR #2077, opened 2025-07-09
[GH-2074] Add libpostal integration

james-willis:

- Yes, and the PR name follows the format `[GH-2074] my subject`. Closes [#2074](https://github.com/apache/sedona/issues/2074)

This add ExpandAddress and ParseAddress functions backed by libpostal, including support for code gen.

## How was this patch tested?
Unit Tests

### DIFF
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -65,6 +65,9 @@ jobs:
           - spark: 4.0.0
             scala: 2.13.8
             jdk: '17'
+          - spark: 3.5.4
+            scala: 2.12.18
+            jdk: '17'
           - spark: 3.5.0
             scala: 2.13.8
             jdk: '11'
@@ -103,13 +106,8 @@ jobs:
         run: |
           SPARK_COMPAT_VERSION=${SPARK_VERSION:0:3}
 
-          if [ "${SPARK_VERSION}" == "3.5.0" ]; then
-              pip install pyspark==3.5.0 pandas shapely apache-sedona pyarrow
-              export SPARK_HOME=$(python -c "import pyspark; print(pyspark.__path__[0])")
-          fi
-
-          if [ "${SPARK_VERSION}" == "4.0.0" ]; then
-              pip install pyspark==4.0.0 pandas shapely apache-sedona pyarrow
+          if [[ "${SPARK_VERSION}" == "3.5"* ]] || [[ "${SPARK_VERSION}" == "4."* ]]; then
+              pip install pyspark==$SPARK_VERSION pandas shapely apache-sedona pyarrow
               export SPARK_HOME=$(python -c "import pyspark; print(pyspark.__path__[0])")
           fi
 
--- a/python/sedona/spark/sql/st_functions.py
+++ b/python/sedona/spark/sql/st_functions.py
@@ -32,6 +32,7 @@ from sedona.spark.sql.dataframe_api import (
 _call_st_function = partial(call_sedona_function, "st_functions")
 
 
+@validate_argument_types
 def ExpandAddress(address: ColumnOrName):  # noqa: N802
     """Normalize an address string into its canonical forms.
 
--- a/python/sedona/spark/sql/st_functions.py
+++ b/python/sedona/spark/sql/st_functions.py
@@ -32,6 +32,25 @@ from sedona.spark.sql.dataframe_api import (
 _call_st_function = partial(call_sedona_function, "st_functions")
 
 
+def ExpandAddress(address: ColumnOrName):  # noqa: N802
+    """Normalize an address string into its canonical forms.
+
+    :param address: The address string or column to normalize.
+    :return: An array of normalized address strings.
+    """
+    return _call_st_function("ExpandAddress", address)
+
+
+@validate_argument_types
+def ParseAddress(address: ColumnOrName):  # noqa: N802
+    """Parse an address string into its components (label/value pairs).
+
+    :param address: The address string or column to parse.
+    :return: An array of maps with label/value pairs for address components.
+    """
+    return _call_st_function("ParseAddress", address)
+
+
 @validate_argument_types
 def GeometryType(geometry: ColumnOrName):
     """Return the type of the geometry as a string.
--- a/python/tests/sql/test_dataframe_api.py
+++ b/python/tests/sql/test_dataframe_api.py
@@ -1785,27 +1785,38 @@ class TestDataFrameAPI(TestBase):
         ).collect()
 
     def test_expand_address_df_api(self):
-        df = (
+        input_df = (
             self.spark.range(1)
             .selectExpr(
                 "'781 Franklin Ave Crown Heights Brooklyn NY 11216 USA' as address"
             )
             .cache()
         )  # cache to avoid Constant Folding Optimization
-        df = df.select(ExpandAddress("address").alias("normalized"))
+
         # Actually running downloads the model and is very expensive, so we just check the plan
         # Checking the plan should allow us to verify that the function is correctly registered
-        df.explain()
+        assert input_df.select(
+            ExpandAddress("address").alias("normalized")
+        ).sameSemantics(
+            input_df.select(f.expr("ExpandAddress(address)").alias("normalized"))
+        )
+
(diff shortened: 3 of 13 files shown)


python/sedona/spark/sql/st_functions.py
```diff
@@ -20,7 +20,7 @@
 from typing import Optional, Union
 
 from pyspark.sql import Column
-from pyspark.sql.functions import lit
+from pyspark.sql.functions import lit, col
```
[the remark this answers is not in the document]
james-willis: Wrong

james-willis:
@Imbruced Any idea what might be causing the test failure in this new version I added to the test matrix? 

spark 3.5.4, java 17, scala 2.12.8

Edit: I see it is that we only set spark home when the spark version is 3.5.0 or 4.0.0. I will update this.

jiayuasu:
Jpostal requires Java 17? Sedona is compiled against Java 11. Will this even pass our Java tests in CI?

jiayuasu:
@james-willis Since the jpostal under Wherobots is already a fork, why not just compile it against Java 11? What specific Java 16+ features were used in the JNI binding and can we drop it?

james-willis:
actually turned out to be really easy to support java 11: https://github.com/wherobots/jpostal/pull/13

.github/workflows/java.yml
```diff
@@ -65,6 +65,9 @@ jobs:
           - spark: 4.0.0
             scala: 2.13.8
             jdk: '17'
+          - spark: 3.5.4
```
jiayuasu, line 68: Why is this needed?

pom.xml
```diff
@@ -422,6 +422,18 @@
             <id>Spark Packages</id>
             <url>https://repos.spark-packages.org/</url>
         </repository>
+<!--        TODO remove me and use jpostal actual release-->
+        <repository>
```
jiayuasu: Do we still need this?

.github/workflows/java.yml
james-willis: since i fixed the jpostal java 17 thing it isnt strictly. I left it because It matches AWS Glue 5.0 and I think thats a good piece of test coverage.

pom.xml
james-willis: yesh only just now did I finish publishing the release version

python/tests/sql/test_dataframe_api.py
```diff
@@ -1783,3 +1783,29 @@ def test_lof(self):
...
+            self.spark.range(1)
+            .selectExpr(
+                "'781 Franklin Ave Crown Heights Brooklyn NY 11216 USA' as address"
+            )
+            .cache()
+        )  # cache to avoid Constant Folding Optimization
+        df = df.select(ExpandAddress("address").alias("normalized"))
+        # Actually running downloads the model and is very expensive, so we just check the plan
+        # Checking the plan should allow us to verify that the function is correctly registered
+        df.explain()
```
jiayuasu: Add a test case to actually check the plan?

python/tests/sql/test_dataframe_api.py
```diff
@@ -1783,3 +1783,29 @@ def test_lof(self):
...
+
+    def test_parse_address_df_api(self):
+        df = (
+            self.spark.range(1)
+            .selectExpr(
+                "'781 Franklin Ave Crown Heights Brooklyn NY 11216 USA' as address"
+            )
+            .cache()
```
jiayuasu, line 1812: Why is the cache needed?

jiayuasu:
Please also remove all `show` and `print` command

python/tests/sql/test_dataframe_api.py
james-willis: Couple reasons I didnt do this:

1. AFAIK there is not a real opportunity to inspect the plan contents. The pyspark API doesn't expose anything to programmatically interact with the query plan: https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/dataframe.html
2. I didn't think it was particularly valuable. If the explain doesn't throw that implies that the function was found on the path, which IMO is the primary purpose of this test.

python/tests/sql/test_dataframe_api.py
james-willis: to avoid constant folding causing the evaluation of the statement in the logical optimizer. without the cache the function call gets evaluated which we are trying to avoid.

pom.xml
james-willis: This is fixed. I only had this here to test that Java 11 version of jpostal before I made a non-snapshot release of it.

python/tests/sql/test_dataframe_api.py
jiayuasu: Please uncache it once it is done

python/tests/sql/test_dataframe_api.py
jiayuasu: If not, then check it in the Scala side? See this approach: https://github.com/apache/sedona/blob/77685486a13ce446bbe9887089f7d16b912cc94a/spark/common/src/test/scala/org/apache/sedona/sql/predicateJoinTestScala.scala#L56

I want to avoid spitting content to stdout as much as possible. I made lots of these mistakes in Sedona source code but want to clean it up at some point.

james-willis: This test is to make sure that the function can be resolved when using the dataframe API in python so rewriting it into a scala test is not helpful.

Maybe I can use [sameSemantics](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.sameSemantics.html#pyspark.sql.DataFrame.sameSemantics) to accomplish this:

```python 
df.select(ExpandAddress("address").alias("normalized")).sameSemantics(
  df.select(f.expr("ExpandAddress(address)").alias("normalized"))
)
```

jiayuasu: let's try

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala
```diff
@@ -22,19 +22,23 @@ import org.apache.sedona.common.geometryObjects.Geography
...
 import org.apache.spark.sql.sedona_sql.expressions.implicits._
 import org.apache.spark.sql.types._
 import org.locationtech.jts.algorithm.MinimumBoundingCircle
 import org.locationtech.jts.geom._
 import org.apache.spark.sql.sedona_sql.expressions.InferrableFunctionConverter._
+import org.apache.spark.sql.sedona_sql.expressions.LibPostalUtils.{getExpanderFromConf, getParserFromConf}
 import org.apache.spark.unsafe.types.UTF8String
 import org.apache.spark.util.Utils
+import com.mapzen.jpostal.{AddressExpander, AddressParser}
+import org.apache.spark.sql.catalyst.expressions.codegen.Block.BlockHelper
```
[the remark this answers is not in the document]
james-willis: It is needed for the code gen in the libpostal functions

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Functions.scala
```diff
@@ -1831,3 +1835,140 @@ case class ST_InterpolatePoint(inputExpressions: Seq[Expression])
...
+
+          if (!${ev.isNull}) {
+            String[] expandedAddressArray = $expanderRef.expandAddress($address.toString());
+            $utf8StringClass[] utf8Strings = new $utf8StringClass[expandedAddressArray.length];
+            for (int j = 0; j < expandedAddressArray.length; j++) {
+              utf8Strings[j] = $utf8StringClass.fromString(expandedAddressArray[j]);
+            }
+            ${ev.value} = new $arrayDataClass(utf8Strings);
+          } else {
+            ${ev.value} = new $arrayDataClass(new $utf8StringClass[0]);
```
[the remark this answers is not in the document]
james-willis: practically this isnt true because of `boolean ${ev.isNull} = ${addressRef.isNull};`. I will update the code anyway

jiayuasu (approved):

PR #2086, opened 2025-07-10
[GH-2049] Retain index information in query results

### LINKED ISSUE #2049: Geopandas.GeoSeries: Implement proper index behavior for _row_wise_operations
After we get index support (#2036), revisit the `align` behavior for the row-wise operations, discussed in this PR originally https://github.com/apache/sedona/pull/2038.

Specifically we should investigate the comments copilot made. The suggested fixes weren't right, but we may be missing something.
- Set the index properly. It seems to be a union.
```python
geoseries = gpd.GeoSeries([Polygon([(0, 0), (1, 0), (1, 1)]), Polygon([(0, 0), (1, 0), (1, 1)]), Polygon([(0, 0), (1, 0), (1, 1)])], index=[1, 2, 3])
geoseries2 = gpd.GeoSeries([Polygon([(0, 0), (1, 0), (1, 1)]), Polygon([(0, 0), (1, 0), (1, 1)]), Polygon([(0, 0), (1, 0), (1, 1)])], index=[0, 1, 2])
res = geoseries.intersection(geoseries2, align=True)
print(res)
print(res.index)

0                              None
1    POLYGON ((0 0, 1 1, 1 0, 0 0))
2    POLYGON ((0 0, 1 1, 1 0, 0 0))
3                              None
dtype: geometry
Index([0, 1, 2, 3], dtype='int64')
```

- Investigate: Is it really an `outer` join really correct in all cases? Or could it be something more along the lines of a union of left and right?

petern48:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2049

- Retain index information in queries (`SPARK_DEFAULT_INDEX_NAME`)
- Select `NATURAL_ORDER_COLUMN_NAME` to prevent having to regenerate it (slight perf improvement)

## How was this patch tested?
Added tests

- No, no need. This was fixing buggy behavior.

### DIFF
--- a/python/sedona/geopandas/geoseries.py
+++ b/python/sedona/geopandas/geoseries.py
@@ -225,8 +225,9 @@ class GeoSeries(GeoFrame, pspd.Series):
         """
         from pyproj import CRS
 
-        tmp_df = self._process_geometry_column("ST_SRID", rename="crs")
-        srid = tmp_df.take([0])[0]
+        tmp = self._process_geometry_column("ST_SRID", rename="crs")
+        ps_series = tmp.take([0])
+        srid = ps_series.iloc[0]
         # Sedona returns 0 if doesn't exist
         return CRS.from_user_input(srid) if srid != 0 and not pd.isna(srid) else None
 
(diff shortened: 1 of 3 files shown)


python/sedona/geopandas/geoseries.py
```diff
@@ -1322,10 +1328,15 @@ def _row_wise_operation(
         # TODO: this does not yet support multi-index
```
zhangfengcdt: Why do we need to support multiple index in geoseries?

petern48: Traditional pandas series can have a [multi-index](https://pandas.pydata.org/docs/user_guide/advanced.html#creating-a-multiindex-hierarchical-index-object). This would be a task that would be far in the backlog though. I mainly just wanted to document that we would need to update the logic here if/when multi-index support is added. I've updated the comment to the following:
```
        # This code assumes there is only one index (SPARK_DEFAULT_INDEX_NAME)
        # and would need to be updated if Sedona later supports multi-index
```

zhangfengcdt: Got it, I think it means "composite" or multi-level index in pandas. The underlying pyspark should already handle this for non-spatial column, but we won't support this on spatial index I believe. 

LGTM

zhangfengcdt (approved):

python/sedona/geopandas/geoseries.py
petern48: Yeah, all of the index work I've been doing lately refers to traditional pandas indexes not geopandas sindex

jiayuasu:
@petern48 please resolve the conflicts

petern48:
@jiayuasu resolved and tests pass

Issue #2083, opened 2025-07-09
GeoPandas.SpatialIndex - Implement intersection Method


zhangfengcdt:
This task is to implement the intersection() method in GeoPandas.SpatialIndex class.

https://geopandas.org/en/stable/docs/reference/api/geopandas.sindex.SpatialIndex.intersection.html#geopandas.sindex.SpatialIndex.intersection

PR #2088, opened 2025-07-10
[GH-2083] Implement intersection Method for GeoPandas.SpatialIndex

zhangfengcdt:

This PR implements intersection Method for GeoPandas.SpatialIndex.

## How was this patch tested?
TestSpatialIndex

### DIFF
--- a/python/sedona/geopandas/sindex.py
+++ b/python/sedona/geopandas/sindex.py
@@ -196,7 +196,38 @@ class SpatialIndex:
         list
             List of indices of matching geometries.
         """
-        raise NotImplementedError("This method is not implemented yet.")
+        log_advice(
+            "`intersection` returns local list of indices of matching geometries onto driver's memory. "
+            "It should only be used if the resulting collection is expected to be small."
+        )
+
+        if self.is_empty:
+            return []
+
+        # Create a polygon from the bounds
+        from shapely.geometry import box
+
+        bbox = box(*bounds)
+
+        if self._is_spark:
+            # For Spark-based spatial index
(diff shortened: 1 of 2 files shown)


python/sedona/geopandas/sindex.py
```diff
@@ -196,7 +196,38 @@ def intersection(self, bounds):
...
+        if self._is_spark:
+            # For Spark-based spatial index
+            from sedona.spark.core.spatialOperator import RangeQuery
+
+            # Execute the spatial range query with the bounding box
+            result_rdd = RangeQuery.SpatialRangeQuery(
+                self._indexed_rdd, bbox, True, True
+            )
+
+            results = result_rdd.collect()
```
jiayuasu, line 221: Why do we use `collect` here? This will crash the driver memory

zhangfengcdt: The query(), intersection(), and nearest() all returns list in geopandas. This is different than the operators on GeoSeries / GeoDataframe for which it could return a new GeoSeries / GeoDataframe and chain another operators. 

The sindex is basically an efficient way to "search" the spatial data instead of transform the data. So mostly of the cases, it should return to driver for the end users. 

I might be misunderstanding the use of sindex though.

zhangfengcdt: I do have a warning in each of these method similar to the to_genpandas() method in GeoSeries / GeoDataframe when users want to collect the results back to geopandas df.

jiayuasu: make sense!

jiayuasu (approved):

PR #2096, opened 2025-07-15
[GH-2095] Implement scalable workaround for GeoSeries.__init__()

### LINKED ISSUE #2095: Geopandas.GeoSeries: Implement scalable workaround for GeoSeries.__init__()

petern48:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2095

In a previous PR, I wrote a workaround for the behavior of creating a Series from a Series. That workaround required the use of `to_pandas()` which is not scalable. I previously submitting a fix to Spark was told this would be a new feature that could not be backported. I then realized, I could just manually implement the logic in our codebase in the mean time since it's all in the constructor.

## How was this patch tested?
Ensured existing tests pass

### DIFF
--- a/python/sedona/geopandas/geoseries.py
+++ b/python/sedona/geopandas/geoseries.py
@@ -155,9 +155,9 @@ class GeoSeries(GeoFrame, pspd.Series):
 
             # Can't to try-except here since the exception would be in spark, so we check the first valid index
             first_idx = data.first_valid_index()
-            if isinstance(
+            if first_idx is None or isinstance(
                 data[first_idx], (BaseGeometry, bytearray, bytes)
-            ) and not pd.isna(data[first_idx]):
+            ):
                 data = data.apply(try_geom_to_ewkb)
             else:
                 raise TypeError(
--- a/python/sedona/geopandas/geoseries.py
+++ b/python/sedona/geopandas/geoseries.py
@@ -186,6 +186,8 @@ class GeoSeries(GeoFrame, pspd.Series):
                 copy=copy,
                 fastpath=fastpath,
             )
+
+            self._anchor = data
         else:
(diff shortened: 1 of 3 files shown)


petern48:
This PR makes the code lose the CRS info again. It's because of the way Sedona serializes the shapely Geometry objects. Unfortunately, it doesn't include the CRS info, so I have to look into the best way to address this.

petern48:
I think what's happening:
When I call `.apply()` to convert geometries into EWKB, Spark applies my specified function, but then calls `GeometryType()`'s `serialize` and `deserialize` function on my input thinking they're still Geometry objects, which leads to messed up bytes. Looks like I need to find a different workaround.

petern48:
@jiayuasu Do you happen to know off the top of your head where Sedona stores it's SRID/CRS info (e.g in a separate column, metadata, in the serialization format)? It's got to be somewhere, since it can be accessed by worker nodes.

EDIT: I've dug into the C code and see that it's serialized in the format. Still not sure why it's not retained once it becomes a shapely object for me.

jiayuasu:
@Kontinuation did you have answers for Peter's question?

petern48:
I figured it all out after quite a bit of digging. SRID info is [included in the serialization process](https://github.com/apache/sedona/blob/8d4fb1bbbafde8db01c9d845ba4d4261636c4681/python/src/geom_buf.c#L224-L225) per geometry object. Just a bunch of various bugs in Sedona and Spark covering things up.

Closing. New working PR is here: https://github.com/apache/sedona/pull/2121

Kontinuation:
I'm submitting a PR https://github.com/apache/sedona/pull/2123 to resolve https://github.com/apache/sedona/issues/2122. Hopefully it will resolve the srid preservation issue.

PR #2098, opened 2025-07-15
[GH-2085] use graphframes 0.9.0

### LINKED ISSUE #2085: Bump graphframes version to 0.9.0
0.9.0 is in SNAPSHOT release. we need to test this new version in sedona and make any changes to maintain compatability. This update should fix the heap space issue we are having with DBSCAN and give spark 4 support.

james-willis:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2085 

- Bump graphframes version to 0.9.0
- deal with changes in how graphframes assigns connected component IDs
- fix Geostats tests that have always been broken

## How was this patch tested?
unit tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -85,7 +85,10 @@
         <spark.compat.version>3.4</spark.compat.version>
         <spark.major.version>3</spark.major.version>
         <log4j.version>2.19.0</log4j.version>
-        <graphframe.version>0.8.3-spark3.4</graphframe.version>
+
+        <graphframes.groupId>graphframes</graphframes.groupId>
+        <graphframes.artifactId>graphframes</graphframes.artifactId>
+        <graphframes.version>0.8.3-spark${spark.compat.version}-s_${scala.compat.version}</graphframes.version>
 
         <flink.version>1.19.0</flink.version>
         <slf4j.version>1.7.36</slf4j.version>
@@ -396,6 +399,18 @@
                 <enabled>true</enabled>
             </releases>
         </repository>
+        <!-- TODO remove when graphframes 0.9.0 is released-->
+        <repository>
+            <name>Central Portal Snapshots</name>
+            <id>central-portal-snapshots</id>
+            <url>https://central.sonatype.com/repository/maven-snapshots/</url>
+            <releases>
+                <enabled>false</enabled>
+            </releases>
+            <snapshots>
+                <enabled>true</enabled>
+            </snapshots>
+        </repository>
         <repository>
             <id>osgeo</id>
             <name>OSGeo Release Repository</name>
@@ -607,6 +622,7 @@
                         <spark.version>${spark.version}</spark.version>
                         <scala.version>${scala.version}</scala.version>
                         <log4j.version>${log4j.version}</log4j.version>
+                        <!--TODO is this the right value? what is this for?-->
                         <graphframe.version>${graphframe.version}</graphframe.version>
                     </properties>
                 </configuration>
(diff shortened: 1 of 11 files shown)


james-willis:
We should wait until 0.9.0 is release and we can point to the non-snapshot version before merging this.

If we disagree with the change in behavior, this PR gives us a chance to discuss before 0.9.0 is released.

Kimahriman:
You don't just want to set the config to maintain the old behavior? It seems odd that the output of the dbscan function is based on a config a user could set for a transitive dependency they don't even know is used for dbscan

james-willis:
I really dislike the way the spark configs are global configurations and so modifying a spark config changes the state of the whole cluster.

I'm thinking of adding a setter to graphframes 0.9.0 so I can set this value on the connectedComponents instance without changing spark configs.

Then the output schema of DBSCAN is not based on the user's setting of the graphframe spark config (one way or the other).

Kimahriman:
Yeah it's definitely not the cleanest thing, and I was thinking the same thing a setting on the connected components object would be much cleaner

pom.xml
```diff
@@ -85,7 +85,12 @@
         <spark.compat.version>3.4</spark.compat.version>
         <spark.major.version>3</spark.major.version>
         <log4j.version>2.19.0</log4j.version>
-        <graphframe.version>0.8.3-spark3.4</graphframe.version>
+
+
+        <graphframes.groupId>io.graphframes</graphframes.groupId>
+        <graphframes.artifactId>graphframes-spark${spark.major.version}_${scala.compat.version}</graphframes.artifactId>
```
Kimahriman: Probably don't need these as variables anymore if we're using the same thing for all Spark versions now, can just inline this

pom.xml
```diff
@@ -85,7 +85,12 @@
...
+        <!--There is no 3.4 version of GraphFrames, so we use 0.9.0-spark3.5-->
+        <graphframes.version>0.9.0-spark3.5</graphframes.version>
```
Kimahriman: Versioning was fixed so this can just be 
```suggestion
        <graphframes.version>0.9.2</graphframes.version>
```

pom.xml
```diff
@@ -752,9 +755,11 @@
                 <hadoop.version>3.4.1</hadoop.version>
                 <log4j.version>2.24.3</log4j.version>
                 <slf4j.version>2.0.16</slf4j.version>
-                <graphframe.version>0.8.3-spark3.5</graphframe.version>
+
                 <scala.version>2.13.12</scala.version>
                 <scala.compat.version>2.13</scala.compat.version>
+                <graphframes.version>0.9.0-spark${spark.compat.version}</graphframes.version>
```
Kimahriman: And then this can be removed

Kimahriman (commented):
Would be great to get this in and a 1.8.0 release with Spark 4 support!

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/GeoStatsFunctions.scala
```diff
@@ -31,8 +31,10 @@ import org.apache.spark.sql.types._
 
 case class ST_DBSCAN(children: Seq[Expression]) extends DataframePhysicalFunction {
 
-  override def dataType: DataType = StructType(
-    Seq(StructField("isCore", BooleanType), StructField("cluster", LongType)))
+  lazy val conf = SedonaConf.fromActiveSession() // only access on driver
```
Kimahriman: Don't think this is used anymore?

Kimahriman (approved):

pom.xml
```diff
@@ -85,7 +85,8 @@
         <spark.compat.version>3.4</spark.compat.version>
         <spark.major.version>3</spark.major.version>
         <log4j.version>2.19.0</log4j.version>
-        <graphframe.version>0.8.3-spark3.4</graphframe.version>
+
```
jiayuasu: nitpick: please remove these unnecessary blank lines

PR #2123, opened 2025-07-19
[GH-2122] Fix deserialization of geometry SRID in python

### LINKED ISSUE #2122: Reading SRID info from serialized geometries using shapely.get_srid is wrong for srid >= 3968
```python
df = spark.sql("""
SELECT
    ST_SetSRID(ST_GeomFromText('POINT (0 0)'), 3857) as geom1,
    ST_SetSRID(ST_GeomFromText('POINT (0 0)'), 3967) as geom2,
    ST_SetSRID(ST_GeomFromText('POINT (0 0)'), 3968) as geom3,
    ST_SetSRID(ST_GeomFromText('POINT (0 0)'), 3969) as geom4
""")
col = df.collect()[0]
print(shapely.get_srid(col[0]))  # 3857 (correct)
print(shapely.get_srid(col[1]))  # 3967 (correct)
print(shapely.get_srid(col[2]))  # -128 (should be 3968)
print(shapely.get_srid(col[3]))  # -127 (should be 3969)
```

The boundary is 3968, which is 10000000. It's being interpreted as -128 instead. Seems like it's being interpreted as a signed 8-bit int instead of an unsigned one or a larger int.

Kontinuation:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2122 

SRID was not deserialized correctly in the Python extension. This PR fix it.

The new geometry serialization format was introduced before we support Shapely 2.0, Shapely 1.8 does not support SRID back then so SRID serialization in Python extension was only implemented but not tested.

**ATTENTION**: This patch only fixes the C extension (fast implementation), there's a slower fallback implementation in pure Python (https://github.com/apache/sedona/blob/master/python/sedona/spark/utils/geometry_serde_general.py) not supporting serialization and deserialization of SRID at all. We need to fix that in a follow-up PR.

## How was this patch tested?

Added new unit test

### DIFF
--- a/python/tests/utils/test_geomserde_speedup.py
+++ b/python/tests/utils/test_geomserde_speedup.py
@@ -15,6 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
+import pytest
 import shapely
 from shapely.geometry import (
     GeometryCollection,
@@ -136,6 +137,9 @@ class TestGeomSerdeSpeedup:
         ]
         self._test_serde_roundtrip(geometry_collections)
 
+    @pytest.mark.skipif(
+        shapely.__version__ < "2", reason="SRID functions require Shapely >= 2.0"
+    )
     def test_srid_roundtrip(self):
         point = wkt_loads("POINT (1 2)")
         point = shapely.set_srid(point, 1000)
--- a/python/tests/utils/test_geomserde_speedup.py
+++ b/python/tests/utils/test_geomserde_speedup.py
@@ -15,6 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
+import shapely
 from shapely.geometry import (
     GeometryCollection,
     LineString,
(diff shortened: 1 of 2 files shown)


python/tests/utils/test_geomserde_speedup.py
```diff
@@ -135,6 +137,15 @@ def test_geometry_collection(self):
         ]
         self._test_serde_roundtrip(geometry_collections)
 
+    @pytest.mark.skipif(
+        shapely.__version__ < "2", reason="SRID functions require Shapely >= 2.0"
+    )
```
Kontinuation, line 142: The shapely requirement of Sedona is >= 1.7.0. I found that the newly added geopandas support depends on features only in Shapely 2, this calls out for the removal of Shapely 1.x support.

Kontinuation: See also https://github.com/apache/sedona/commit/aaf3fb866c2ef0dd72a9c9a25330d15a77a206e8

jiayuasu:
I think we should remove Shapely 1.X support in the upcoming 1.8.X release.

PR #2150, opened 2025-07-24
[GH-2149] Geopandas: Implement `to_file`, `from_file`, `read_file`

### LINKED ISSUE #2149: Geopandas: Implement `to_file`, `from_file`, `read_file`

petern48:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2149

- Implement `to_file`, `from_file`, `read_file`
- Fix GeoDataFrame.getitem to handle non-int indexes properly

## How was this patch tested?
Added tests

### DIFF
--- a/python/sedona/geopandas/geodataframe.py
+++ b/python/sedona/geopandas/geodataframe.py
@@ -32,6 +32,7 @@ from pyspark.pandas import Series as PandasOnSparkSeries
 from pyspark.pandas._typing import Dtype
 from pyspark.pandas.frame import DataFrame as PandasOnSparkDataFrame
 from pyspark.pandas.internal import InternalFrame
+from pyspark.pandas.utils import log_advice
 
 from sedona.geopandas._typing import Label
 from sedona.geopandas.base import GeoFrame
@@ -351,10 +352,10 @@ class GeoDataFrame(GeoFrame, pspd.DataFrame):
 
             try:
                 result = sgpd.GeoSeries(ps_series)
-                first_idx = ps_series.first_valid_index()
-                if first_idx is not None:
-                    geom = ps_series.iloc[int(first_idx)]
-                    srid = shapely.get_srid(geom)
+                not_null = ps_series[ps_series.notnull()]
+                first_geom = not_null.iloc[0]
+                if first_geom:
+                    srid = shapely.get_srid(first_geom)
 
                     # Shapely objects stored in the ps.Series retain their srid
                     # but the GeoSeries does not, so we manually re-set it here
@@ -425,7 +426,7 @@ class GeoDataFrame(GeoFrame, pspd.DataFrame):
             # instead of calling e.g assert not dtype ourselves.
             # This way, if Spark adds support later, than we inherit those changes naturally
             super().__init__(data, index=index, columns=columns, dtype=dtype, copy=copy)
-        elif isinstance(data, PandasOnSparkDataFrame):
+        elif isinstance(data, (PandasOnSparkDataFrame, SparkDataFrame)):
 
             super().__init__(data, index=index, columns=columns, dtype=dtype, copy=copy)
         elif isinstance(data, PandasOnSparkSeries):
@@ -436,14 +437,6 @@ class GeoDataFrame(GeoFrame, pspd.DataFrame):
                 pass
 
             super().__init__(data, index=index, columns=columns, dtype=dtype, copy=copy)
-        elif isinstance(data, SparkDataFrame):
-            assert columns is None
-            assert dtype is None
-            assert not copy
-
-            if index is None:
-                internal = InternalFrame(spark_frame=data, index_spark_columns=None)
-                object.__setattr__(self, "_internal_frame", internal)
         else:
             # below are not distributed dataframe types
             if isinstance(data, pd.DataFrame):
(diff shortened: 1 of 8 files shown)


python/sedona/geopandas/geoseries.py
```diff
@@ -4718,7 +4796,8 @@ def _to_geoframe(self, name=None):
             renamed = self.rename("geometry")
         else:
             renamed = self
-        return GeoDataFrame(pspd.DataFrame(renamed._internal))
+
+        return GeoDataFrame(pspd.DataFrame(renamed._internal).to_spark())
```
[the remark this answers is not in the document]
petern48: No, I've added a comment for why this method call is important. Otherwise, some tests will fail.

python/sedona/geopandas/geodataframe.py
```diff
@@ -512,6 +508,9 @@ def __init__(
     # ============================================================================
     # GEOMETRY COLUMN MANAGEMENT
     # ============================================================================
+    def __repr__(self):
+        geoseries = self.to_geopandas()
+        return geoseries.__repr__()
```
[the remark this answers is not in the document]
petern48: I'd rather print the actual dataframe.

EDIT: Actually this brought up a good point. We should be able to print the representation without calling `to_geopandas()`. Now that we've moved towards storing everything as geometries and not WKB, the natural spark output displays correctly (instead of byte arrays like how we started). I've removed the `__repr__()`s, so we simply use pyspark pandas implementation, which now displays nicely.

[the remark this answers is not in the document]
zhangfengcdt: Cool, I think this is the correct approach.

python/sedona/geopandas/io.py
```diff
@@ -0,0 +1,242 @@
...
+        spark_df = spark_df.drop(SPARK_DEFAULT_INDEX_NAME)
+
+    if spark_fmt == "geoparquet":
+        writer = spark_df.write.format("geoparquet")
+
+        # if not saving index we sort by GeoHash to optimize reading
+        if not index and df.active_geometry_name:
+            from sedona.spark import ST_GeoHash
+
+            spark_df = spark_df.orderBy(ST_GeoHash(df.geometry.spark.column, 5))
```
zhangfengcdt: why do we need to hard-code 5 here?  consider making this configurable from the argument?

python/sedona/geopandas/geodataframe.py
```diff
@@ -1552,7 +1484,7 @@ def buffer(
         mitre_limit=5.0,
         single_sided=False,
         **kwargs,
-    ) -> GeoDataFrame:
+    ) -> sgpd.GeoSeries:
```
zhangfengcdt, line 1487: This does not seem to be consistent with the comment below: Returns a GeoDataFrame with all ...

python/sedona/geopandas/io.py
```diff
@@ -0,0 +1,242 @@
...
+
+    spark = default_session()
+
+    # If format is not specified, infer it from the file extension
+    if format is None:
+        if os.path.isdir(filename):
+            raise ValueError(
+                f"Inferring the format from the file extension is not supported for directories: {filename}"
+            )
+        if filename.endswith(".shp"):
```
zhangfengcdt: Can we make these case-insensitive matching?

python/sedona/geopandas/io.py
petern48: There's no default value, so i need to specify something. I'm open to other alternative suggestions, but I just picked the value used in the docs to be honest.
https://sedona.apache.org/1.5.0/api/sql/Optimizer/?h=st_geohash#push-spatial-predicates-to-geoparquet

python/sedona/geopandas/geodataframe.py
petern48: I've fixed the doc here to say GeoSeries, but in reality we should remove this implementation does not belong in geodataframe. I just needed to fix things to make sure tests still pass. This will be resolved in the upcoming PR to make all geoseries functions callable from geodataframe.

python/sedona/geopandas/io.py
petern48: Done, also make the format parameter case-insensitive.

python/sedona/geopandas/io.py
zhangfengcdt: @jiayuasu Do you happen to know the best default number for precision in ST_Geohash function? Do we need to pass it as an extra parameter for creating the index? For default geo references, 

 - Precision 5-7 is typically optimal for most geospatial datasets
  - Precision 5 (current): ~5km × 5km cells - good for country/regional datasets
  - Precision 6: ~1.2km × 0.6km cells - better for city-level datasets
  - Precision 7: ~150m × 150m cells - good for high-resolution datasets

petern48: We can also just drop this logic if it doesn't make sense to set a default. I just added it as a nice to have.

zhangfengcdt: Ok, that's remove this logic and we are good then. We can add this later when it is clear. Thanks!

zhangfengcdt (approved):

jiayuasu (approved):

PR #2167, opened 2025-07-25
[GH-2166] More pathing fixes for Python package refactor

### LINKED ISSUE #2166: Fix Breaking Changes from #1930
lots of pieces of the python interface were broken by #1930 

Lets do our best to put the interface back together along with adding deprecation warnings

james-willis:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2166

I'm fixing the breaking changes from #1930

## How was this patch tested?
unit tests


python/tests/test_path_compatibility_all.py
```diff
@@ -19,7 +19,9 @@
 # We will drop this test file in the future when we remove path compatibility for Apache Sedona < 1.8.0
 
 from sedona.spark import *
+from sedona.spark.maps.SedonaMapUtils import SedonaMapUtils
```
jiayuasu: The whole point of `test_path_compatibilty_all.py` is to test `from sedona.spark import *`. All other tests should remain the same as `test_path_compatibilty.py`

python/tests/test_path_compatibility.py
```diff
@@ -118,3 +121,30 @@ def test_sql_module_imports(self):
...
+    def test_raster_utils_imports(self):
+        # Test raster utils imports
+        assert SedonaUtils is not None
+
+    def test_import_df_functions_from_sedona_sql(self):
+        # one from each module
+        from sedona.sql import ST_MakePoint, ST_Y, ST_Touches, ST_Envelope_Aggr
+
+    def test_geoarrow_imports(self):
+        from sedona.geoarrow import create_spatial_dataframe, dataframe_to_arrow
```
jiayuasu: Why test this again since we already have them tested above?

python/tests/test_path_compatibility.py
```diff
@@ -118,3 +121,30 @@ def test_sql_module_imports(self):
...
+
+    def test_sedona_util_imports(self):
+        from sedona.utils import KryoSerializer
+        from sedona.utils import SedonaKryoRegistrator
+
+    def test_maps_imports(self):
+        # Test Map imports
+        from sedona.maps import SedonaKepler, SedonaMapUtils, SedonaPyDeck
```
jiayuasu: Please use the `is not None` way to test. So you can copy the same code to the other file.

james-willis: I think this testing structure is better.

With those the imports are global so if there is an import error no tests will run. With the imports scoped locally like this the tests encapsulate what they are meant to test; the import.

The `is not None` seems redundant to me; if you were able to import it, isnt it almost certainly `not None`?

jiayuasu: I agree. However, the reason why I didn't do it is because I want to copy the same test code to the other file `test_path_compatibilty_all.py`. 

`from sedona.spark import *` has to be placed at the file header level so I cannot add it to individual test case.

james-willis: I migrated the whole file to this scoped imports style. Are you ok with that or do you want to convert it all to the same style as the `test_path_compatability_all.py`file?

jiayuasu: I would prefer if you can convert it back to the same style as the `test_path_compatability_all.py`

python/sedona/sql/__init__.py
```diff
@@ -14,3 +14,11 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+
+# These allow use to access the __all__
+from sedona.spark.sql import *
+from sedona.spark import sql
+
+print("xxxxxxxxxxxx")
```
jiayuasu: Please remove these?

james-willis: sorry

python/sedona/sql/__init__.py
```diff
@@ -14,3 +14,11 @@
...
+print(sql.__all__)
```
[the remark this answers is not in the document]
james-willis: fixed

PR #2177, opened 2025-07-28
[GH-2716] Always use native byte order when writing WKBs

### LINKED ISSUE #2176: GeoParquet: Always write WKB in native byte order
GeoParquet does not configure the byte order of JTS's WKBWriter explicitly when writing WKBs. The default byte order is big-endian. This introduces the overhead of byte-swapping when parsing such WKBs on today's most commonly used processors. The overhead does not exhibit when using JTS's WKBReader, but may occur when using other libraries focusing on performance.

We should write WKBs in the native byte order of the system. This will produce little-endian WKBs for most of the time, and the performance of decoding WKBs might be better when decoding on the same system, and even when transmitted to other most commonly used systems.

Kontinuation:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2176 

* Refactored code to always use `GeomUtils.createWKBWriter`, which returns a WKB writer using the native byte order.
* Removed redundant WKB serde in some snowflake UDFs.

## How was this patch tested?

* Passing existing tests
* Add a new test to verify that the GeoParquet files written by sedona has native byte order
* Snowflake integration test passes: https://github.com/wherobots/sedona-snowflake-tester/pull/56

- No, this PR does not affect any public API so no need to change the documentation. However, this change might be breaking, so we'd better add this to the release note of 1.8.0.

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/utils/GeomUtils.java
+++ b/common/src/main/java/org/apache/sedona/common/utils/GeomUtils.java
@@ -195,12 +195,7 @@ public class GeomUtils {
     if (geometry == null) {
       return null;
     }
-    int endian =
-        ByteOrder.nativeOrder() == ByteOrder.BIG_ENDIAN
-            ? ByteOrderValues.BIG_ENDIAN
-            : ByteOrderValues.LITTLE_ENDIAN;
-    WKBWriter writer =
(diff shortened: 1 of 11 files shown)


jiayuasu:
Note to ourselves. this might be a breaking change, especially for Overture users who use Sedona to generate their data.

PR #2178, opened 2025-07-28
[CI] Clean up Python Bandit skips

jbampton:

- No this is a CI update. The PR name follows the format `[CI] my subject`

Improves Bandit coverage by reducing the global skips and then using inline `nosec` skips

## How was this patch tested?

`pre-commit run --all-files`

### DIFF
--- a/python/tests/sql/test_predicate.py
+++ b/python/tests/sql/test_predicate.py
@@ -174,7 +174,7 @@ class TestPredicate(TestBase):
 
         string = "100.01,200.01,100.5,200.01,100.5,200.5,100.01,200.5,100.01,200.01"
         equal_df = self.spark.sql(
-            f"select * from polygonDf where ST_Equals(polygonDf.polygonshape, ST_LineStringFromText('{string}', ',')) "  # nosec B608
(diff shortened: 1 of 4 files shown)


python/tests/sql/test_predicate.py
```diff
@@ -174,7 +174,7 @@ def test_st_equals_for_st_linestring_and_st_polygon(self):
 
         string = "100.01,200.01,100.5,200.01,100.5,200.5,100.01,200.5,100.01,200.01"
         equal_df = self.spark.sql(
-            f"select * from polygonDf where ST_Equals(polygonDf.polygonshape, ST_LineStringFromText('{string}', ',')) "
+            f"select * from polygonDf where ST_Equals(polygonDf.polygonshape, ST_LineStringFromText('{string}', ',')) "  # nosec B608
```
Kontinuation: Can we skip checking source files in python/tests directory? Tests are not exposed as attack surface so I believe that it is reasonable to skip checking test sources.

jiayuasu:
Not sure if it is a good idea to check those sql sentences. They are commonly used in our test cases and seem to have no much downside

jiayuasu (approved):

PR #2184, opened 2025-07-28
[GH-2183] Fix ST_Buffer logic cap_style when specifying 'side' parameter + Implement rest of geoseries.buffer()

### LINKED ISSUE #2183: ST_Buffer incorrectly sets the cap_style when specifying 'side' parameter

petern48:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2183

This PR fixes the following issues in ST_Buffer:
- We previously were defaulting to cap_style square when `side=both` was set, but we should be doing it when 'side=left' or 'side=right' is set.
- Setting this value was not just a default, but a potential override.

Also implemented the rest of geoseries.buffer()

## How was this patch tested?
Added tests

- No, this PR does not affect any public API so no need to change the documentation. (Fixes a bug so the code behaves as described in the docs)


common/src/main/java/org/apache/sedona/common/Functions.java
```diff
@@ -386,12 +388,16 @@ else if (singleParam[0].equalsIgnoreCase(listBufferParameters[3])
...
+          // Default value is 'both'
           continue;
         } else if (singleParam[1].equalsIgnoreCase(sideOptions[1])
             || singleParam[1].equalsIgnoreCase(sideOptions[2])) {
           bufferParameters.setSingleSided(true);
+
+          // Specifying 'left' or 'right' defaults to square end cap style when side is specified
+          if (!endCapSpecified) {
+            bufferParameters.setEndCapStyle(BufferParameters.CAP_SQUARE);
+          }
```
furqaankhan, line 400: Good catch! Can you add Java tests for this change? And update the docs as well for this behavior change.

furqaankhan (changes requested):

common/src/main/java/org/apache/sedona/common/Functions.java
petern48: Added tests and updated the docs toe be more clear @furqaankhan

jiayuasu (approved):

petern48:
PostGIS (finally) accepted my PR to update their docs to clarify this behavior as well https://github.com/postgis/postgis/pull/824 🙌

PR #2188, opened 2025-07-29
[GH-1643] Add `mvn spotless:apply` as a pre-commit hook

### LINKED ISSUE #1643: Run `mvn spotless:apply` with `pre-commit` so that it runs on `git commit`
As an extra check it would be good to add `mvn spotless:apply` to pre-commit so that it runs on `git commit`

Seems both Java and Scala are not listed as supported languages for pre-commit but they do have:

- docker
- docker_image
- script
- system

"System hooks provide a way to write hooks for system-level executables which don't have a supported language above (or have special environment requirements that don't allow them to run in isolation such as pylint).

This hook type will not be given a virtual environment to work with – if it needs additional dependencies the consumer must install them manually.

Support: the support of system hooks depend on the executables."

https://pre-commit.com/#supported-languages

I found some issues about pre-commit and spotless:

https://github.com/diffplug/spotless/issues/623

https://github.com/diffplug/spotless/issues/178

jbampton:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #1643 

Added our first "Java" based hook to run `mvn spotless:apply`

## How was this patch tested?

Tested with Java 11.

Tested first with `pre-commit run --all-files` and also then with `git commit`

Tried failing the hook with messy Java and saw the correct reformat and the hook failed.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -40,6 +40,19 @@ repos:
         files: \.(js|ya?ml)$
         language: node
         additional_dependencies: ['prettier@3.6.2']
+      - id: maven-spotless-apply
+        name: maven spotless apply
+        description: automatically formats Java and Scala code using mvn spotless:apply.
+        entry: mvn spotless:apply
+        language: system # Indicates that the 'entry' command should be run directly as a system command.
+        types: [java, scala] # Specifies that this hook should run on Java and Scala files.
+        pass_filenames:
+          false # Crucial: tells pre-commit NOT to pass filenames as arguments to 'mvn spotless:apply'.
+          # Spotless typically scans the whole project based on its configuration.
+        always_run:
+          true # Ensures this hook runs even if no Java files are changed.
+          # This is useful for spotless:apply which might affect files not staged.
+        stages: [pre-commit] # Specifies that this hook runs during the 'commit' stage.
       - id: check-zip-file-is-not-committed
         name: check no zip files are committed
         description: Zip files are not allowed in the repository


jbampton:
Seems that 4 versions of Java are already installed by default in the Ubuntu 24 image as seen below and that is why this "system" or Java hook has passed on GitHub Actions.  And a version of Maven is installed as well.  But when running this hook on your local machine you do need Java and most likely Java 11.  I did try running this hook at first with Java 21 and the hook failed with a Java code exception.  The result lead to using Java 11 successfully. 

https://github.com/actions/runner-images/blob/ubuntu24/20250720.1/images/ubuntu/Ubuntu2404-Readme.md#java

https://github.com/actions/runner-images/blob/ubuntu24/20250720.1/images/ubuntu/Ubuntu2404-Readme.md#project-management

Screenshot from 2025-07-29 16-29-13

Screenshot from 2025-07-29 16-27-28

jiayuasu (approved):

.pre-commit-config.yaml
```diff
@@ -40,6 +40,19 @@ repos:
...
+        description: automatically formats Java and Scala code using mvn spotless:apply.
+        entry: mvn spotless:apply
+        language: system # Indicates that the 'entry' command should be run directly as a system command.
+        types: [java, scala] # Specifies that this hook should run on Java and Scala files.
+        pass_filenames:
+          false # Crucial: tells pre-commit NOT to pass filenames as arguments to 'mvn spotless:apply'.
+          # Spotless typically scans the whole project based on its configuration.
+        always_run:
+          true # Ensures this hook runs even if no Java files are changed.
+          # This is useful for spotless:apply which might affect files not staged.
```
petern48, line 54: @jbampton Could you elaborate on what you meant by this? I'm getting some noticeable delays (up to ~2-3 seconds) on my commits now even when I'm purely working in Python and no Java.

petern48: Is it just to format java files even when they aren't staged?

jiayuasu: Yes. I think this is the goal but yeah, I forgot this will affect the python development. Do you think we should remove it?

petern48: It depends since most of the development is in Java (at least for now). I still don't quite see how this is particularly useful for Java devs. Does it really make a difference if `mvn spotless:apply` doesn't run on your unstaged java files? Won't it still run whenever you do finally decide to commit it? That's the only time it matters imo.

Issue #2186, opened 2025-07-28
[GH-2186] Integrate S2Geography into Sedona SQL


ZhuochengShang:
PR link: https://github.com/apache/sedona/pull/2210

PR #2210, opened 2025-07-31
[GH-2186]Integrate Geography UDT, UDF into Apache Sedona

ZhuochengShang:
- Geography UDT
- Geography UDF2:
- constructors point/polyline/polygon/collections from WKB or WKT
- distance functions 
- predicates functions with "equals", "contains", "intersects"

### DIFF
--- a/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Constructors.scala
+++ b/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/Constructors.scala
@@ -18,7 +18,7 @@
  */
 package org.apache.spark.sql.sedona_sql.expressions
 
-import org.apache.sedona.common.{Constructors, Functions}
+import org.apache.sedona.common.{Constructors, Functions, S2Constructors}
 import org.apache.sedona.common.enums.FileDataSplitter
 import org.apache.sedona.sql.utils.GeometrySerializer
 import org.apache.spark.sql.catalyst.InternalRow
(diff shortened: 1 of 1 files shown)


jiayuasu:
@ZhuochengShang @Kontinuation 

We should make sure the ST functions of Geography have the same name as the geometry ones. Do not add S2 in any of the names.

In addition, we probably should also have 2 additional functions: to_geography and to_geometry, try_to_geography, try_to_geometry. See Snowflake docs: https://docs.snowflake.com/en/sql-reference/functions/try_to_geography

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/UDT/GeographyUDT.scala
```diff
@@ -18,26 +18,26 @@
...
+import org.apache.sedona.common.geometrySerde.GeometrySerializer
 import org.apache.spark.sql.catalyst.util.{ArrayData, GenericArrayData}
 import org.apache.spark.sql.types._
 import org.json4s.JsonDSL._
 import org.json4s.JsonAST.JValue
-import org.apache.sedona.common.geometryObjects.Geography;
+import org.apache.sedona.common.S2Geography.{S2Geography, S2GeographySerializer}
 
-class GeographyUDT extends UserDefinedType[Geography] {
+class GeographyUDT extends UserDefinedType[S2Geography] {
```
jiayuasu, line 28: Sorry for being nitpicky here. But please do not rename this to S2Geography

PR #2221, opened 2025-08-04
[GH-2222] add Singlepoint, Singlepolyline for correctlty round trip WKT write

ZhuochengShang:
Add new Geography kind: SINGLEPOINT, SINGLEPOLYLINE
Solve the issue of read Geography from WKB/WKT as POINT/LINESTRING and decode into same WKT text

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/S2Geography/PointGeography.java
+++ b/common/src/main/java/org/apache/sedona/common/S2Geography/PointGeography.java
@@ -48,6 +48,11 @@ public class PointGeography extends S2Geography {
   /** Constructs especially for CELL_CENTER */
   PointGeography(GeographyKind kind, S2Point point) {
     super(kind); // can be POINT or CELL_CENTER
+    if (kind != GeographyKind.POINT
+        && kind != GeographyKind.SINGLEPOINT
+        && kind != GeographyKind.CELL_CENTER) {
+      throw new IllegalArgumentException("Invalid GeographyKind for PolylineGeography: " + kind);
+    }
     points.add(point);
   }
 
--- a/common/src/main/java/org/apache/sedona/common/S2Geography/PointGeography.java
+++ b/common/src/main/java/org/apache/sedona/common/S2Geography/PointGeography.java
@@ -51,7 +51,7 @@ public class PointGeography extends S2Geography {
     if (kind != GeographyKind.POINT
         && kind != GeographyKind.SINGLEPOINT
         && kind != GeographyKind.CELL_CENTER) {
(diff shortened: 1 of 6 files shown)


common/src/main/java/org/apache/sedona/common/S2Geography/SinglePointGeography.java
```diff
@@ -22,7 +22,7 @@
 
 public class SinglePointGeography extends PointGeography {
   public SinglePointGeography(S2Point p) {
-    super(p);
+    super(GeographyKind.SINGLEPOINT, p);
```
jiayuasu, line 25: What is the difference between PointGeography and SinglePointGeography? Is this something already defined in S2Geography C++?

OGC Simple Feature uses Point, MultiPoint, Polygon, MultiPolygon, ...

ZhuochengShang: PointGeography accepts a list of S2Point instances, while SinglePointGeography must contain exactly one S2Point. Here we added two new GeographyKind: SINGLEPOINT and SINGLEPOLYLINE. The reason is, previously, with only POINT defined, both WKT “POINT” and “MULTIPOINT” were serialized as “MULTIPOINT.” By introducing these more specific kinds, we can output correct WKT format string.

jiayuasu: OK. Does S2Geography C++ have similar issues or it is just for Java side?

ZhuochengShang: it is the Java side

common/src/main/java/org/apache/sedona/common/S2Geography/PointGeography.java
```diff
@@ -46,8 +46,13 @@ public PointGeography() {
   }
 
   /** Constructs especially for CELL_CENTER */
-  private PointGeography(GeographyKind kind, S2Point point) {
+  PointGeography(GeographyKind kind, S2Point point) {
     super(kind); // can be POINT or CELL_CENTER
+    if (kind != GeographyKind.POINT
+        && kind != GeographyKind.SINGLEPOINT
+        && kind != GeographyKind.CELL_CENTER) {
+      throw new IllegalArgumentException("Invalid GeographyKind for PolylineGeography: " + kind);
```
jiayuasu: Should this be `PointGeography`?

jiayuasu (approved):

PR #2254, opened 2025-08-08
[GH-2255]new Geography constructor from point, polyline, polygon

ZhuochengShang:
Modify GeogFromWKB so that it supports both input as String and Byte[]
Add new constructor functions :

ST_S2PointFromText
ST_S2LineFromText
ST_S2LinestringFromText
ST_S2PolyFromText

ST_S2MPointFromText
ST_S2MLineFromText
ST_S2MPolyFromText

ST_S2PointFromWKB
ST_S2LineFromWKT

ST_TryToGeography
ST_ToGeography

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/geography/Constructors.java
+++ b/common/src/main/java/org/apache/sedona/common/geography/Constructors.java
@@ -18,9 +18,7 @@
  */
 package org.apache.sedona.common.geography;
 
-import org.apache.sedona.common.S2Geography.Geography;
-import org.apache.sedona.common.S2Geography.WKBReader;
-import org.apache.sedona.common.S2Geography.WKTReader;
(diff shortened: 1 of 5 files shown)


jiayuasu:
I don't think we need constructors like ST_PointFromText because there is no way we can differentiate it from the geometry one.

Please remove these from this PR and instead let's focus on the following counterpart functions of the geometry side (https://sedona.apache.org/latest/api/sql/Constructor/).

* ST_GeogFromWKB (the Binary type input, not the string type input)
* ST_GeogFromEWKB (the Binary type input, not the string type input)
* ST_GeogFromEWKT
* ST_GeogFromGeohash
* ST_GeogFromText (alias to ST_GeogFromWKT)

Let's start with ST_GeogFromWKB. Please do one function per PR.

PR #2257, opened 2025-08-08
[GH-2258] add MULTIPOLYGON Type in Geography

ZhuochengShang:

Add new kind as MULTIPOLYGON, then instead of returning GEOCOLLECTION kind


jiayuasu (approved):

common/src/main/java/org/apache/sedona/common/S2Geography/Geography.java
```diff
@@ -66,7 +66,8 @@ public enum GeographyKind {
     ENCODED_SHAPE_INDEX(6),
     CELL_CENTER(7),
     SINGLEPOINT(8),
-    SINGLEPOLYLINE(9);
+    SINGLEPOLYLINE(9),
+    MULTIPOLYGON(10);
```
jiayuasu, line 70: The original S2geography lib does not have a MULTIPOLYGON type. Why do you need one here?

ZhuochengShang: This is causing a issue when I was verifies the DataFrame API for reading from WKB and outputting as a string.
Currently, a MultiPolygon is assigned the GeoCollection kind, so the WKT output is returned as "GEOCOLLECTION" instead of "MULTIPOLYGON".
The goal is to ensure that when reading WKB and calling toString() for WKT, the output uses the correct type format.

PR #2262, opened 2025-08-11
[GH-2263]add new constructor ST_GeogFromEWKT

ZhuochengShang:
add new constructor ST_GeogFromEWKT


jiayuasu (changes requested):
Please also add SQL API doc in this PR.

docs/api/sql/geography/Constructor.md
```diff
@@ -18,3 +18,104 @@
  -->
 
 ## ST_GeogFromWKT
+
+Introduction: Construct a Geography from WKT. If SRID is not set, it defaults to 0 (unknown).
+
+!!!note  The output WKT string uses the FIXED precision model, but all computations are performed using JTS’s default 16-digit precision model. Users can specify a different precision model when converting to a string if needed.
```
jiayuasu: Please fix the indent.

docs/api/sql/geography/Constructor.md
```diff
@@ -18,3 +18,104 @@
...
+```
+
+Output:
+
+```
+LINESTRING (0 0, 3 3, 4 4)
+```
+
+
+## ST_GeogFromWKB
```
jiayuasu: Please don't put the doc of other function in this PR. Please only keep ST_GeogFromEWKT, and ST_GeogFromWKT

docs/api/sql/geography/Constructor.md
jiayuasu: And, I don't think there is a way for users to specify the precision model in SQL API. Maybe we should just skip this comment.

docs/api/sql/geography/Constructor.md
```diff
@@ -18,3 +18,58 @@
...
+
+SQL example:
+
+```sql
+SELECT ST_GeogFromEWKT('SRID=4326;LINESTRING (0 0, 3 3, 4 4)')
+```
+
+Output:
+
+```
```
jiayuasu, line 73: Please fix this to include SRID otherwise it is misleading.

common/src/main/java/org/apache/sedona/common/geography/Constructors.java
```diff
@@ -41,6 +41,35 @@ public static Geography geogFromWKT(String wkt, int srid) throws ParseException
...
+      }
+    }
+    try {
+      return geogFromWKT(wkt, SRID);
+    } catch (ParseException e) {
+      String msg = e.getMessage();
+      // Normalize JUST the common JTS phrasing
+      if (msg != null && msg.startsWith("Unknown geometry type:")) {
+        msg = msg.replaceFirst("geometry", "geography");
+      }
```
Kontinuation: Is it cleaner to emit exception messages containing "geography" instead of "geometry" in S2Geography/WKTReader.java in the first place, so that we don't need to do this find and replace here?

docs/api/sql/geography/Constructor.md
```diff
@@ -18,3 +18,94 @@
...
+SELECT ST_GeogFromWKT('LINESTRING (1 2, 3 4, 5 6)', 4326)
+```
+
+Output:
+
+```
+SRID=4326; LINESTRING (1 2, 3 4, 5 6)
+```
+
+## ST_GeogFromWKT
```
Kontinuation: `## ST_GeogFromWKT` seems to appear twice in the doc, if I interpret the diff correctly. We need to remove one of them.

jiayuasu (approved):

PR #2275, opened 2025-08-13
[GH-2276] new constructor of convert Geography to Geometry

ZhuochengShang:
function of convert Geography to Geometry

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/geography/Constructors.java
+++ b/common/src/main/java/org/apache/sedona/common/geography/Constructors.java
@@ -21,7 +21,6 @@ package org.apache.sedona.common.geography;
 import com.google.common.geometry.*;
 import java.util.*;
 import org.apache.sedona.common.S2Geography.*;
-import org.locationtech.jts.algorithm.Orientation;
 import org.locationtech.jts.geom.*;
 import org.locationtech.jts.io.ParseException;
 
@@ -173,7 +172,7 @@ public class Constructors {
       // Build & close ring once (x=lng, y=lat)
       Coordinate[] cs = new Coordinate[n + 1];
       for (int i = 0; i < n; i++) {
-        S2LatLng ll = S2LatLng.fromPoint(L.vertex(i));
+        S2LatLng ll = S2LatLng.fromPoint(L.vertex(i)).normalized();
         cs[i] = new Coordinate(ll.lngDegrees(), ll.latDegrees());
       }
       cs[n] = cs[0];
@@ -219,15 +218,14 @@ public class Constructors {
 
   private static LinearRing ensureOrientation(
       LinearRing ring, boolean wantCCW, GeometryFactory gf) {
-    Coordinate[] cs = ring.getCoordinates();
-    boolean isCCW = org.locationtech.jts.algorithm.Orientation.isCCW(cs);
-    if (!wantCCW && isCCW) {
-      cs = CoordinateArrays.copyDeep(cs);
+    boolean isCCW = org.locationtech.jts.algorithm.Orientation.isCCW(ring.getCoordinates());
+    // If the actual orientation doesn't match the desired orientation, fix it.
+    if (isCCW != wantCCW) {
+      Coordinate[] cs = CoordinateArrays.copyDeep(ring.getCoordinates());
       CoordinateArrays.reverse(cs);
       return gf.createLinearRing(cs);
     }
-    boolean holeIsCCW = Orientation.isCCW(ring.getCoordinates());
-    System.out.println("hole CW = " + !holeIsCCW); // should print true
+    // Otherwise, the ring is already correctly oriented, so return it as is.
     return ring;
   }
 
--- a/common/src/main/java/org/apache/sedona/common/geography/Constructors.java
+++ b/common/src/main/java/org/apache/sedona/common/geography/Constructors.java
@@ -19,14 +19,14 @@
 package org.apache.sedona.common.geography;
 
 import com.google.common.geometry.*;
+import java.io.IOException;
 import java.util.*;
 import org.apache.sedona.common.S2Geography.*;
-import org.locationtech.jts.geom.*;
-import java.io.IOException;
 import org.apache.sedona.common.S2Geography.Geography;
 import org.apache.sedona.common.S2Geography.WKBReader;
 import org.apache.sedona.common.S2Geography.WKTReader;
 import org.apache.sedona.common.utils.GeoHashDecoder;
+import org.locationtech.jts.geom.*;
 import org.locationtech.jts.io.ParseException;
 
 public class Constructors {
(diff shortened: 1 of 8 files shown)


common/src/main/java/org/apache/sedona/common/geography/Constructors.java
```diff
@@ -67,4 +68,174 @@ public static Geography geogCollFromText(String wkt, int srid) throws ParseExcep
...
   }
+
+  public static Geometry geogToGeometry(Geography geography, GeometryFactory geometryFactory)
+      throws ParseException {
+    return geogToGeometry(geography, geometryFactory, geography.getSRID());
+  }
+
+  public static Geometry geogToGeometry(
+      Geography geography, GeometryFactory geometryFactory, int srid) {
+    if (geography == null) return null;
```
Kontinuation: The `srid` is not used in `geogToGeometry(Geography geography, GeometryFactory geometryFactory, int srid)`.

A better approach is to define these 2 functions:

```
Geometry geogToGeometry(Geography geography, GeometryFactory geometryFactory); #1
Geometry geogToGeometry(Geography geography, int srid); #2
```

`GeometryFactory` object has an SRID property, the geometry objects created by that factory will automatically have that SRID. The variant #1 will implicitly use the SRID of the `geometryFactory`.

Variant #2 creates a new `GeometryFactory` object with specified `srid`, and calls #1 to create geometry objects with specified SRID.

common/src/main/java/org/apache/sedona/common/geography/Constructors.java
```diff
@@ -67,4 +68,174 @@ public static Geography geogCollFromText(String wkt, int srid) throws ParseExcep
...
+        S2Polyline pl = lines.get(i);
+        int n = pl.numVertices();
+        Coordinate[] cs = new Coordinate[n];
+        for (int k = 0; k < n; k++) {
+          S2LatLng ll = S2LatLng.fromPoint(pl.vertex(k));
+          cs[k] = new Coordinate(ll.lngDegrees(), ll.latDegrees());
+        }
+        lss[i] = gf.createLineString(cs);
+      }
+      return lss.length == 1 ? lss[0] : gf.createMultiLineString(lss);
```
Kontinuation: Should be determine the type of converted geometry by inspecting the type of `g` (`PolylineGeography` or `SinglePolylineGeography`)? It is easy for us to preserve the original type here.

common/src/main/java/org/apache/sedona/common/geography/Constructors.java
```diff
@@ -67,4 +68,174 @@ public static Geography geogCollFromText(String wkt, int srid) throws ParseExcep
...
+    if (g instanceof PolygonGeography) {
+      S2Polygon s2p = ((PolygonGeography) g).polygon;
+      return s2LoopsToJts(s2p.getLoops(), gf);
+    } else if (g instanceof MultiPolygonGeography) {
+      List<Geography> parts = ((MultiPolygonGeography) g).getFeatures();
+      Polygon[] polys = new Polygon[parts.size()];
+      for (int i = 0; i < parts.size(); i++) {
+        polys[i] = (Polygon) s2LoopsToJts(((PolygonGeography) parts.get(i)).polygon.getLoops(), gf);
+      }
+      return polys.length == 1 ? polys[0] : gf.createMultiPolygon(polys);
```
Kontinuation: I think it is OK to always return multi polygon to preserve its original geography type.

common/src/main/java/org/apache/sedona/common/geography/Constructors.java
```diff
@@ -67,4 +68,174 @@ public static Geography geogCollFromText(String wkt, int srid) throws ParseExcep
...
+    List<List<LinearRing>> holesPerShell = new ArrayList<>();
+
+    // Stack of current ancestor shells: each frame = {shellIndex, depth}
+    //    depth 0: Shell A
+    //          depth 1: Hole H1  (a lake in A)
+    //          depth 2: Shell S2 (an island in that lake A)
+    //                depth 3: Hole H3 (a pond on that island)
+    //    depth 0: Shell B     (disjoint area)
+    //          depth 1: Hole H2   (a lake in B)
+    Deque<int[]> shellStack = new ArrayDeque<>();
```
Kontinuation: It is a stack by looking at the naming of this variable and the actual usage of it, maybe ArrayList is a better container for this?

Kontinuation (approved):

docs/api/sql/geography/Constructor.md
```diff
@@ -140,3 +140,27 @@ Output:
...
+`ST_GeogToGeometry (geog: Geography)`
+
+`ST_GeogToGeometry (geog: Geography, srid: Integer)`
+
+Since: `v1.8.0`
+
+SQL example:
+
+```sql
+SELECT ST_GeogToGeometry(ST_GeogFromWKT('MULTILINESTRING ((90 90, 20 20, 10 40), (40 40, 30 30, 40 20, 30 10))', 4326)))
```
Kontinuation: This example query has unbalanced parenthesis.

Kontinuation (changes requested):

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/st_constructors.scala
```diff
@@ -23,6 +23,8 @@ import org.apache.spark.sql.sedona_sql.DataFrameShims._
 import org.apache.spark.sql.sedona_sql.expressions.geography.{ST_GeogCollFromText, ST_GeogFromEWKB, ST_GeogFromText, ST_GeogFromWKB, ST_GeogFromWKT}
 import org.apache.spark.sql.sedona_sql.expressions.geography.{ST_GeogCollFromText, ST_GeogFromEWKT, ST_GeogFromText, ST_GeogFromWKB, ST_GeogFromWKT}
 import org.apache.spark.sql.sedona_sql.expressions.geography.{ST_GeogCollFromText, ST_GeogFromGeoHash, ST_GeogFromText, ST_GeogFromWKB, ST_GeogFromWKT}
+import org.apache.spark.sql.sedona_sql.DataFrameShims.{wrapExpression, _}
+import org.apache.spark.sql.sedona_sql.expressions.geography.{ST_GeogCollFromText, ST_GeogFromEWKB, ST_GeogFromEWKT, ST_GeogFromText, ST_GeogFromWKB, ST_GeogFromWKT, ST_GeogToGeometry}
```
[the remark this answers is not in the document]
jiayuasu: The AI is right. The import line got duplicated many times.

common/src/main/java/org/apache/sedona/common/geography/Constructors.java
```diff
@@ -79,4 +83,185 @@ public static Geography geogFromGeoHash(String geoHash, Integer precision) {
       throw new RuntimeException(e);
     }
   }
+
+  public static Geometry geogToGeometry(Geography geography, int srid) {
+    int SRID = srid == geography.getSRID() ? srid : geography.getSRID();
```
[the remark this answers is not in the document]
jiayuasu: Please don't complicate the logic here. I suggest we only allow 1 signature that is `geogToGeometry(Geography geography)`. The function always takes SRID from the input geography.

spark/common/src/main/scala/org/apache/sedona/sql/UDF/Catalog.scala
```diff
@@ -25,6 +25,7 @@ import org.apache.spark.sql.sedona_sql.expressions._
 import org.apache.spark.sql.sedona_sql.expressions.geography.{ST_GeogCollFromText, ST_GeogFromGeoHash, ST_GeogFromText, ST_GeogFromWKB, ST_GeogFromWKT}
 import org.apache.spark.sql.sedona_sql.expressions.geography.{ST_GeogCollFromText, ST_GeogFromEWKT, ST_GeogFromText, ST_GeogFromWKB, ST_GeogFromWKT}
 import org.apache.spark.sql.sedona_sql.expressions.geography.{ST_GeogCollFromText, ST_GeogFromEWKB, ST_GeogFromEWKT, ST_GeogFromText, ST_GeogFromWKB, ST_GeogFromWKT}
+import org.apache.spark.sql.sedona_sql.expressions.geography.{ST_GeogCollFromText, ST_GeogFromEWKB, ST_GeogFromEWKT, ST_GeogFromText, ST_GeogFromWKB, ST_GeogFromWKT, ST_GeogToGeometry}
```
jiayuasu: The input line repeated 3 times

jiayuasu:
@Kontinuation Does it make sense to save Geography to WKB and load into geometry using WKB? I suppose the performance is bad?

PR #2283, opened 2025-08-14
[GH-2284]add constructor of ST_TryToGeometry and ST_ToGeometry

ZhuochengShang:
add constructor of ST_TryToGeography and ST_ToGeography

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/geography/Constructors.java
+++ b/common/src/main/java/org/apache/sedona/common/geography/Constructors.java
@@ -74,46 +74,16 @@ public class Constructors {
     return geogFromWKT(wkt, srid);
   }
 
-  public static Geography tryToGeography(String geogString) {
+  public static Geometry tryToGeometry(Geography geography) {
     try {
-      return toGeography(geogString);
+      return toGeometry(geography);
     } catch (Exception e) {
       return null;
     }
   }
 
-  public static Geography toGeography(String geogString) throws ParseException {
-    if (geogString == null || geogString.trim().isEmpty()) {
-      return null;
-    }
-    geogString = geogString.trim();
-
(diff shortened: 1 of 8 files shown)


Kontinuation:
Does `ST_TryToGeography` need to support converting geometry values into geography values? This PR only supports converting string values into geography values.

common/src/main/java/org/apache/sedona/common/geography/Constructors.java
```diff
@@ -70,6 +70,48 @@ public static Geography geogCollFromText(String wkt, int srid) throws ParseExcep
...
+    }
+    // WKB hex string (only hex digits)
+    if (isHex(geogString)) {
+      byte[] wkbBytes = WKBReader.hexToBytes(geogString);
+      return Constructors.geogFromWKB(wkbBytes);
+    }
+    // GeoHash (short alphanumeric string without spaces)
+    if (isGeoHash(geogString)) {
+      return Constructors.geogFromGeoHash(geogString, 16);
+    }
```
Kontinuation: I'm not sure whether we should support GeoHash. A GeoHash value could happen to only contain hex characters, we'll interpret such data as WKB hex string and get errors or silently get unwanted data.

jiayuasu:
@ZhuochengShang any update on this?

jiayuasu:
We only need to support ST_GeomToGeography. We don't need to support string input

jiayuasu:
I think we already have ST_GeogToGemetry. We should close this PR?

ZhuochengShang:
> I think we already have ST_GeogToGemetry. We should close this PR?

yes, I think so

Kontinuation:
I'm closing this in favor of https://github.com/apache/sedona/pull/2275.

PR #2285, opened 2025-08-14
[GH-2286] ST_Envelope on Geography

ZhuochengShang:
envelope function of Geography


common/src/main/java/org/apache/sedona/common/geography/Functions.java
```diff
@@ -0,0 +1,72 @@
...
+    Geography envelope;
+    if (!rect.lng().isInverted()) {
+      // No antimeridian crossing → ONE polygon
+      envelope = new PolygonGeography(rectToPolygon(lngLo, latLo, lngHi, latHi));
+    } else {
+      // Crossing → split into two polygons
+      S2Polygon left = rectToPolygon(lngLo, latLo, 180.0, latHi);
+      S2Polygon right = rectToPolygon(-180.0, latLo, lngHi, latHi);
+      envelope =
+          new GeographyCollection(Geography.GeographyKind.MULTIPOLYGON, List.of(left, right));
```
Kontinuation: Why not use `new MultiPolygonGeography(List.of(left, right))`?

common/src/main/java/org/apache/sedona/common/geography/Functions.java
```diff
@@ -0,0 +1,72 @@
...
+import com.google.common.geometry.*;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.sedona.common.S2Geography.Geography;
+import org.apache.sedona.common.S2Geography.GeographyCollection;
+import org.apache.sedona.common.S2Geography.PolygonGeography;
+
+public class Functions {
+
+  public static Geography getEnvelope(Geography geography) {
```
Kontinuation: I suggest that we add a parameter `splitAtAntiMeridian` to control whether the lat-lng-rect was splitted at anti-meridian or not. `Geography` could naturally handle anti-meridian, the cross anti-meridian problem stems only when we treat geography objects as planar shapes, or we are exporting data in a format/spec where anti-meridian splitting is required.

docs/api/sql/geography/Function.md
```diff
@@ -16,3 +16,26 @@
...
+`ST_Envelope (A: Geography, splitAtAntiMeridian: Boolean)`
+
+Since: `v1.8.0`
+
+SQL Example
+
+```sql
+SELECT ST_Envelope(ST_GeogFromWKT('MULTIPOLYGON (((177.285 -18.28799, 180 -18.28799, 180 -16.02088, 177.285 -16.02088, 177.285 -18.28799)),((-180 -18.28799, -179.7933 -18.28799, -179.7933 -16.02088, -180 -16.02088, -180 -18.28799)))
+', false))
+```
```
Kontinuation: This example SQL is incorrect. `false` should be passed as into ST_Envelope instead of ST_GeogFromWKT.

docs/api/sql/geography/Function.md
```diff
@@ -16,3 +16,26 @@
  specific language governing permissions and limitations
  under the License.
  -->
+
+## ST_Envelope
+
+Introduction: This function returns the bounding box (envelope) of A. If the envelope crosses the antimeridian (the 180° longitude line), you can set the split parameter to true. This will return a Geography object containing two separate Polygon objects, split along that line.
```
Kontinuation: We should emphasize that the bounding box is in **cylindrical** topology rather than a spherical one.

The unclear semantics of this function is probably the reason why ST_ENVELOPE is depcrecated by snowflake: https://docs.snowflake.com/en/sql-reference/functions/st_envelope

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/geography/Functions.scala
```diff
@@ -0,0 +1,22 @@
+/*
```
jiayuasu: This file has no actual code. Is it needed?

Kontinuation: I missed that when reviewing it. I have removed this file.

PR #2298, opened 2025-08-18
[GH-2302] add constructor of Geometry to Geography

ZhuochengShang:
add constructor of Geometry to Geography


common/src/main/java/org/apache/sedona/common/geography/Constructors.java
```diff
@@ -264,4 +264,271 @@ private static Geometry collectionToGeom(Geography g, GeometryFactory gf) {
...
+      return lineToGeog(geom);
+    } else if ("MultiLineString".equalsIgnoreCase(type)) {
+      return mLineToGeog(geom);
+    } else if ("Polygon".equalsIgnoreCase(type)) {
+      return polyToGeog(geom);
+    } else if ("MultiPolygon".equalsIgnoreCase(type)) {
+      return mPolyToGeog(geom);
+    } else if ("GeometryCollection".equalsIgnoreCase(type)) {
+      return geomCollToGeog(geom);
+    } else {
```
Kontinuation: I prefer directly using `instanceof` to check the geometry type. Reference: https://github.com/apache/sedona/blob/658bf60548be0fa3c60ea8b0a9637dabab4b30bc/common/src/main/java/org/apache/sedona/common/geometrySerde/GeometrySerializer.java#L42-L59

The order of the checks is important. We need to check for GeometryCollection last because Multi geometries are subclasses of GeometryCollection.

common/src/main/java/org/apache/sedona/common/geography/Constructors.java
```diff
@@ -264,4 +264,271 @@ private static Geometry collectionToGeom(Geography g, GeometryFactory gf) {
...
+
+  private static Geography lineToGeog(Geometry geom) throws IOException {
+    LineString ls = (LineString) geom;
+
+    // Build S2 points
+    List<S2Point> pts = toS2Points(ls.getCoordinates());
+    if (pts.size() < 2) {
+      // empty or degenerate → empty single polyline
+      SinglePolylineGeography empty = new SinglePolylineGeography();
+      empty.setSRID(geom.getSRID());
```
Kontinuation: SRID is only handled in this place but not other places. We can set SRID in `geomToGeography` instead of setting it everywhere.

common/src/main/java/org/apache/sedona/common/geography/Constructors.java
```diff
@@ -264,4 +264,271 @@ private static Geometry collectionToGeom(Geography g, GeometryFactory gf) {
...
+      } else if (g instanceof MultiLineString) {
+        sub = mLineToGeog(g);
+      } else if (g instanceof Polygon) {
+        sub = polyToGeog(g);
+      } else if (g instanceof MultiPolygon) {
+        sub = mPolyToGeog(g);
+      } else if (g instanceof GeometryCollection) {
+        sub = geomCollToGeog(g);
+      }
+      if (sub != null) coll.features.add(sub);
```
Kontinuation: Why not call `geomToGeography` but replicate the code here?

common/src/main/java/org/apache/sedona/common/geography/Constructors.java
```diff
@@ -264,4 +264,271 @@ private static Geometry collectionToGeom(Geography g, GeometryFactory gf) {
     }
     return gf.createGeometryCollection(gs);
   }
+
+  public static Geography geomToGeography(Geometry geom) throws IOException {
```
Kontinuation: We are not doing IO here so IOException is not a proper exception type.

common/src/test/java/org/apache/sedona/common/Geography/ConstructorsTest.java
```diff
@@ -245,4 +245,27 @@ public void polygon_threeHoles() throws Exception {
...
+            + ")";
+    Geometry g = new org.locationtech.jts.io.WKTReader().read(wkt);
+    g.setSRID(4326);
+    Geography got = Constructors.geomToGeography(g);
+    String expected = "SRID=4326; " + wkt;
+    assertEquals(4326, got.getSRID());
+    org.locationtech.jts.io.WKTWriter wktWriter = new org.locationtech.jts.io.WKTWriter();
+    wktWriter.setPrecisionModel(new PrecisionModel(PrecisionModel.FIXED));
+    assertEquals(expected, got.toString());
+  }
```
Kontinuation, line 270: We need more tests to achieve full coverage of geometry types and their internal structures.

common/src/main/java/org/apache/sedona/common/geography/Constructors.java
```diff
@@ -264,4 +264,271 @@ private static Geometry collectionToGeom(Geography g, GeometryFactory gf) {
...
+    S2Builder builder = new S2Builder.Builder().build();
+    S2PointVectorLayer layer = new S2PointVectorLayer();
+    // must call build() before reading out the points
+    S2Error error = new S2Error();
+    for (int i = 0; i < pts.length; i++) {
+      double lon = pts[i].x;
+      double lat = pts[i].y;
+      S2Point s2Point = S2LatLng.fromDegrees(lat, lon).toPoint();
+
+      builder.startLayer(layer);
```
Kontinuation: Adding the same layer repeatedly does not look correct.

common/src/main/java/org/apache/sedona/common/geography/Constructors.java
```diff
@@ -264,4 +264,271 @@ private static Geometry collectionToGeom(Geography g, GeometryFactory gf) {
...
+      Geometry g = gc.getGeometryN(i);
+      Geography sub = null;
+
+      if (g instanceof Point) {
+        sub = pointToGeog((Point) g);
+      } else if (g instanceof MultiPoint) {
+        MultiPoint mp = (MultiPoint) g;
+        for (int k = 0; k < mp.getNumGeometries(); k++) {
+          coll.features.add(pointToGeog((Point) mp.getGeometryN(k)));
+        }
```
Kontinuation: This flattens multi point as separate geography objects instead of a single PointGeography. This does not look right.

common/src/main/java/org/apache/sedona/common/geography/Constructors.java
```diff
@@ -264,4 +264,241 @@ private static Geometry collectionToGeom(Geography g, GeometryFactory gf) {
...
+      return new SinglePointGeography();
+    }
+    return new SinglePointGeography(points.get(0));
+  }
+
+  private static Geography mPointToGeog(Geometry geom) throws IOException {
+    Coordinate[] pts = geom.getCoordinates();
+    if (pts.length == 0 || Double.isNaN(pts[0].x) || Double.isNaN(pts[0].y)) {
+      return new PointGeography();
+    }
```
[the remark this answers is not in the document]
Kontinuation: I believe that Copilot is making a mistake. Or conditions short circuits and won't evaluate pts[0] when pts is empty.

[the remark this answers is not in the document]
Kontinuation: I believe that Copilot is making a mistake. Or conditions short circuits and won't evaluate pts[0] when pts is empty.

common/src/main/java/org/apache/sedona/common/geography/Constructors.java
```diff
@@ -264,4 +264,241 @@ private static Geometry collectionToGeom(Geography g, GeometryFactory gf) {
...
+        features.add(layer.getPolyline());
+      }
+      // Empty/degenerate parts are skipped
+    }
+    return new PolylineGeography(features);
+  }
+
+  private static Geography polyToGeog(Geometry geom) throws IOException {
+    // Build S2 loops: exterior + holes
+    Polygon poly = (Polygon) geom;
```
Kontinuation: We can define this function as `Geography polyToGeog(Polygon poly)` to avoid the cast. Same suggestion for other functions.

common/src/main/java/org/apache/sedona/common/geography/Constructors.java
```diff
@@ -264,4 +264,241 @@ private static Geometry collectionToGeom(Geography g, GeometryFactory gf) {
...
+
+    return new MultiPolygonGeography(
+        Geography.GeographyKind.MULTIPOLYGON, Collections.unmodifiableList(polys));
+  }
+
+  private static Geography geomCollToGeog(Geometry geom) throws Exception {
+    if (!(geom instanceof GeometryCollection)) {
+      throw new IllegalArgumentException(
+          "geomCollToGeog expects GeometryCollection, got " + geom.getGeometryType());
+    }
```
Kontinuation: Define it to accept `GeometryCollection geom` parameter to avoid this instanceof check.

common/src/test/java/org/apache/sedona/common/Geography/ConstructorsTest.java
```diff
@@ -245,4 +245,70 @@ public void polygon_threeHoles() throws Exception {
     assertEquals(expected, got.toString());
     assertEquals(0, got.getSRID());
   }
+
+  @Test
+  public void geomToGeography() throws Exception {
```
Kontinuation: Change the name of this test to `MultiPolygonGeomToGeography` to be clear.

common/src/main/java/org/apache/sedona/common/geography/Constructors.java
```diff
@@ -264,4 +264,241 @@ private static Geometry collectionToGeom(Geography g, GeometryFactory gf) {
...
+    S2Point s2Point = S2LatLng.fromDegrees(lat, lon).toPoint();
+    S2Builder builder = new S2Builder.Builder().build();
+    S2PointVectorLayer layer = new S2PointVectorLayer();
+    builder.startLayer(layer);
+    builder.addPoint(s2Point);
+
+    S2Error error = new S2Error();
+    if (!builder.build(error)) {
+      throw new IOException("Failed to build S2 point layer: " + error.text());
+    }
```
Kontinuation: We are still throwing IOExceptions where there's no IO involved. We can throw `IllegalArgumentException` to indicate that the input is not right.

Actually, S2Error simply originates from an `IllegalArgumentException`: https://github.com/google/s2-geometry-library-java/blob/57151fc0f7fe543b75e9f743bf897e81928d5c3f/library/src/com/google/common/geometry/S2Builder.java#L895-L906

common/src/main/java/org/apache/sedona/common/geography/Constructors.java
```diff
@@ -264,4 +264,241 @@ private static Geometry collectionToGeom(Geography g, GeometryFactory gf) {
...
+  private static List<S2Point> toS2Points(Coordinate[] coords) throws IOException {
+    List<S2Point> points = new ArrayList<>(coords.length);
+    for (int i = 0; i < coords.length; i++) {
+      double lon = coords[i].x;
+      double lat = coords[i].y;
+      S2Point s2Point = S2LatLng.fromDegrees(lat, lon).toPoint();
+      points.add(s2Point);
+    }
+    return points;
+  }
```
Kontinuation: I don't see the code dropping NaNs and consecutive duplicates in the loop, the comment or the implementation may not be correct.

Should we add a test case for linestring or multi-linestring containing consecutive duplicates?

Issue #2240, opened 2025-08-06
Error when reading back nested geometry from Parquet with pyspark 3.5


EmmanuelNicolau-TomTom:
## Expected behaviour 

Writing a dataframe with a geometry column nested within an array to Parquet, can be read back without issues.

## Actual behaviour

```
Py4JJavaError: An error occurred while calling o232.collectToPython.
: org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 26.0 failed 1 times, most recent failure: Lost task 0.0 in stage 26.0 (TID 36) (17e1c21c3c4d executor driver): java.lang.IllegalArgumentException: Spark type: StructType(StructField(geometry,BinaryType,true)) doesn't match the type: StructType(StructField(geometry,org.apache.spark.sql.sedona_sql.UDT.GeometryUDT@1114c728,true)) in column vector
        at org.apache.spark.sql.execution.datasources.parquet.ParquetColumnVector.<init>(ParquetColumnVector.java:70)
        at org.apache.spark.sql.execution.datasources.parquet.ParquetColumnVector.<init>(ParquetColumnVector.java:117)
        at org.apache.spark.sql.execution.datasources.parquet.VectorizedParquetRecordReader.initBatch(VectorizedParquetRecordReader.java:286)
        at org.apache.spark.sql.execution.datasources.parquet.VectorizedParquetRecordReader.initBatch(VectorizedParquetRecordReader.java:306)
        at org.apache.spark.sql.execution.datasources.parquet.ParquetFileFormat.$anonfun$buildReaderWithPartitionValues$2(ParquetFileFormat.scala:293)
        at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.org$apache$spark$sql$execution$datasources$FileScanRDD$$anon$$readCurrentFile(FileScanRDD.scala:217)
        at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.nextIterator(FileScanRDD.scala:279)
        at org.apache.spark.sql.execution.datasources.FileScanRDD$$anon$1.hasNext(FileScanRDD.scala:129)
        at org.apache.spark.sql.execution.FileSourceScanExec$$anon$1.hasNext(DataSourceScanExec.scala:593)
        at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.columnartorow_nextBatch_0$(Unknown Source)
[...]

PR #2359, opened 2025-09-19
[GH-2240] Fix write and read nested geometry array using vectorized parquet reader 

zhangfengcdt:

This PR addresses SPARK-48942, a bug that occurs when reading nested geometry arrays from Parquet files using Spark's vectorized reader. The fix implements compatibility checks for UserDefinedTypes (UDTs) in nested structures and adds workaround utilities for schema transformation.

- Adds UDT compatibility checking in Parquet column vector operations to handle type mismatches 
- Implements utility classes for transforming nested GeometryUDT schemas to avoid the bug 
- Provides comprehensive test coverage for various nested geometry scenarios

## How was this patch tested?
new tests are added to geoparquetIOTests.scala

### DIFF
--- a/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/UDT/TransformNestedUDTParquet.scala
+++ b/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/UDT/TransformNestedUDTParquet.scala
@@ -100,53 +100,45 @@ class TransformNestedUDTParquet(spark: SparkSession) extends Rule[LogicalPlan] {
   }
 
   /**
-   * Transform a schema to handle nested UDT by processing each top-level field. This preserves
-   * top-level GeometryUDT fields while transforming nested ones to BinaryType.
+   * Transform a schema to handle nested UDT by processing each field. This preserves top-level
+   * GeometryUDT fields while transforming nested ones to BinaryType.
    */
   private def transformSchemaForNestedUDT(schema: StructType): StructType = {
-    StructType(
-      schema.fields.map(field => field.copy(dataType = transformTopLevelUDT(field.dataType))))
+    StructType(schema.fields.map(field =>
+      field.copy(dataType = transformDataType(field.dataType, isTopLevel = true))))
   }
 
   /**
-   * Transform a top-level field's data type, preserving GeometryUDT at the top level but
-   * converting nested GeometryUDT to BinaryType.
+   * Recursively transform data types based on nesting level.
+   * @param dataType
+   *   the data type to transform
+   * @param isTopLevel
+   *   true if this is a top-level field (preserves GeometryUDT), false if nested (converts
+   *   GeometryUDT to BinaryType)
+   * @return
+   *   transformed data type
    */
-  private def transformTopLevelUDT(dataType: DataType): DataType = {
+  private def transformDataType(dataType: DataType, isTopLevel: Boolean): DataType = {
     dataType match {
-      case ArrayType(elementType, containsNull) =>
-        ArrayType(transformNestedUDTToBinary(elementType), containsNull)
-      case MapType(keyType, valueType, valueContainsNull) =>
-        MapType(
-          transformNestedUDTToBinary(keyType),
-          transformNestedUDTToBinary(valueType),
-          valueContainsNull)
-      case StructType(fields) =>
-        StructType(
-          fields.map(field => field.copy(dataType = transformNestedUDTToBinary(field.dataType))))
-      case _: GeometryUDT => dataType // Preserve top-level GeometryUDT
-      case other => other
-    }
-  }
+      case _: GeometryUDT =>
+        if (isTopLevel) dataType else BinaryType // Preserve at top-level, convert if nested
 
-  /**
-   * Recursively transform nested data types, converting ALL GeometryUDT to BinaryType. This is
-   * used for nested structures where GeometryUDT must be converted.
-   */
-  private def transformNestedUDTToBinary(dataType: DataType): DataType = {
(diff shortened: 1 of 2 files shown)


spark/common/src/main/java/org/apache/spark/sql/execution/datasources/geoparquet/internal/ParquetColumnVector.java
```diff
@@ -31,11 +33,65 @@
 import org.apache.spark.sql.types.DataTypes;
 import org.apache.spark.sql.types.MapType;
 import org.apache.spark.sql.types.StructType;
+import org.apache.spark.sql.types.UserDefinedType;
 
 /**
  * Contains necessary information representing a Parquet column, either of primitive or nested type.
  */
 final class ParquetColumnVector {
```
jiayuasu, line 38: I am confused. This is introducing changes to files in `datasources/geoparquet` folder, which is specific to `geoparquet` data source. But your test case is against `parquet` data source. So looks like the change in this file is not used at all?

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/UDT/TransformNestedUDTParquet.scala
```diff
@@ -0,0 +1,153 @@
...
+          transformNestedUDTToBinary(keyType),
+          transformNestedUDTToBinary(valueType),
+          valueContainsNull)
+      case StructType(fields) =>
+        StructType(
+          fields.map(field => field.copy(dataType = transformNestedUDTToBinary(field.dataType))))
+      case udt: UserDefinedType[_] => transformNestedUDTToBinary(udt.sqlType)
+      case other => other
+    }
+  }
```
Kontinuation, line 144: I'm not sure if we can merge these 3 functions into one.

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/transform/TransformNestedUDTForParquet.scala
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.spark.sql.sedona_sql.UDT.GeometryUDT
+import org.apache.spark.sql.types._
+
+/**
+ * Expression that transforms nested GeometryUDT to BinaryType to work around SPARK-48942.
+ *
+ * This expression recursively transforms the schema of nested structures containing GeometryUDT,
+ * converting them to BinaryType for compatibility with Spark's vectorized Parquet reader.
+ */
+case class TransformNestedUDTForParquet(child: Expression) extends UnaryExpression {
```
Kontinuation: I cannot find where it is being used.

spark/common/src/test/scala/org/apache/sedona/sql/geoparquetIOTests.scala
```diff
@@ -784,6 +784,68 @@ class geoparquetIOTests extends TestBaseScala with BeforeAndAfterAll {
...
+      df.write.mode("overwrite").format("parquet").save(testPath)
+
+      // The fix allows vectorized reading to handle UDT compatibility properly
+      val readDf = sparkSession.read.format("parquet").load(testPath)
+
+      // Verify the geometry data is correct
+      val result = readDf.collect()
+      assert(result.length == 1)
+      val nestedArray = result(0).getSeq[Any](0)
+      assert(nestedArray.length == 1)
```
Kontinuation, line 809: What is the expected type of nested UDT values (binary or geometry object)? According to `TransformNestedUDTParquet.scala` I guess it is binary.

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/transform/TransformNestedUDTForParquet.scala
zhangfengcdt: This one has been renamed/moved under the udt package, I should have removed it from this old package.

spark/common/src/main/java/org/apache/spark/sql/execution/datasources/geoparquet/internal/ParquetColumnVector.java
zhangfengcdt: I was trying to see if we can support nested GeometryUDT by make this change but realized that based on the geoparquet spec, this should not be supported. 

The original implementation already reject this so I will revert the changes here.

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/UDT/TransformNestedUDTParquet.scala
zhangfengcdt: Yes, I think we can combine transformTopLevelUDT() and transformNestedUDTToBinary() for sure since they are just different in a boollean flag. I'd keep transformSchemaForNestedUDT separate as the entry point for schema.

spark/common/src/test/scala/org/apache/sedona/sql/geoparquetIOTests.scala
zhangfengcdt: The type read back is actually geometry.   The parquet metadata stores GeometryUDT information in the Spark schema metadata, and when it is read back Spark automatically reads this back from the SPARK_METADATA_KEY.  

The TransformNestedUDTParquet rule just fixes / ensures that nested GeometryUDT gets properly handled regardless of which metadata source is used.

I have added some tests after read back to test if the regular geometry operations work.

Issue #2230, opened 2025-08-05
[EPIC] Implement More Geopandas Functions


petern48:
This is an umbrella issue that will encompass many other issues for implementing additional geopandas functionality. 

Setup:
- Read the [Sedona Geopandas developer guide](https://sedona.apache.org/latest-snapshot/community/geopandas/)
- Setup build environment by following these [compile docs](https://github.com/apache/sedona/blob/master/docs/setup/compile.md)

Implementation Steps
1. Pick a `GeoSeries` function to implement from the (subset) list below or from the [full list in the original documentation](https://geopandas.org/en/stable/docs/reference/geoseries.html).
2. Look for a similar function that has been implemented in Sedona Geopandas already  - Chances are, you'll just need to use one of the helper functions `_query_geometry_column` or `_row_wise_operations`
3. Implement function in `geoseries.py` and `base.py`. Usually, this is as simple as calling the corresponding Sedona function (e.g `ST_IsRing` for `.is_ring()`. The `base.py` part is important so geodataframe can also execute it.
4. Add documentation under the `base.py` implementation by copy and pasting from geopandas docstrings.
5. Add tests to `test_geoseries.py` and `test_match_geopandas_series.py` following similar conventions.
- In `test_geoseries.py`, include the example in the docstring as part of the test.
- For `test_match_geopandas_series.py`, follow the convention of similar functions. Most of the time, you just run the function on each of the `self.geoms` and compare output with geopandas using the helper functions.
7. Create a new issue on GitHub. Either create a subissue of this [EPIC] issue or explicitly mention this issue. Then, submit a PR linked to the issue you just created (not this issue). Go ahead and CC @petern48 on your PR too, so I can review
8. Repeat!

**Note about AI use**: I have zero problem with contributors using AI to help them write their code faster. HOWEVER, I strongly urge you to at least run your code and tests locally on your laptop, you're not going to have a good time trying to iterate on failed CI runs. If you need help fixing your code to pass the tests, it's fine to ask for help. But if you constantly are blindly pushing changes an LLM told you and waiting for CI to tell you if your code is correct, you will be putting an annoying pain on reviewers, since we need to manually approve your CI to run after each change. I also recommend you learn to do your first one on your own, by following an example PR or implementation. It should be fairly easy to figure out.

Example PR: https://github.com/apache/sedona/pull/2232

List of unimplemented functions along with existing functions to model their implementations off of:

Unary predicates (boolean operations): model off of `is_empty`
- [x] is_ring
[...]

PR #2386, opened 2025-10-10
[GH-2385] Implement `is_ccw`

### LINKED ISSUE #2385: Geopandas: Implement `is_ccw`
parent #2230

yunchipang:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2385

Implements `is_ccw`

## How was this patch tested?

Tests added

### DIFF
--- a/python/sedona/spark/geopandas/base.py
+++ b/python/sedona/spark/geopandas/base.py
@@ -389,16 +389,16 @@ class GeoFrame(metaclass=ABCMeta):
     @property
     def is_ccw(self):
         """
-        Return a ``Series`` of ``dtype('bool')`` with value ``True`` if a 
+        Return a ``Series`` of ``dtype('bool')`` with value ``True`` if a
         LineString or LinearRing is counterclockwise.
 
-        Note that there are no checks on whether lines are actually closed 
-        and not self-intersecting, while this is a requirement for ``is_ccw``. 
-        The recommended usage of this property for LineStrings is 
-        ``GeoSeries.is_ccw & GeoSeries.is_simple`` and for LinearRings 
+        Note that there are no checks on whether lines are actually closed
+        and not self-intersecting, while this is a requirement for ``is_ccw``.
+        The recommended usage of this property for LineStrings is
+        ``GeoSeries.is_ccw & GeoSeries.is_simple`` and for LinearRings
         ``GeoSeries.is_ccw & GeoSeries.is_valid``.
 
-        This property will return False for non-linear geometries and for 
+        This property will return False for non-linear geometries and for
         lines with fewer than 4 points (including the closing point).
 
         Examples


python/tests/geopandas/test_geoseries.py
```diff
@@ -882,7 +882,21 @@ def test_is_ring(self):
         self.check_pd_series_equal(result, expected)
 
     def test_is_ccw(self):
-        pass
+        s = GeoSeries(
+            [
+                LinearRing([(0, 0), (0, 1), (1, 1), (0, 0)]),
+                LinearRing([(0, 0), (1, 1), (0, 1), (0, 0)]),
```
petern48, line 888:
```suggestion
                LinearRing([(0, 0), (1, 1), (0, 1), (0, 0)]),
                LineString([(0, 0), (1, 1), (0, 1), (-1, -1)]),
```
Let's add a line that isn't a full ring that should return True.

python/tests/geopandas/test_geoseries.py
```diff
@@ -882,7 +882,21 @@ def test_is_ring(self):
...
+                LineString([(0, 0), (1, 1), (0, 1)]),
+                Point(3, 3),
```
petern48, line 890:
```suggestion
                Point(3, 3),
                LineString(),
                MultiLineString([LineString([(0, 0), (1, 1), (0, 1), (-1, -1)])]),
                GeometryCollection([LineString([(0, 0), (1, 1), (0, 1), (-1, -1)])]),
```

A few more cases, I'd like to test. The first is `LINESTRING EMPTY`. The other two are all collections of a single linestring that would return True. Make sure to validate that the expected result is the same as geopandas.

python/tests/geopandas/test_match_geopandas_series.py
```diff
@@ -585,7 +585,10 @@ def test_is_ring(self):
             self.check_pd_series_equal(sgpd_result, gpd_result)
 
     def test_is_ccw(self):
-        pass
+        for geom in self.linestrings:
```
petern48, line 588:
```suggestion
        for geom in self.geoms:
```

Since the docs explicitly specify the expected result for non-linear geometries is False, then we should test that they do correctly return that result.

petern48: I didn't catch it originally (totally my bad), but the same goes for your [other PR](https://github.com/apache/sedona/pull/2378) for `is_closed` since the docstring mentions the result should be False. Could you try updating it to `self.geoms` there too? I think it would pass, but it fails, we can instead leave alone and file a new (bug) issue to fix the implementation in a separate PR.

python/sedona/spark/geopandas/geoseries.py
```diff
@@ -966,13 +966,12 @@ def is_ring(self):
 
     @property
     def is_ccw(self):
-        # Implementation of the abstract method.
-        raise NotImplementedError(
-            _not_implemented_error(
-                "is_ccw",
-                "Tests if LinearRing geometries are oriented counter-clockwise.",
-            )
+        spark_expr = stf.ST_IsPolygonCCW(self.spark.column)
```
petern48, line 969: After seeing the test failures, I took a look at the [documentation](https://sedona.apache.org/latest-snapshot/api/sql/Function/#st_ispolygonccw) and [implementation](https://github.com/apache/sedona/blob/6dc39e753b9d2087d27d2a6072b95d3eda91c53f/common/src/main/java/org/apache/sedona/common/Functions.java#L1493-L1508) of `ST_IsPolygonCCW` function. Unfortunately, it doesn't look like `ST_IsPolygonCCW` will not work for us, since it only handles Polygons and returns False for all linestrings. Looking at the [shapely implementation](https://github.com/shapely/shapely/blob/458b8bf6e6b17c7515151a72f221738e86fce6b9/src/ufuncs.c#L141-L173) (which geopandas uses) `is_ccw` is a function with some custom logic, so there doesn't seem to be a direct function equivalent in Sedona currently.

This doesn't seem very easy to implement, so I think it's better to skip this one for now. Sorry about that @yunchipang

python/tests/geopandas/test_match_geopandas_series.py
yunchipang: @petern48 thanks for the heads up! sorry I was not aware of the situation. I've raised #2387 to fix this for `is_closed` and I'm going to close this PR. please lmk which ones in #2230 could be easier to implement if you have some in mind. i'd love to help!

petern48: Sure! I think in general, most should be pretty easy. As I just mentioned in the issue, I updated the list with corresponding Sedona functions. I'd recommend any of those, since I expect them to be straightforward. It's possible that some are not, but the only way to be sure is to try it out. I think most cases will be as simple as your `is_closed` PR.

PR #2387, opened 2025-10-11
[GH-2377] test `is_closed` on all geo types

### LINKED ISSUE #2377: Geopandas: Implement `is_closed`
parent #2230

yunchipang:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2377 

Expands `test_is_closed` to cover all geometry types, not just LineString/LinearRing, to ensure compliance with the documented behavior.

## How was this patch tested?

Tests added

### DIFF
--- a/python/sedona/spark/geopandas/geoseries.py
+++ b/python/sedona/spark/geopandas/geoseries.py
@@ -979,7 +979,7 @@ class GeoSeries(GeoFrame, pspd.Series):
         # Only check LineStrings; return False for all other geometry types
         spark_expr = F.when(
             stf.ST_GeometryType(self.spark.column) == "ST_LineString",
-            stf.ST_IsClosed(self.spark.column)
+            stf.ST_IsClosed(self.spark.column),
         ).otherwise(False)
 
         result = self._query_geometry_column(


petern48:
I can't trigger CI to run myself, so we'll have to wait for @jiayuasu

jiayuasu:
> I can't trigger CI to run myself, so we'll have to wait for @jiayuasu

It's probably the time to make you a Sedona committer😊 @petern48

petern48:
This failure is unrelated.
```
   unknown package:
        Expected sha256 
6f1242f872dc33d3b4269dcaebf1838a359f9097e1cc848b0e11367bce010e4d
             Got        
a8d15178943c4e5de93d1b057c060bcaf9f3f3e37723e5c6bbac5719472c8926
```
I filed https://github.com/apache/sedona/issues/2388

petern48:
But look at that. Glad I caught it in https://github.com/apache/sedona/pull/2386#discussion_r2422367836
```
FAILED tests/geopandas/test_match_geopandas_series.py::TestMatchGeopandasSeries::test_is_closed - AssertionError: Series are different

Series values are different (100.0 %)
[index]: [0, 1, 2]
[left]:  [True, True, True]
[right]: [False, False, False]
```

We can fix this using a SQL CASE statement.
```
SELECT
    CASE
        WHEN ST_GeometryType(column) = 'ST_LINESTRING` THEN ST_Closed(column)
        ELSE false
    END
```
`ST_GeometryType()` should return 'ST_LINESTRING` instead of `ST_LINEARRING`, so there's no need to specify `LINEARRING`.

Could you try translating this fix to Spark @yunchipang? You can look at `boundary` as an example of how to implement a similar case statement. Feel free to ping if you get stuck.

https://github.com/apache/sedona/blob/6dc39e753b9d2087d27d2a6072b95d3eda91c53f/python/sedona/spark/geopandas/geoseries.py#L1024-L1033

python/sedona/spark/geopandas/geoseries.py
```diff
@@ -976,7 +976,12 @@ def is_ccw(self):
 
     @property
     def is_closed(self):
-        spark_expr = stf.ST_IsClosed(self.spark.column)
+        # Only check LineStrings; return False for all other geometry types
+        spark_expr = F.when(
+            stf.GeometryType(self.spark.column).isin(["LINESTRING"]),
```
yunchipang: @petern48 hope i understood it correctly

petern48:
```suggestion
            stf.ST_GeometryType(self.spark.column) == "ST_LineString",
```

Yep, you got it! Just one thing (that I failed to make clearer). We actually want to use `ST_GeometryType` here instead of `GeometryType`. This issue https://github.com/apache/sedona/issues/2389 explains why using `GeometryType` would fail a few edge cases. My suggestion also changes `.isin([])` to `==`, which isn't necessary but better style since we only need to compare to a single string here, not a list.

I realize the `boundary()` example I gave made both of those "mistakes" too, so that's my bad for not clarifying that. You got the main gist of the PR anyway. Nice job!

petern48:
Just in case you didn't know, you can set up `pre-commit` to run automatically after each of your local commits by running `pre-commit install` from the root dir.

Super **optional**. Some people don't like it since it takes some time to run. I personally do use it, since I *always* forget to run it myself. Zero problem if you don't want to, but I figured I would mention it as an FYI.

PR #2416, opened 2025-10-21
[GH-2389] replace `GeometryType()` with `ST_GeometryType()`

### LINKED ISSUE #2389: bug(Geopandas): Replace all uses of `GeometryType()` with `ST_GeometryType()`
In some functions, like below, we use `GeometryType()` to check which datatype a geometry is. I just realized that `GeometryType` also specifies `M` dimension, if it exists in the input (e.g `POINTM` instead of just `POINT`), so the string won't match properly in the `=` comparison, if the geometry has an `M` dimension. The easiest way to fix this is probably to use `ST_GeometryType()` instead (which returns the exact same string regardless of the dimensionality). Slight difference to note is that it adds a `ST_` prefix to the output and it's not in all caps, so we need to update the expected string accordingly (e.g, change `GEOMETRYCOLLECTION` to `ST_GeometryCollection`)

https://github.com/apache/sedona/blob/6dc39e753b9d2087d27d2a6072b95d3eda91c53f/python/sedona/spark/geopandas/geoseries.py#L1024-L1033

e.g 

```sql
SELECT ST_GeometryType(ST_GeomFromText('GEOMETRYCOLLECTION M (POINT M (1 2 3))'));
-- ST_GeometryCollection
SELECT GeometryType(ST_GeomFromText('GEOMETRYCOLLECTION M (POINT M (1 2 3))'));
-- GEOMETRYCOLLECTIONM  (notice the M at the end)
```

Notice the second output has an `M` suffix since it has an `M` dimension. Whether it's in caps or has a `ST_` prefix doesn't matter since it's consistent for all queries.

yunchipang:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2389

Replaces all uses of `GeometryType()` with `ST_GeometryType()`, because `GeometryType()` appends dimension suffixes (e.g., `"GEOMETRYCOLLECTIONM"`) that break string comparisons, while `ST_GeometryType()` returns consistent strings regardless of dimensionality.

## How was this patch tested?

test added

### DIFF
--- a/python/tests/geopandas/test_geoseries.py
+++ b/python/tests/geopandas/test_geoseries.py
@@ -1608,7 +1608,7 @@ e": "Feature", "properties": {}, "geometry": {"type": "Point", "coordinates": [3
         line = LineString([(0, 0), (1, 1)])
         result = s.crosses(line)
         expected = pd.Series([None, False])
-        assert_series_equal(result.to_pandas(), expected)
+        self.check_pd_series_equal(result, expected)
 
     def test_disjoint(self):
         pass


python/tests/geopandas/test_geoseries.py
```diff
@@ -1462,6 +1462,37 @@ def test_crosses(self):
         df_result = s.to_geoframe().crosses(s2, align=False).to_pandas()
         assert_series_equal(df_result, expected)
 
+    def test_geometry_type_with_m_dimension(self):
```
petern48: Personally, I think it's better to test each of the functions in their own respective test functions, rather than lump than all together. How about we remove this test and add tests to the existing `test_crosses`, `test_length`, and `test_boundary`? We can write slightly more precise tests and there's a small correctness bug in this one causing [the CI failure](https://github.com/apache/sedona/actions/runs/18671518097/job/53233436054?pr=2416), anyways.

python/tests/geopandas/test_geoseries.py
```diff
@@ -1462,6 +1462,37 @@ def test_crosses(self):
         df_result = s.to_geoframe().crosses(s2, align=False).to_pandas()
         assert_series_equal(df_result, expected)
```
petern48, line 1597:
```suggestion

        # Sedona ST_Crosses doesn't support GeometryCollection, so it returns NULL for now.
        # https://github.com/apache/sedona/issues/2417
        # Once this is resolved, we can update this expected result of this test.
        # Ensure M-dimension doesn't break things.
        s = GeoSeries(
            [
                wkt.loads("GEOMETRYCOLLECTION M (POINT M (1 2 3))"),
                wkt.loads("LINESTRING M (0 0 1, 1 1 2)"),
            ]
        )
        line = LineString([(0, 0), (1, 1)])
        result = s.crosses(line)
        expected = pd.Series([None, False])
        assert_series_equal(result.to_pandas(), expected)

```

I essentially copied the original test you had below to here (and made the expected results more precise). ST_Crosses is weird tbh, since it has a workaround. I filed https://github.com/apache/sedona/issues/2417 and mentioned it in the comments.

python/tests/geopandas/test_geoseries.py
petern48, line 1207:
```suggestion

        # Ensure M-dimension doesn't break things.
        s = GeoSeries(
            [
                wkt.loads("GEOMETRYCOLLECTION M (POINT M (1 2 3))"),
            ]
        )
        result = s.boundary
        expected = gpd.GeoSeries(
            [
                None,
            ]
        )
        self.check_sgpd_equals_gpd(result, expected)

```

Again just replicating your test from below

petern48, line 676:
```suggestion

        s = GeoSeries(
            [
                wkt.loads("POINT M (0 0 0)"),
                wkt.loads("LINESTRING M (0 0 0, 1 1 0)"),
                wkt.loads("POLYGON M ((0 0 0, 1 0 0, 1 1 0, 0 0 0))"),
                wkt.loads(
                    "GEOMETRYCOLLECTION M (POINT M (0 0 0), LINESTRING M (0 0 0, 1 1 0), POLYGON M ((0 0 0, 1 0 0, 1 1 0, 0 0 0)))"
                ),
            ]
        )
        result = s.length.to_pandas()
        expected = pd.Series([0.000000, 1.414214, 3.414214, 4.828427])
        assert_series_equal(result, expected)

```

For `.length()`, I played around with it locally and realize I'm not super confident about 3D support right now. I propose we simply take the existing `.length()` test and add a m-dimension of `0` to each of them, so that it's basically a 2d computation. Supporting / testing 3D support is out of scope for this PR.

petern48 (commented):
Thanks for picking this up. I played with this locally myself and realizing testing M-dimensions isn't great at the moment. I suggested some changes, but they mostly re-use the same test cases you originally proposed.

petern48:
Last thing, I'll ask for you to do, could we add just one more case for `test_is_closed`? In a [previous PR](https://github.com/apache/sedona/pull/2387) of yours, we switched from using `GeometryType` to `ST_GeometryType` as well (https://github.com/apache/sedona/pull/2387/commits/9daec30509bb7517ea5543958613ad932f06373c).

https://github.com/apache/sedona/blob/9ea7c23633cf97b30a662465aa061feecfdaa79c/python/sedona/spark/geopandas/geoseries.py#L995-L999

I didn't ask you to add an `M`-dimension test that time since it felt out-of-scope, but this would be a good time to add it now. Ideally, it would be a test case involving a `LINESTRING M` that should result in `True`, since that's what would fail if we used `GeometryType` instead (since it would return `LINESTRINGM` and not match with the. Similar to what I proposed above for `.length()`, you could simply take one of the existing cases that return `True` and add an `M` dimension of `0` to each of the points.

petern48:
Hi @yunchipang, were you planning to address the PR feedback? This contribution would still be very much appreciated 🙏 . The suggestions I left before should get you very near the end. (I believe I also tested those locally, so you won't have to deal with those failing)

yunchipang:
@petern48 thanks for the ping! i've been a bit tied up recently, but I’ll get to fixing those soon. if this is blocking you, please feel free to take it over — sorry about the delay and thanks for the reviews!🙏🏻

yunchipang:
@petern48 suggested fix has been addressed! i could figure out the error CI is throwing could you take a quick look? thanks!

python/tests/geopandas/test_geoseries.py
```diff
@@ -677,6 +677,21 @@ def test_length(self):
...
+                wkt.loads("LINESTRING M (0 0 0, 1 1 0)"),
+                wkt.loads("POLYGON M ((0 0 0, 1 0 0, 1 1 0, 0 0 0))"),
+                wkt.loads(
+                    "GEOMETRYCOLLECTION M (POINT M (0 0 0), LINESTRING M (0 0 0, 1 1 0), POLYGON M ((0 0 0, 1 0 0, 1 1 0, 0 0 0)))"
+                ),
+            ]
+        )
+        result = s.length.to_pandas()
+        expected = pd.Series([0.000000, 1.414214, 3.414214, 4.828427])
+        assert_series_equal(result, expected)
```
```suggestion
        result = s.length
        expected = pd.Series([0.000000, 1.414214, 3.414214, 4.828427])
        self.check_pd_series_equal(result, expected)
```

python/tests/geopandas/test_geoseries.py
```diff
@@ -1462,6 +1500,21 @@ def test_crosses(self):
...
+        s = GeoSeries(
+            [
+                wkt.loads("GEOMETRYCOLLECTION M (POINT M (1 2 3))"),
+                wkt.loads("LINESTRING M (0 0 1, 1 1 2)"),
+            ]
+        )
+        line = LineString([(0, 0), (1, 1)])
+        result = s.crosses(line)
+        expected = pd.Series([None, False])
+        assert_series_equal(result.to_pandas(), expected)
```
```suggestion
        self.check_pd_series_equal(result, expected)
```

petern48 (commented):
```
assert_series_equal(result, expected)
NameError: name 'assert_series_equal' is not defined
```
[CI Fail](https://github.com/apache/sedona/actions/runs/19353224437/job/55369322969?pr=2416)

Ah yeah, this can definitely be confusing. It's good that you asked. We recently did some cleaning up to clean up our tests, in this a recent [issue/PR](https://github.com/apache/sedona/issues/2379) that essentially replaces all `assert_series_equal` with `self.check_pd_series_equal`. Notably it removed the import for `assert_series_equal`, which the tests in this PR relies on. If you run the tests locally, the tests pass. But once this branch is merged with main, the import is deleted and the tests fail. Here's how we should fix it:

1. merge main into this branch (this should remove most of the `assert_series_equal` calls in the file)
2. replace all `assert_series_equal(result.to_pandas(), expected)` with `self.check_pd_series_equal(result, expected)` instead. Notice we remove the `.to_pandas()` call here too along with the replacement.

*Technically merging in main shouldn't quite be necessary since there are no conflicts. But doing so will reveal the issue more clearly and allow you to reproduce the failed tests locally. In this case, there are only a few changes we need to make. I expect things to pass once the following suggestions are applied.

Issue #2419, opened 2025-10-22
Support creating permanent views in Sedona and Spark


zhangfengcdt:
Currently, we cannot create permanent views if the view uses Sedona expressions, for example,

```
      CREATE VIEW nyc_buildings_with_functions AS
      SELECT * FROM (
        SELECT
          nyc_buildings_geom_test.PROP_ADDR AS name,
          nyc_buildings_geom_test.geom AS building_geom,
          avg(elevation_zones_test.elevation) AS elevation
        FROM
          nyc_buildings_geom_test
        JOIN
          elevation_zones_test
        ON
          st_intersects(elevation_zones_test.zone_geom, elevation_zones_test.zone_geom)
        GROUP BY
          nyc_buildings_geom_test.PROP_ADDR, nyc_buildings_geom_test.geom
      )
      WHERE elevation > 0

```
It fails with the following errors:

[INVALID_TEMP_OBJ_REFERENCE] Cannot create the persistent object `spark_catalog`.`default`.`nyc_buildings_with_functions` of the type VIEW because it references to the temporary object `st_intersects` of the type FUNCTION. Please make the temporary object `st_intersects` persistent, or make the persistent object `spark_catalog`.`default`.`nyc_buildings_with_functions` temporary.
org.apache.spark.sql.AnalysisException: [INVALID_TEMP_OBJ_REFERENCE] Cannot create the persistent object `spark_catalog`.`default`.`nyc_buildings_with_functions` of the type VIEW because it references to the temporary object `st_intersects` of the type FUNCTION. Please make the temporary object `st_intersects` persistent, or make the persistent object `spark_catalog`.`default`.`nyc_buildings_with_functions` temporary.
	at org.apache.spark.sql.errors.QueryCompilationErrors$.notAllowedToCreatePermanentViewByReferencingTempFuncError(QueryCompilationErrors.scala:2891)

We will need to fix this so that users can create permanent (not temp) views like this.

PR #2420, opened 2025-10-22
[GH-2419] Register Sedona functions as built-in functions to support permanent VIEW creation

zhangfengcdt:

Sedona functions are registered as temporary functions in the session `FunctionRegistry`, which prevents creating permanent VIEWs that reference them:

  ```sql
  CREATE VIEW my_view AS
  SELECT ST_AsText(geom) FROM buildings;
```

  -- Error: Cannot create view that references temporary function: ST_AsText

Spark's CreateViewCommand validates that permanent VIEWs cannot reference temporary functions. Since Sedona functions are only registered in the session registry (not the built-in registry), they are treated as temporary.

This PR registers Sedona functions into Spark's built-in function registry (FunctionRegistry.builtin) in addition to the session registry. This makes Spark treat Sedona functions as built-in functions (similar to COUNT, MAX, SUM) rather than temporary functions, bypassing the CREATE VIEW validation.

## How was this patch tested?
 The new test validates:
  - Permanent VIEW creation with Sedona spatial functions
  - VIEW queries return correct results
  - Proper assertions instead of console output

### DIFF
--- a/spark/common/src/main/scala/org/apache/sedona/sql/UDF/AbstractCatalog.scala
+++ b/spark/common/src/main/scala/org/apache/sedona/sql/UDF/AbstractCatalog.scala
@@ -92,8 +92,15 @@ abstract class AbstractCatalog {
         expressionInfo,
         functionBuilder)
     }
-    aggregateExpressions.foreach(f =>
-      sparkSession.udf.register(f.getClass.getSimpleName, functions.udaf(f)))
+    aggregateExpressions.foreach { f =>
+      sparkSession.udf.register(f.getClass.getSimpleName, functions.udaf(f))
+      FunctionRegistry.builtin.registerFunction(
+        FunctionIdentifier(f.getClass.getSimpleName),
+        new ExpressionInfo(f.getClass.getCanonicalName, null, f.getClass.getSimpleName),
+        (_: Seq[Expression]) =>
+          throw new UnsupportedOperationException(
+            s"Aggregate function ${f.getClass.getSimpleName} cannot be used as a regular function"))
+    }
   }
 
   def dropAll(sparkSession: SparkSession): Unit = {
--- a/spark/common/src/test/scala/org/apache/sedona/sql/functionTestScala.scala
+++ b/spark/common/src/test/scala/org/apache/sedona/sql/functionTestScala.scala
@@ -4182,11 +4182,6 @@ class functionTestScala
       WHERE elevation > 0
     """)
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: spark/common/src/main/scala/org/apache/sedona/sql/UDF/AbstractCatalog.scala [hunk 1]
```diff
@@ -86,6 +87,10 @@ abstract class AbstractCatalog {
         functionIdentifier,
         expressionInfo,
         functionBuilder)
+      FunctionRegistry.builtin.registerFunction(
```
jiayuasu, line 90: Can you confirm if this gets each function registered twice? (once for temp function, one for built-in)?

How about `aggregateExpressions?`

zhangfengcdt: It looks like we need to register to both the `sessionState.functionRegistry` and the buildin simple function registry. They server different purpose, and if we remove the first one, functions cannot be used any more.

I have added the three aggregate functions to buildin registry as well and also new test on creating views on spatial aggregation queries.

spark/common/src/main/scala/org/apache/sedona/sql/UDF/AbstractCatalog.scala
```diff
@@ -86,9 +87,20 @@ abstract class AbstractCatalog {
...
+        functionIdentifier,
+        expressionInfo,
+        functionBuilder)
+    }
+    aggregateExpressions.foreach { f =>
+      sparkSession.udf.register(f.getClass.getSimpleName, functions.udaf(f))
+      FunctionRegistry.builtin.registerFunction(
```
haizhou-zhao, line 97: For folks in the industry who're using HiveMetastore, IcebergRestCatalog or other table metastore options: might be a good idea to leave an option open for them to register geospatial related udf with various metastore options if they prefer to do things that way (i.e. one draw back I could think of with having too many buildin func is it complicates the resolving path for every command, whether or not they contain geospatial udf). Don't have to do this in this PR, but might be great to have a tracking issue or TODO in some form.

jiayuasu: @haizhou-zhao pleas feel free to create a github issue on Sedona and we can then track it

jiayuasu:
Thanks. Merging.

Issue #2379, opened 2025-10-10
chore(geopandas): Replace `assert_series_equal` calls with `pd.check_pd_series_equal()`


petern48:
While reviewing https://github.com/apache/sedona/pull/2378, I noticed there are some sprinkles of `assert_series_equal()` calls in the test files from earlier development. We should instead use the `self.check_pd_series_equal()` wrapper because it validates that the result of our Geopandas functions is `ps.Series` (pyspark series) and not `pd.Series` (pandas series).

https://github.com/apache/sedona/blob/6cc4f6f49b0de9fe1d5a59eea41aa4db801815e7/python/tests/geopandas/test_geopandas_base.py#L106-L109

Usually, the `to_pandas()` call is inlined, but sometimes it occurs on another line like below. We should update these cases, too.
https://github.com/apache/sedona/blob/56f196988fffb0f3fed0bf1c0f4153d41b976a6c/python/tests/geopandas/test_geoseries.py#L672-L674

Definition of Done:
Running the following grep command should ideally only return the call in the `check_pd_series_equal` function (and the import). There's definitely a lot more than that at the moment. If we can't replace every call, then it's a bad sign that maybe we're returning a non-scalable `pd.Series` instead of the spark version.

```bash
grep -r "assert_series_equal" sedona/python/tests/geopandas
```

Subham-KRLX:
hi I would like to work on this issue can you please assign this issue to me.

petern48:
Just did, go for it

PR #2421, opened 2025-10-22
Replace assert_series_equal with check_pd_series_equal in geopandas t…

Subham-KRLX:
This PR addresses issue #2379 by replacing all assert_series_equal calls in the python/tests/geopandas directory with the self.check_pd_series_equal wrapper method. This ensures the tests correctly validate PySpark Series (ps.Series) instead of pandas Series (pd.Series). All related test cases have been updated accordingly, and existing .to_pandas() calls are maintained for proper comparison.

### DIFF
--- a/python/tests/geopandas/test_geoseries.py
+++ b/python/tests/geopandas/test_geoseries.py
@@ -166,7 +166,7 @@ class TestGeoSeries(TestGeopandasBase):
 
         # Test that GeoDataFrame.area also works
         df_result = self.geoseries.to_geoframe().area
-        self.check_pd_series_equal(result, df_result)
+        self.check_pd_series_equal(result, df_result.to_pandas())
 
     def test_buffer(self):
 


python/tests/geopandas/test_geoseries.py
```diff
@@ -163,7 +163,7 @@ def test_plot(self):
     def test_area(self):
         result = self.geoseries.area.to_pandas()
         expected = pd.Series([0.0, 0.0, 5.23, 5.23])
-        assert_series_equal(result, expected)
+        self.check_pd_series_equal(result, expected)
```
```suggestion
        result = self.geoseries.area
        expected = pd.Series([0.0, 0.0, 5.23, 5.23])
        self.check_pd_series_equal(result, expected)
```

python/tests/geopandas/test_geoseries.py
```diff
@@ -235,23 +235,23 @@ def test_x(self):
         )
         result = geoseries.x.to_pandas()
```
```suggestion
        result = geoseries.x
```

python/tests/geopandas/test_geoseries.py
petern48: Here's what the overall change each time should look like:
1. Remove the `.to_pandas()` call.
2. Replace `assert_series_equal` with `self.check_pd_series_equal`.

python/tests/geopandas/test_geoseries.py
petern48: Since you replaced the `assert_series_equal()` function below already, we just need to remove the `.to_pandas()` call. Now, please replicate this pattern of changes for the rest.

python/tests/geopandas/test_geoseries.py
```diff
@@ -716,7 +716,7 @@ def test_is_valid_reason(self):
                 None,
             ]
         )
-        assert_series_equal(result, expected)
+        self.check_pd_series_equal(result, expected)
```
petern48: Careful, there are also some cases like this, where the `.to_pandas()` call is higher up and easy to miss. We should remove the `.to_pandas()` call here too so that `self.check_pd_series_equal()` passes.

petern48 (commented):
Thanks for picking this up. Looks like there was a slight misunderstanding, so I'll clarify here. As you can see in the function definition below, `self.check_pd_series_equal` takes a `ps.Series` object, not a `pd.Series` object, as the first argument.

https://github.com/apache/sedona/blob/6cc4f6f49b0de9fe1d5a59eea41aa4db801815e7/python/tests/geopandas/test_geopandas_base.py#L106-L109

`to_pandas()` converts `ps.Series` to `pd.Series`. So, in addition to replacing the function call, we should remove the `.to_pandas()` call as well.

It would be helpful for you to set up your local dev environment, so you can run the tests locally and see if they pass. Instructions for doing that are here in the [compile docs](https://github.com/apache/sedona/blob/master/docs/setup/compile.md). It should work pretty smoothly, but if it doesn't, please don't be afraid to ask for help.

I've included suggestions below to show how to make the changes. Could you replicate these fixes throughout the rest of the file?

Subham-KRLX:
@petern48 Thank you very much for the helpful feedback  I am currently working on a different pull request and also preparing for my exams so I will need some additional time to address these changes properly.

petern48:
Looks like we're currently failing the pre-commit. Generally, this just means a small formatting error.

image

I recommend setting up `pre-commit` to automatically run on each commit locally. You can do this by running `make checkinstall`. Or equivalently `pip install pre-commit` and `pre-commit install`.

Then to trigger the fix, you can either push a new commit (literally anything) or manually run `pre-commit run --all-files`.

I'm not sure what's up with the [doc build failure](https://github.com/apache/sedona/actions/runs/19204359251/job/54897463011?pr=2421), let's just not worry about it. We'll see if it happens again after the next commit. I'd guess it's caused by something outside of this PR.

python/tests/geopandas/test_geoseries.py
```diff
@@ -112,7 +111,7 @@ def test_to_spark_pandas(self):
         crs = "EPSG:3857"
         result = GeoSeries(data, index=index, crs=crs).to_spark_pandas()
         ps_df = ps.Series(data, index=index)
-        assert_series_equal(result.to_pandas(), ps_df.to_pandas())
+        self.check_pd_series_equal(result, ps_df)
```
```suggestion
        self.check_pd_series_equal(result, ps_df.to_pandas())
```

Got to be careful here. This is what caused the test failure. `self.check_pd_series_equal()` takes a `ps.Series` for the first arg and a `pd.Series` for the second. Here, by removing the `.to_pandas()` call from both, you tried inputting a `ps.Series` and `ps.Series`.

I suggested a change that should fix it. Again, I recommend setting up your local dev environment using the link I sent above and running the tests locally (at least the tests that you changed). For more interesting PRs, you really wouldn't have a good time having to wait for a maintainer to trigger CI to run to see if your tests pass.

petern48:
I ran a grep and saw a few more places we can clean this up. Could you also remove these `assert_series_equal` imports in the following files: `test_match_geopandas_series.py` and `test_geodataframe.py`. They're also not used.

https://github.com/apache/sedona/blob/0b061e3b153885530943b1e943d7863e8923cc03/python/tests/geopandas/test_match_geopandas_series.py#L25

https://github.com/apache/sedona/blob/0b061e3b153885530943b1e943d7863e8923cc03/python/tests/geopandas/test_geodataframe.py#L40

Once these are removed, the tests pass, and `pre-commit` passes, this PR should be ready to be merged.

petern48:
I'm logging off for the night after approving this last workflow. In case you haven't already, again I recommend following the instructions I mentioned above [here](https://github.com/apache/sedona/pull/2421#issuecomment-3507574692), to make it automatically reformats your code on each commit. It's easy to do, and you only need to do it once, and it will save you a lot of time avoiding failed CI runs for trivial formatting reasons.

Subham-KRLX:
Thank you so much @petern48 for all your help and patience! I really appreciate the guidance.
I will definitely set up pre-commit locally following your instructions to avoid formatting issues in future PRs.

petern48 (approved):
Alright, CI passes (ignoring the unrelated docs build failure). Thanks for doing this small improvement to the test suite.

petern48:
Went ahead and merged, since the docs CI failure is surely unrelated. Thanks for contributing to Sedona Geopandas @Subham-KRLX 👏

PR #2447, opened 2025-10-24
[GH-2406] Make ST_Distance return null instead of 0.0 when arg is empty geom

### LINKED ISSUE #2406: bug: ST_Distance involving an empty geometry returns 0.0 instead of null
Sedona:
```sql
select st_distance(st_geomfromtext('point empty'), st_geomfromtext('multipoint empty'))
-- 0.0
```

PostGIS
```sql
select st_distance(st_geomfromtext('point empty'), st_geomfromtext('multipoint empty'))
-- null
```

Same results if only one geometry is empty
```sql
select st_distance(st_geomfromtext('point empty'), st_geomfromtext('multipoint empty'))
-- 0.0 (Sedona)
-- null (PostGIS)
```

petern48:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2406

Make ST_Distance return null instead of 0.0 when either arg is null. This is the same behavior as PostGIS

## How was this patch tested?
Added test. Removed geopandas workaround

- No, this is a niche edge case that was undocumented in the first place.

### DIFF
--- a/spark/common/src/test/scala/org/apache/sedona/sql/SpatialJoinSuite.scala
+++ b/spark/common/src/test/scala/org/apache/sedona/sql/SpatialJoinSuite.scala
@@ -269,7 +269,10 @@ class SpatialJoinSuite extends TestBaseScala with TableDrivenPropertyChecks {
       }
     }
 
-    it("ST_Distance join involving empty geometries should evaluate to false") {
+    it("ST_Distance involving empty geometries should work as a predicate") {
+      // ST_Distance returns null when either arg is an empty geometry,
+      // while this test doesn't involve an actual spatial join, it tests that
+      // a distance-based spatial join doesn't fail due to this edge case.
       val result1 = sparkSession.sql(
         "SELECT * FROM df1 WHERE ST_Distance(df1.geom, ST_GeomFromText('POINT EMPTY')) < 1")
       assert(result1.count() == 0)


jiayuasu (commented):
Have you tested this in distance join? 

Like putting in Where clause ST_Distance(geoCol1, geoCol2) < 10

spark/common/src/test/scala/org/apache/sedona/sql/SpatialJoinSuite.scala
```diff
@@ -268,6 +268,15 @@ class SpatialJoinSuite extends TestBaseScala with TableDrivenPropertyChecks {
         assert(resultRows.isEmpty)
       }
     }
+
+    it("ST_Distance join involving empty geometries should evaluate to false") {
+      val result1 = sparkSession.sql(
```
jiayuasu, line 276: This is not a spatial join. You need to have 2 columns in the predicate, not a constant

petern48: I know, this *technically* isn't a spatial join (so yeah, maybe I can change the test name), but isn't this sufficient for what we want to test? If I were to write a test for this behavior that **is technically** a spatial join. I would join an existing df with a table with all empty geometries. That would be equivalent to just testing against a constant empty geometry, wouldn't it? Doing this was just more convenient since an empty geom df didn't already exist.

Personally, I didn't find a need to test distance joins. We're just modifying the behavior of a predicate. This shouldn't change anything in the spatial join outside of whether the predicate evaluates to true or not.

`null < 1` -> `null`. And a result of `null` is treated as `False` for joining purposes. It evaluates to something successfully and doesn't error. Same behavior as PostGIS. Should be simple as that, right? I don't see any other way for things to fail.

If you disagree, is there a certain case you'd be concerned about? Or do you have a suggestion for how to test instead?

jiayuasu: I agree. This PR just simply changes the behavior of a predicate. The current tests would suffice. We can keep it this way, and probably you shouldn't call it spatial join.

I think I am just curious how the current [spatial join logic](https://jiayuasu.github.io/files/paper/GeoSpark_Geoinformatica_2018.pdf) (distance join and range join) handles `EMPTY` geometries. Our spatial joins go through several phases (1) distributed spatial partitioning (if the data is too large for a broadcast join) (2) spatial index building (3) filtering (4) predicate eval. IIRC, empty geometries probably get pruned out at Phase 1 or Phase 2. So the ST_Distance eval bug probably didn't surface before.

petern48: Hmm, that's a good point. I now see that the difference in optimizer rules could make an actual spatial join different than this test. I'm not sure why it was working in the past. Without digging through the code, I would guess maybe there might be a rule somewhere that swaps this out for a ST_DWithin join? Looking at [the dwithin implementation](https://github.com/apache/sedona/blob/1eb5769be3b1ab7266a9dcecc6e3407ba28ab57a/common/src/main/java/org/apache/sedona/common/Predicates.java#L70-L82), it uses JTS's distance() and not the Functions.distance() that I modified here. If this is the case, it is possible the distance join is buggy. It is odd to me since `(0.0 < 1) => True`, while `(null < 1) => null => False`.

Regardless, I think for now I'd rather wrap up this PR and potentially look into this later. It at least doesn't seem to be a regression. I filed [this issue](https://github.com/apache/sedona/issues/2450) to check if we're getting the same results as geopandas (which I know is not quite postgis, but it's a good easy starting point)

Issue #2465, opened 2025-10-30
Log4j Class Initialization Deadlock in SedonaKryoRegistrator Class


joonaspessi:
# Log4j Class Initialization Deadlock in KryoSedonaRegistrator Class

## Problem Description

I'm experiencing occasional deadlocks in AWS EMR 7.8.0 Spark clusters that cause executors to hang indefinitely during JVM initialization. After investigation with Java thread dumps, I have identified that the static logger initialization in `SedonaKryoRegistrator` is one of the contributing factors to this deadlock. This deadlock seems to happen extremely rarely but still there seems to be possibility for it as I have seen it now with multiple Spark clusters.

## Environment

- **Platform:** AWS EMR 7.8.0
- **Spark Version:** 3.5.4-amzn-0
- **Sedona Version:** 1.7.0
- **Log4j Version:** log4j-1.2-api-2.20.0 (bundled and delivered with spark 3.5.x)
- **Cluster Configuration:** Spark cluster with 500 Executors where each executor have 4 cores. (So there are quite many JVM instances where this problem can occur)
- **Java version:** Amazon Corretto 17
- **Architecture:** ARM64/Graviton

## What Happens

The deadlock occurs during Spark executor JVM startup due to a race condition between multiple threads concurrently initializing Log4j classes (log4j-1.2-api-2.20.0). The circular class initialization dependency chain is:

```
Priority → Level (requires Level class)
Level extends Priority (requires Priority class)
LogManager → Level.DEBUG (requires both Level and Priority)
```

### Thread Dump Analysis

When the deadlock occurs, we observe two threads stuck:

**Thread # 59: "Executor task launch worker for task 381.0 in stage 3.0 (TID 383)"**
```
java.lang.Thread.State: RUNNABLE
at org.apache.log4j.LogManager.<clinit>(LogManager.java:72)
- waiting on the Class initialization monitor for org.apache.log4j.Priority
at org.apache.log4j.Logger.getLogger(Logger.java:35)
at org.apache.sedona.core.serde.SedonaKryoRegistrator.<clinit>(SedonaKryoRegistrator.java:40)
at java.lang.Class.forName0(java.base@17.0.15/Native Method)
at java.lang.Class.forName(java.base@17.0.15/Class.java:467)
at org.apache.spark.util.SparkClassUtils.classForName(SparkClassUtils.scala:43)
at org.apache.spark.util.Utils$.classForName(Utils.scala:96)
[...]

PR #2467, opened 2025-10-31
[GH-2465] Log4j Class Initialization Deadlock in SedonaKryoRegistrator Class. Closes issue #2465

joonaspessi:

Spark 3.x is currently including log4j2 version `2.20.0`, that can cause deadlock on Spark executor initialization.
This PR removes logger initialization from Kryo registrators to mitigate possible risk of deadlock. The downside is that there's no easy way to debug anymore Kryo registration, but this downside should be still acceptable than using Classes that are not thread safe and potentially cause Deadlock in Spark executor initialization..

For further information about the deadlock see https://github.com/apache/sedona/issues/2465

## How was this patch tested?

Code complies, tests pass and  pre-commit passes.


jiayuasu:
Thank you!

PR #2502, opened 2025-11-14
Implement minimum_bounding_radius

### LINKED ISSUE #2495: Geopandas: Implement minimum_bounding_radius
### Use case:

Add support for GeoSeries.minimum_bounding_radius

### Description:

Implements GeoSeries.minimum_bounding_radius, which computes the radius of the smallest enclosing circle (minimum bounding circle) for each geometry.

### Related Issue:

Links to epic https://github.com/apache/sedona/issues/2230

chay0112:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2495

- Implements minimum_bounding_radius

## How was this patch tested?

- Included Unit and Parity tests

### DIFF
--- a/python/sedona/spark/geopandas/base.py
+++ b/python/sedona/spark/geopandas/base.py
@@ -613,38 +613,48 @@ class GeoFrame(metaclass=ABCMeta):
 
     @property
     def convex_hull(self):
-        """
-        Return the convex hull of each geometry.
+        """Return a ``GeoSeries`` of geometries representing the convex hull
+        of each geometry.
 
-        The convex hull is the smallest convex Polygon that contains
-        all the points of the geometry.
+        The convex hull of a geometry is the smallest convex `Polygon`
+        containing all the points in each geometry, unless the number of points
+        in the geometric object is less than three. For two points, the convex
+        hull collapses to a `LineString`; for 1, a `Point`.
 
         Examples
         --------
-        >>> from shapely.geometry import Point, Polygon, LineString
         >>> from sedona.spark.geopandas import GeoSeries
+        >>> from shapely.geometry import Polygon, LineString, Point, MultiPoint
         >>> s = GeoSeries(
         ...     [
-        ...         Polygon([(0, 0), (1, 0), (1, 1), (0, 1)]),
-        ...         LineString([(0, 0), (2, 1)]),
+        ...         Polygon([(0, 0), (1, 1), (0, 1)]),
+        ...         LineString([(0, 0), (1, 1), (1, 0)]),
+        ...         MultiPoint([(0, 0), (1, 1), (0, 1), (1, 0), (0.5, 0.5)]),
+        ...         MultiPoint([(0, 0), (1, 1)]),
         ...         Point(0, 0),
         ...     ]
         ... )
         >>> s
-        0    POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0))
-        1    LINESTRING (0 0, 2 1)
-        2    POINT (0 0)
+        0                       POLYGON ((0 0, 1 1, 0 1, 0 0))
+        1                           LINESTRING (0 0, 1 1, 1 0)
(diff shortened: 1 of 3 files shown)


python/tests/geopandas/test_match_geopandas_series.py
```diff
@@ -781,7 +781,15 @@ def test_minimum_bounding_circle(self):
             self.check_sgpd_equals_gpd(sgpd_result, gpd_result, tolerance=0.5)
 
     def test_minimum_bounding_radius(self):
-        pass
+        # minimum_bounding_radius was added from geopandas 1.0.0 and later
+        if parse_version(gpd.__version__) < parse_version("1.0.0"):
+            pytest.skip(
+                "geopandas minimum_bounding_radius requires version 1.0.0 or higher"
+            )
```
```suggestion
```

It was added much earlier, in version 0.13: See [here](https://geopandas.org/en/stable/docs/changelog.html#:~:text=Added%20minimum_bounding_radius()%20as%20GeoSeries%20method%20(%232827).). Avoid blindly copying `pytest.skips`. We should avoid them by default and only include them if necessary to pass tests (and also for a good reason).

python/tests/geopandas/test_geoseries.py
```diff
@@ -1309,7 +1309,30 @@ def test_minimum_bounding_circle(self):
...
+        )
+
+        expected = gpd.GeoSeries(
+            [
+                Polygon([(0, 0), (1, 0), (1, 1), (0, 1)]),
+                LineString([(0, 0), (2, 0)]),
+                Point(0, 0),
+                None,
+            ]
+        ).minimum_bounding_radius()
```
petern48: This should be a `pd.Series([])` of the floats.

`test_geoseries.py` is for hard-coding the expected output. 
`test_match_geopandas_series.py` is for comparing if the result is the same as the original geopandas, which is what you did here.

Also slight preference that we use the examples in the doc string in `base.py`.

python/sedona/spark/geopandas/base.py
```diff
@@ -742,8 +742,39 @@ def minimum_bounding_circle(self):
...
+        ...     [
+        ...         Point(0, 0),
+        ...         LineString([(0, 0), (1, 1)]),
+        ...         Polygon([(0, 0), (3, 0), (3, 3), (0, 3)]),
+        ...     ]
+        ... )
+        >>> s.minimum_bounding_radius()
+        0    0.000000
+        1    ...
+        2    ...
```
petern48: We can't just say the answer is "`...`". The whole point of the docstring is to explain what the behavior is. We should be **copy-pasting the docstring directly from geopandas** (and then making small tweaks to make it import sedona instead of geopandas), as mentioned in instruction 4 of the [epic issue](https://github.com/apache/sedona/issues/2230).

Here's the corresponding docstring in geopandas: https://github.com/geopandas/geopandas/blob/503ace4b76c36a9b46fd9b9f85cd68ca046f1376/geopandas/base.py#L1917-L1947

If these were AI-generated, we can't rely on AI to generate the correct docstrings bc sometimes geopandas' behavior is non-intuitive, and LLMs won't get that. Please do also fix the docstrings for the other functions you implemented (in this PR is fine). Most of the docstrings should all be in that same [base.py](https://github.com/geopandas/geopandas/blob/503ace4b76c36a9b46fd9b9f85cd68ca046f1376/geopandas/base.py) file in the geopandas repo.

python/tests/geopandas/test_match_geopandas_series.py
chay0112: Ah yes, I relied on Github co-pilot for this, But I'll refer to the documentation you mentioned from next time. Thanks

python/sedona/spark/geopandas/base.py
chay0112: Sure, Just addressed this, copied as is from geopandas and made import changes little bit to match with sedona.

petern48 (approved):
Thanks for addressing feedback. Another great PR 🙌

PR #2512, opened 2025-11-20
[GH-2504] Geopandas: Implement force_3d

### LINKED ISSUE #2504: Geopandas: Implement force_3d
### Use case:

- Add support for GeoSeries.force_3d

### Description:

- Implements GeoSeries.force_3d, which converts each geometry in the series to a 3D geometry by adding a Z coordinate. If a geometry is already 3D, it is returned unchanged; otherwise, a default Z value (typically 0.0) is added to all coordinates. This ensures all geometries in the series have three dimensions.

### Related Issue:

- Links to epic https://github.com/apache/sedona/issues/2230

chay0112:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2504 

- Implement force_3d

## How was this patch tested?

### DIFF
--- a/python/sedona/spark/geopandas/base.py
+++ b/python/sedona/spark/geopandas/base.py
@@ -955,6 +955,10 @@ class GeoFrame(metaclass=ABCMeta):
         2D geometries will get the provided Z coordinate; 3D geometries
         are unchanged (unless their Z coordinate is ``np.nan``).
 
+        Note: Sedona's behavior may differ from Geopandas' for M and ZM geometries.
+        For M geometries, Sedona will replace the M coordinate and add the Z coordinate.
+        For ZM geometries, Sedona will drop the M coordinate and retain the Z coordinate.
+
         Parameters
         ----------
         z : float | array_like (default 0)
--- a/python/tests/geopandas/test_geoseries.py
+++ b/python/tests/geopandas/test_geoseries.py
@@ -1594,6 +1594,22 @@ e": "Feature", "properties": {}, "geometry": {"type": "Point", "coordinates": [3
         result = s.force_3d(z)
         self.check_sgpd_equals_gpd(result, expected)
 
+        # 4. Ensure M and ZM geometries are handled correctly
+        s = sgpd.GeoSeries(
+            [
+                shapely.wkt.loads("POINT M (1 2 3)"),
+                shapely.wkt.loads("POINT ZM (1 2 3 4)"),
+            ]
+        )
+        result = s.force_3d(7.5)
+        expected = gpd.GeoSeries(
+            [
+                shapely.wkt.loads("POINT Z (1 2 7.5)"),
+                shapely.wkt.loads("POINT Z (1 2 3)"),
+            ]
+        )
+        self.check_sgpd_equals_gpd(result, expected)
+
     def test_line_merge(self):
         pass
 
--- a/python/tests/geopandas/test_match_geopandas_series.py
+++ b/python/tests/geopandas/test_match_geopandas_series.py
@@ -883,7 +883,7 @@ class TestMatchGeopandasSeries(TestGeopandasBase):
             pytest.skip("geopandas force_3d requires version 1.0.0 or higher")
         # 1) Promote 2D to 3D with z = 4
         for geom in self.geoms:
-            if isinstance(geom[0], (LinearRing, GeometryCollection, MultiPolygon)):
+            if isinstance(geom[0], (LinearRing)):
                 continue
             sgpd_result = GeoSeries(geom).force_3d(4)
             gpd_result = gpd.GeoSeries(geom).force_3d(4)


python/sedona/spark/geopandas/base.py
```diff
@@ -949,8 +949,68 @@ def force_2d(self):
...
+            Z coordinate to be assigned
+
+        Returns
+        -------
+        GeoSeries
+
+        Examples
+        --------
+        >>> from shapely import Polygon, LineString, Point
+        >>> s = geopandas.GeoSeries(
```
```suggestion
        >>> from shapely import Polygon, LineString, Point
        >>> from sedona.spark.geopandas import GeoSeries
        >>> s = GeoSeries(
```
nit

python/sedona/spark/geopandas/base.py
```diff
@@ -949,8 +949,68 @@ def force_2d(self):
         """
         return _delegate_to_geometry_column("force_2d", self)
 
-    # def force_3d(self, z=0):
-    #     raise NotImplementedError("This method is not implemented yet.")
+    def force_3d(self, z=0):
+        """Force the dimensionality of a geometry to 3D.
+
+        2D geometries will get the provided Z coordinate; 3D geometries
+        are unchanged (unless their Z coordinate is ``np.nan``).
```
petern48, line 956: Can you add a test for this in `test_geoseries.py`? I see that this is from the original geopandas docstring, so I'm not entirely sure if we follow the same behavior naturally in Sedona. If we don't let me know, and we can figure out what to do next.

We could test something like whether `Point(1, 1, np.nan)`  becomes `Point(1, 1, 3)` when calling `.force_3d(3)`

python/sedona/spark/geopandas/base.py
```diff
@@ -949,8 +949,68 @@ def force_2d(self):
...
+
+        Note that for empty geometries, 3D is only supported since GEOS 3.9 and then
+        still only for simple geometries (non-collections).
+
```
```suggestion
```

I think we can delete this part. The `GEOS 3.9` part doesn't apply to use because we're built not on [GEOS](https://github.com/libgeos/geos) (which is a C library). And I think we do support collections since the `test_match_geopandas_series` test passed naturally (`self.geoms` in that test contains collections). Could you add a single test in `test_geoseries.py` that tests a GEOMETRYCOLLECTION with a  non-zero z-arg?

e.g. `GeometryCollection(Point(1, 1), LineString([(1, 1), (0, 1), (1, 0)]))` and `force_3d(3)`

python/tests/geopandas/test_match_geopandas_series.py
```diff
@@ -878,7 +878,50 @@ def test_force_2d(self):
...
 
     def test_force_3d(self):
-        pass
+        # force_3d was added from geopandas 1.0.0
+        if parse_version(gpd.__version__) < parse_version("1.0.0"):
+            pytest.skip("geopandas force_3d requires version 1.0.0 or higher")
+        # 1) Promote 2D to 3D with default z=0.0
+        for geom in self.geoms:
+            sgpd_result = GeoSeries(geom).force_3d()
+            gpd_result = gpd.GeoSeries(geom).force_3d()
```
```suggestion
            sgpd_result = GeoSeries(geom).force_3d(4)
            gpd_result = gpd.GeoSeries(geom).force_3d(4)
```
This is a minor change, and I don't think Sedona would fail it, but I think this would be a slightly better test if we test a non-zero z argument. These "match" tests in `test_match_geopandas_series.py` are the most comprehensive/powerful, so I'd prefer to avoid a common default case like 0. e.g. the chances that we have a bug that always set z=0 in some case is much higher than the chance that it always set z=4.

chay0112:
> Looks good. Just some minor feedback about improving the docs and the tests. Should be good to merge afterwards.

@petern48 Thanks Peter, Great insights as always. Just addressed them.

python/tests/geopandas/test_geoseries.py
```diff
@@ -1520,7 +1520,79 @@ def test_force_2d(self):
...
+
+        # 3. Array-like z: use ps.Series
+        z = [0, 2, 2, 3, 4, 5]
+        expected = gpd.GeoSeries(
+            [
+                Point(1, 2, 0),
+                Point(0.5, 2.5, 2),
+                Point(1, 1, 2),
+                LineString([(1, 1, 2), (0, 1, 2), (1, 0, 2)]),
+                Polygon([(0, 0, 3), (0, 10, 3), (10, 10, 3), (0, 0, 3)]),
```
```suggestion
                LineString([(1, 1, 3), (0, 1, 3), (1, 0, 3)]),
                Polygon([(0, 0, 4), (0, 10, 4), (10, 10, 4), (0, 0, 4)]),
```

I noticed our tests here were slightly buggy. Here's what the expected results should look like. These geoms correspond to the 3 and 4 in `z = [0, 2, 2, 3, 4, 5]`. What's concerning is that the tests pass anyway. Experimenting locally, I noticed we can change the z-coordinate however we like, and the tests still pass. I think what's happening is that the `check_sgpd_equals_gpd()` doesn't yet check anything past the x and y dimensions...

(sigh) Unfortunately, what this means is that we'll need to modify the test code to test this properly before we can merge this PR. I filed an issue for this: https://github.com/apache/sedona/issues/2513. You're welcome to try taking a stab at it or just move on to something else (while keeping this PR open). My time is limited, so I wouldn't be able to investigate it for at least a few days. I don't love having to put a hold on merging your great work here, but this is just how it is sometimes. Sorry about this :(

chay0112: Got it, Thanks for the insights Peter. No problem, I'll move on to a new one.

petern48:
Alright @chay0112, the [fix](https://github.com/apache/sedona/pull/2517) has been merged. Think this should be ready once that last suggestion is accepted.

chay0112:
> Alright @chay0112, the [fix](https://github.com/apache/sedona/pull/2517) has been merged. Think this should be ready once that last suggestion is accepted.

Thanks for your fix Peter, I believe I need to rephrase my tests in test_match_geopandas_series.py because things like Line Strings to Linear Rings and Polygons to Multipolygons would fail

chay0112:
@petern48 I made some adjustments to the tests in test_match_geopandas_series.py because the previous assertions were failing for M and ZM geometry cases. Please review the changes and let me know if there’s a better or cleaner approach. Feedback is welcome.

petern48:
Responding to the deletions in this commit: https://github.com/apache/sedona/pull/2512/commits/f90875aae014649e325cbc63ef9de748a6afc739. Here are the failures from the test_match_geopandas tests (I ran locally). Next time, ideally report back when you see behavior like this, instead of relying on me to investigate myself (sometimes, it may take longer that way). Sometimes, it's reasonable to skip; sometimes it's major enough to just put a halt to it. Alright let's see.

Error for LinearRing:
```
ValueError: Geometry equality check failed for LINESTRING Z (0 0 4, 1 0 4, 1 1 4, 0 1 4, 0 0 4) and LINEARRING Z (0 0 4, 1 0 4, 1 1 4, 0 1 4, 0 0 4)
```
Sedona is returning the equivalent of a LineString. Sedona doesn't support LinearRings since LineStrings are enough. Yep this is fine to skip ✅. 

Error for MultiPolygon:
```
ValueError: Geometry equality check failed for POLYGON Z ((0 0 4, 0 1 4, 1 0 4, 0 0 4), (0.1 0.1 4, 0.1 0.2 4, 0.2 0.1 4, 0.1 0.1 4)) and MULTIPOLYGON Z (((0 0 4, 0 1 4, 1 0 4, 0 0 4), (0.1 0.1 4, 0.1 0.2 4, 0.2 0.1 4, 0.1 0.1 4))
```

Ah interesting, it's the same geometry, but we return polygon instead of multipolygon. This is a minor bug in Sedona we found. Ideally, we want to return MultiPolygon. I dug in real quick and submitted a fix for this: https://github.com/apache/sedona/pull/2526

Trying it on PostGIS, I see returning MultiPolygon is the desired behavior as well
```
SELECT ST_AsText(ST_Force3D(ST_GeomFromText('MULTIPOLYGON(((0 0, 0 1, 1 0, 0 0), (0.1 0.1, 0.1 0.2, 0.2 0.1, 0.1 0.1)))'), 4));
-- MULTIPOLYGON Z (((0 0 4,0 1 4,1 0 4,0 0 4),(0.1 0.1 4,0.1 0.2 4,0.2 0.1 4,0.1 0.1 4)))
```

Error for GeometryCollection
```
ValueError: Geometry equality check failed for GEOMETRYCOLLECTION Z (MULTIPOINT Z ((0 0 4), (1 1 4)), MULTILINESTRING Z ((0 0 4, 1 1 4), (2 2 4, 3 3 4)), POLYGON Z ((0 0 4, 0 1 4, 1 0 4, 0 0 4), (0.1 0.1 4, 0.1 0.2 4, 0.2 0.1 4, 0.1 0.1 4))) and GEOMETRYCOLLECTION Z (MULTIPOINT Z ((0 0 4), (1 1 4)), MULTILINESTRING Z ((0 0 4, 1 1 4), (2 2 4, 3 3 4)), MULTIPOLYGON Z (((0 0 4, 0 1 4, 1 0 4, 0 0 4), (0.1 0.1 4, 0.1 0.2 4, 0.2 0.1 4, 0.1 0.1 4))))
```

Fails for the same reason as multipolygon (since it has a multipolygon inside of it).

Error for M coordinates. Hmm, yeah this one is interesting. Give me a bit more time to figure out what to do here.
```
# For POINT M (1 2 3) input
ValueError: Geometry equality check failed for POINT Z (1 2 7.5) and POINT M (1 2 3)

# For POINT ZM (1 2 3 4) input
ValueError: Geometry equality check failed for POINT Z (1 2 3) and POINT Z (1 2 7.5)
```

chay0112:
> Responding to the deletions in this commit: [f90875a](https://github.com/apache/sedona/commit/f90875aae014649e325cbc63ef9de748a6afc739). Here are the failures from the test_match_geopandas tests (I ran locally). Next time, ideally report back when you see behavior like this, instead of relying on me to investigate myself (sometimes, it may take longer that way). Sometimes, it's reasonable to skip; sometimes it's major enough to just put a halt to it. Alright let's see.
> 
> Error for LinearRing:
> 
> ```
> ValueError: Geometry equality check failed for LINESTRING Z (0 0 4, 1 0 4, 1 1 4, 0 1 4, 0 0 4) and LINEARRING Z (0 0 4, 1 0 4, 1 1 4, 0 1 4, 0 0 4)
> ```
> 
> Sedona is returning the equivalent of a LineString. Sedona doesn't support LinearRings since LineStrings are enough. Yep this is fine to skip ✅.
> 
> Error for MultiPolygon:
> 
> ```
> ValueError: Geometry equality check failed for POLYGON Z ((0 0 4, 0 1 4, 1 0 4, 0 0 4), (0.1 0.1 4, 0.1 0.2 4, 0.2 0.1 4, 0.1 0.1 4)) and MULTIPOLYGON Z (((0 0 4, 0 1 4, 1 0 4, 0 0 4), (0.1 0.1 4, 0.1 0.2 4, 0.2 0.1 4, 0.1 0.1 4))
> ```
> 
> Ah interesting, it's the same geometry, but we return polygon instead of multipolygon. This is a minor bug in Sedona we found. Ideally, we want to return MultiPolygon. I dug in real quick and submitted a fix for this: #2526
> 
> Trying it on PostGIS, I see returning MultiPolygon is the desired behavior as well
> 
> ```
> SELECT ST_AsText(ST_Force3D(ST_GeomFromText('MULTIPOLYGON(((0 0, 0 1, 1 0, 0 0), (0.1 0.1, 0.1 0.2, 0.2 0.1, 0.1 0.1)))'), 4));
> -- MULTIPOLYGON Z (((0 0 4,0 1 4,1 0 4,0 0 4),(0.1 0.1 4,0.1 0.2 4,0.2 0.1 4,0.1 0.1 4)))
> ```
> 
> Error for GeometryCollection
> 
> ```
> ValueError: Geometry equality check failed for GEOMETRYCOLLECTION Z (MULTIPOINT Z ((0 0 4), (1 1 4)), MULTILINESTRING Z ((0 0 4, 1 1 4), (2 2 4, 3 3 4)), POLYGON Z ((0 0 4, 0 1 4, 1 0 4, 0 0 4), (0.1 0.1 4, 0.1 0.2 4, 0.2 0.1 4, 0.1 0.1 4))) and GEOMETRYCOLLECTION Z (MULTIPOINT Z ((0 0 4), (1 1 4)), MULTILINESTRING Z ((0 0 4, 1 1 4), (2 2 4, 3 3 4)), MULTIPOLYGON Z (((0 0 4, 0 1 4, 1 0 4, 0 0 4), (0.1 0.1 4, 0.1 0.2 4, 0.2 0.1 4, 0.1 0.1 4))))
> ```
> 
> Fails for the same reason as multipolygon (since it has a multipolygon inside of it).
> 
> Error for M coordinates. Hmm, yeah this one is interesting. Give me a bit more time to figure out what to do here.
> 
> ```
> # For POINT M (1 2 3) input
> ValueError: Geometry equality check failed for POINT Z (1 2 7.5) and POINT M (1 2 3)
> 
> # For POINT ZM (1 2 3 4) input
> ValueError: Geometry equality check failed for POINT Z (1 2 3) and POINT Z (1 2 7.5)
> ```

@petern48 Sure, I'll report the anomalies from next time, Thanks for a very clear explanation. Take your time. Thanks

python/tests/geopandas/test_match_geopandas_series.py
```diff
@@ -878,7 +878,48 @@ def test_force_2d(self):
         self.check_sgpd_equals_gpd(sgpd_3d, gpd_3d)
 
     def test_force_3d(self):
-        pass
+        # force_3d was added from geopandas 1.0.0
+        if parse_version(gpd.__version__) < parse_version("1.0.0"):
+            pytest.skip("geopandas force_3d requires version 1.0.0 or higher")
+        # 1) Promote 2D to 3D with z = 4
+        for geom in self.geoms:
+            if isinstance(geom[0], (LinearRing, GeometryCollection, MultiPolygon)):
```
```suggestion
            if isinstance(geom[0], (LinearRing)):
```

This should theoretically pass now.

python/sedona/spark/geopandas/base.py
```diff
@@ -949,8 +949,66 @@ def force_2d(self):
         """
         return _delegate_to_geometry_column("force_2d", self)
 
-    # def force_3d(self, z=0):
-    #     raise NotImplementedError("This method is not implemented yet.")
+    def force_3d(self, z=0.0):
+        """Force the dimensionality of a geometry to 3D.
+
+        2D geometries will get the provided Z coordinate; 3D geometries
+        are unchanged (unless their Z coordinate is ``np.nan``).
```
petern48, line 956:
```suggestion
        are unchanged (unless their Z coordinate is ``np.nan``).
        
        Note: Sedona's behavior may differ from Geopandas' for M and ZM geometries.
        For M geometries, Sedona will replace the M coordinate and add the Z coordinate.
        For ZM geometries, Sedona will drop the M coordinate and retain the Z coordinate.
```

python/tests/geopandas/test_geoseries.py
```diff
@@ -1520,7 +1520,79 @@ def test_force_2d(self):
...
+                GeometryCollection(
+                    [
+                        Point(1, 1, 5),
+                        LineString([(1, 1, 5), (0, 1, 5), (1, 0, 5)]),
+                    ]
+                ),
+            ]
+        )
+        result = s.force_3d(z)
+        self.check_sgpd_equals_gpd(result, expected)
```
petern48, line 1595: Could you add these cases below?

```
ST_Force3D(st_geomfromtext('POINT M (1 2 3)'), 7.5)
-- Result: POINT Z(1 2 7.5)

ST_Force3D(st_geomfromtext('POINT ZM (1 2 3 4)'), 7.5)
-- Result: POINT Z(1 2 3)
```

petern48: For those curious about the difference in behavior, Geopandas returns the following instead. It's strange and inconsistent imo... I would guess it's unintentional and just a result of adding M support later on.

```python
gpd.GeoSeries([shapely.wkt.loads("POINT M (1 2 3)")]).force_3d(7.5)
0    POINT M (1 2 3)

gpd.GeoSeries([shapely.wkt.loads("POINT ZM (1 2 3 4)")]).force_3d(7.5)
0    POINT Z (1 2 7.5)
```

petern48 (commented):
Sorry for the delay @chay0112. Thanks for being patient. I took another look and saw that Sedona follows PostGIS behavior (which is what we want), and Geopandas behavior apparently is just different. I think we can keep this difference in behavior, but specify this difference in the docstring. Left some suggestions, but I think this should be ready once addressed.

petern48:
Also, could you mention "geopandas" in the PR title (somehow). The PR title eventually becomes the commit message when merged, so it's helpful to specify these details. Could be something like one of the following (I typically do the first one).
- [GH-XXXX] Geopandas: Implement force_3d
- [GH-XXXX] Implement geopandas force_3d

python/tests/geopandas/test_match_geopandas_series.py
chay0112: Hi Peter, Thanks for your suggestions.
I believe if we remove Multipolygon and GeometryCollections the test would still fail for geoms, something like below
`FAILED
============================================================================== FAILURES ==============================================================================
_______________________________________________________________ TestMatchGeopandasSeries.test_force_3d _______________________________________________________________

python/tests/geopandas/test_match_geopandas_series.py:890: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
python/tests/geopandas/test_geopandas_base.py:70: in check_sgpd_equals_gpd
    cls.assert_geometry_almost_equal(a, e, tolerance)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cls = <class 'tests.geopandas.test_match_geopandas_series.TestMatchGeopandasSeries'>
left_geom = <POLYGON Z ((0 0 4, 0 1 4, 1 0 4, 0 0 4), (0.1 0.1 4, 0.1 0.2 4, 0.2 0.1 4, ...>
right_geom = <MULTIPOLYGON Z (((0 0 4, 0 1 4, 1 0 4, 0 0 4), (0.1 0.1 4, 0.1 0.2 4, 0.2 0...>, tolerance = 0.01

ValueError: Geometry equality check failed for POLYGON Z ((0 0 4, 0 1 4, 1 0 4, 0 0 4), (0.1 0.1 4, 0.1 0.2 4, 0.2 0.1 4, 0.1 0.1 4)) and MULTIPOLYGON Z (((0 0 4, 0 1 4, 1 0 4, 0 0 4), (0.1 0.1 4, 0.1 0.2 4, 0.2 0.1 4, 0.1 0.1 4)))
`

petern48: Ah, I realize you're not *yet* familiar with iterating with Java changes. You need to update your code to reflect the changes I made in this [PR](https://github.com/apache/sedona/pull/2526). I'm guessing you likely followed this [guide](https://github.com/apache/sedona/blob/0523fbdc600d86b8088feba39db829a5dccc11d8/docs/setup/compile.md#L4) for dev env setup, since it was mentioned in the original geopandas EPIC issue. 

1. First, make sure you pull in the latest main and merge it into this branch.
2. Now, even after you get the changes, you need to rebuild the Sedona JAR and move it to your SPARK_HOME directory. These are the two commands in section 3) of those compile docs.

Basically, you just need to redo the 2 commands in section 3, and then the changes should be reflected when you run pytest.
```
mvn clean install -DskipTests -Dgeotools
cp spark-shaded/target/sedona-spark-shaded-*.jar $SPARK_HOME/jars/
```

Note, if you run into trouble with the first command succeeding, you may need to specify versions in the `mvn` build command. e.g. Here's what I run each time: `mvn clean install -DskipTests -Dgeotools -Dspark=3.4 -Pscala2.12`. The same command likely would still work for you if you set up the same versions as in those docs. @chay0112 Give it a try. Knowing how to do this is also very helpful for getting into fixing small Java bugs, like I had done earlier.

chay0112:
@petern48 Just addressed all the comments. It looks like the docs build is failing, and I noticed the same issue in other PRs as well. I'm not sure of the root cause, but if that's not a blocker, this PR is ready for review. No rush though !

python/tests/geopandas/test_match_geopandas_series.py
chay0112: Cool, Thanks for letting me know.

petern48:
Yeah, the doc build failure should be unrelated. Recently was fixed here: https://github.com/apache/sedona/pull/2551

PR #2517, opened 2025-11-23
[GH-2513] chore(python): Make 'assert_geometry_almost_equal' check for Z and M dimensions too

### LINKED ISSUE #2513: chore(geopandas): Make `check_sgpd_equals_gpd()` test for Z and M dimensions too
Unfortunately, it seems like `check_sgpd_equals_gpd()` passes when comparing geoms that only differ by Z-dimension. We need to modify the test code to test for Z and M dimensions. We'll then want to write a negative test that shows that `check_sgpd_equals_gpd()` fails when we expect it too. e.g comparing Point(1, 1, 3) and Point(1, 1, 4) should fail

petern48:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2513

Make 'assert_geometry_almost_equal' check for Z and M dimensions too for shapely >= 2.1.0

## How was this patch tested?
Added test


python/tests/test_base.py
```diff
@@ -122,14 +125,31 @@ def assert_geometry_almost_equal(
         right_geom: Union[str, BaseGeometry],
         tolerance=1e-6,
     ):
+        """
+        Assert that two geometries are almost equal.
+
+        Note: this function will only check Z and M dimensions for shapely >= 2.1.0 (python >= 3.10)
+
+        When comparing geometries with Z or M dimensions, this function will ignore `tolerance` and check for exact equality.
```
petern48, line 133: Reasons for this:

- Z and M support isn't introduced until [Shapely 2.1.0](https://shapely.readthedocs.io/en/2.1.0/release/2.x.html#version-2-1-0-2025-04-03)

  - `equals_exact` only compares X and Y dimensions and not Z/M (as mentioned [here](https://shapely.readthedocs.io/en/2.1.0/reference/shapely.equals_identical.html) in the equals_identical docs, but not the equals_exact docs...)

- We now use `equals_identical` to check Z/M, but it doesn't support a `tolerance` parameter. I think this is fine, given the rare need for tolerance.

jiayuasu (approved):
Thank you!

PR #2526, opened 2025-11-25
[GH-2525] bug: Make ST_Force3D return multipolygons even if it's a single polygon inside

### LINKED ISSUE #2525: bug: ST_Force3d() returns polygon when input is multipolygon with a single nested polygon
See https://github.com/apache/sedona/pull/2512#issuecomment-3573577084. Basically, we return polygon instead of multipolygon.

petern48:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2525

When ST_Force3D received a MultiPolygon as input, it would previously downgrade it into a Polygon if possible. This PR makes it return a multipolygon even if it's a single polygon inside, which is the same behavior as Postgis.

## How was this patch tested?
Added tests

- No, this is a minor edge case

### DIFF
--- a/common/src/test/java/org/apache/sedona/common/FunctionsTest.java
+++ b/common/src/test/java/org/apache/sedona/common/FunctionsTest.java
@@ -2461,6 +2461,24 @@ public class FunctionsTest extends TestBase {
   }
 
   @Test
+  public void force3DMultiPolygonWithSinglePolygon() {
+    // Test that a MultiPolygon with a single polygon remains a MultiPolygon after force3D
+    Polygon polygon = GEOMETRY_FACTORY.createPolygon(coordArray(0, 0, 10, 0, 10, 10, 0, 10, 0, 0));
+    MultiPolygon multiPolygon = GEOMETRY_FACTORY.createMultiPolygon(new Polygon[] {polygon});
+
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/apache/sedona/common/utils/GeometryForce3DTransformer.java
```diff
@@ -49,4 +49,16 @@ public static Geometry transform(Geometry geometry, double zValue) {
     GeometryForce3DTransformer transformer = new GeometryForce3DTransformer(zValue);
     return transformer.transform(geometry);
   }
+
+  protected Geometry transformMultiPolygon(MultiPolygon geom, Geometry parent) {
```
petern48, line 53: This is simply an override of this [method](https://github.com/locationtech/jts/blob/ab57bffd4250f14416315249565e585a76c0c489/modules/core/src/main/java/org/locationtech/jts/geom/util/GeometryTransformer.java#L283) in JTS's `GeometryTransformer` class to handle MultiPolygons separately.

petern48:
R Build failure is unrelated. Looks like it's already in the process of being fixed: https://github.com/apache/sedona/pull/2524

PR #2529, opened 2025-11-25
[GH-2527] Implement GeoPandas concave_hull

### LINKED ISSUE #2527: Geopandas: Implement concave_hull
### Use case:

- Add support for Geoseries.concave_hull

### Description:

- Implements GeoSeries.concave_hull, which computes the concave hull for each geometry in the series. The concave hull is the smallest concave polygon that contains all points of the geometry, controlled by a ratio parameter that determines the level of concavity and an option to allow holes. For geometries with fewer than three points, the result is a LineString or Point. This method ensures each geometry in the series is replaced by its concave hull representation, matching the behavior of GeoPandas.

### Related Issue:

- Links to epic https://github.com/apache/sedona/issues/2230

chay0112:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2527 

- Implement concave_hull functionality

## How was this patch tested?

### DIFF
--- a/python/sedona/spark/geopandas/base.py
+++ b/python/sedona/spark/geopandas/base.py
@@ -608,8 +608,73 @@ class GeoFrame(metaclass=ABCMeta):
         """
         return _delegate_to_geometry_column("centroid", self)
 
-    # def concave_hull(self, ratio=0.0, allow_holes=False):
-    #     raise NotImplementedError("This method is not implemented yet.")
+    def concave_hull(self, ratio=0.0, allow_holes=False):
+        """Return a ``GeoSeries`` of geometries representing the concave hull
+        of vertices of each geometry.
+
+        The concave hull of a geometry is the smallest concave `Polygon`
+        containing all the points in each geometry, unless the number of points
+        in the geometric object is less than three. For two points, the concave
+        hull collapses to a `LineString`; for 1, a `Point`.
+
+        The hull is constructed by removing border triangles of the Delaunay
+        Triangulation of the points as long as their "size" is larger than the
+        maximum edge length ratio and optionally allowing holes. The edge length factor
+        is a fraction of the length difference between the longest and shortest edges
+        in the Delaunay Triangulation of the input points. For further information
+        on the algorithm used, see
+        https://libgeos.org/doxygen/classgeos_1_1algorithm_1_1hull_1_1ConcaveHull.html
+
+        Parameters
+        ----------
+        ratio : float, (optional, default 0.0)
+            Number in the range [0, 1]. Higher numbers will include fewer vertices
+            in the hull.
+        allow_holes : bool, (optional, default False)
+            If set to True, the concave hull may have holes.
+
+        Examples
+        --------
+        >>> from sedona.spark.geopandas import GeoSeries
+        >>> from shapely.geometry import Polygon, LineString, Point, MultiPoint
+        >>> s = GeoSeries(
+        ...     [
+        ...         Polygon([(0, 0), (1, 1), (0, 1)]),
+        ...         LineString([(0, 0), (1, 1), (1, 0)]),
+        ...         MultiPoint([(0, 0), (1, 1), (0, 1), (1, 0), (0.5, 0.5)]),
+        ...         MultiPoint([(0, 0), (1, 1)]),
+        ...         Point(0, 0),
+        ...     ],
+        ...     crs=3857
+        ... )
+        >>> s
+        0                       POLYGON ((0 0, 1 1, 0 1, 0 0))
+        1                           LINESTRING (0 0, 1 1, 1 0)
+        2    MULTIPOINT ((0 0), (1 1), (0 1), (1 0), (0.5 0...
+        3                            MULTIPOINT ((0 0), (1 1))
+        4                                          POINT (0 0)
+        dtype: geometry
+
+        >>> s.concave_hull()
+        0                      POLYGON ((0 1, 1 1, 0 0, 0 1))
+        1                      POLYGON ((0 0, 1 1, 1 0, 0 0))
+        2    POLYGON ((0.5 0.5, 0 1, 1 1, 1 0, 0 0, 0.5 0.5))
+        3                               LINESTRING (0 0, 1 1)
+        4                                         POINT (0 0)
+        dtype: geometry
+
+        See Also
+        --------
+        GeoSeries.convex_hull : convex hull geometry
+
+        Notes
+        -----
+        The algorithms considers only vertices of each geometry. As a result the
+        hull may not fully enclose input geometry. If that happens, increasing ``ratio``
+        should resolve the issue.
+
+        """
+        return _delegate_to_geometry_column("concave_hull", self, ratio, allow_holes)
 
     @property
     def convex_hull(self):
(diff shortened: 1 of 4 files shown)


chay0112:
Hi @petern48 

I'm encountering a test failure in our Sedona/GeoPandas parity suite for concave_hull(Passing in my local). The error is due to the vertex order of polygons returned by Sedona and GeoPandas being different, even though the polygons are geometrically identical. Our current comparison uses equals_exact(assert_geometry_almost_equal), which might be sensitive to vertex order:

`=========================== short test summary info ============================
FAILED tests/geopandas/test_match_geopandas_series.py::TestMatchGeopandasSeries::test_concave_hull - ValueError: Geometry equality check failed for POLYGON ((1 4, 3 4, 4 4, 3 1, 4 0, 0 0, 1 1, 0 4, 1 4)) and POLYGON ((1 4, 3 4, 4 4, 4 0, 3 1, 0 0, 1 1, 0 4, 1 4))`

[https://github.com/apache/sedona/actions/runs/19687051539/job/56394868966?pr=2529](url)

Would it be acceptable to change our comparison from equals_exact to equals for polygons, so that we check geometric equality rather than strict vertex order? Or is there a recommended workaround for this situation, or something I might be missing in our test setup?

Thank you for your guidance!

python/tests/geopandas/test_match_geopandas_series.py
```diff
@@ -722,8 +722,35 @@ def test_centroid(self):
...
+            )
+            gpd_result = gpd.GeoSeries(geom).concave_hull(
+                ratio=ratio, allow_holes=allow_holes
+            )
+            self.check_sgpd_equals_gpd(sgpd_result, gpd_result)
+
+        mixed = [self.points[1], self.linestrings[1], self.polygons[1], None]
+        sgpd_result = GeoSeries(mixed).concave_hull()
+        gpd_result = gpd.GeoSeries(mixed).concave_hull()
+        self.check_sgpd_equals_gpd(sgpd_result, gpd_result)
```
petern48, line 804:
```suggestion
```
I don't think this test is necessary. We're already testing all of those cases separately in the `for geom in self.geoms` above. The function is executed on each geometry separately, so whether they're mixed together or not doesn't matter.

python/tests/geopandas/test_match_geopandas_series.py
```diff
@@ -722,8 +722,35 @@ def test_centroid(self):
...
+    @pytest.mark.skipif(
+        parse_version(gpd.__version__) < parse_version("0.14.0"),
+        reason="geopandas concave_hull requires version 0.14.0 or higher",
+    )
     def test_concave_hull(self):
-        pass
+        for geom in self.geoms:
+            sgpd_result = GeoSeries(geom).concave_hull()
+            gpd_result = gpd.GeoSeries(geom).concave_hull()
+            self.check_sgpd_equals_gpd(sgpd_result, gpd_result)
```
petern48: We should test multiple ratio values here, not just the default of 0.0. We could do something like: Do one third of the iterations w/ 0.0, then the next third w/ 0.5, and the last third w/ 1.0. (Note the range of valid values for this argument is [0, 1].

Then also test allow_holes below (as you already are doing).

python/tests/geopandas/test_geoseries.py
```diff
@@ -1238,7 +1238,34 @@ def test_centroid(self):
...
+                Polygon([(0, 0), (1, 1), (0, 1)]),
+                LineString([(0, 0), (1, 1), (1, 0)]),
+                MultiPoint([(0, 0), (1, 1), (0, 1), (1, 0), (0.5, 0.5)]),
+                MultiPoint([(0, 0), (1, 1)]),
+                Point(0, 0),
+            ],
+            crs=3857,
+        )
+
+        result = s.concave_hull()
```
petern48: Would be nice to have a test for allow_holes=True in this file too. But we should first get to the bottom of the errors.

petern48:
> Would it be acceptable to change our comparison from equals_exact to equals for polygons, so that we check geometric equality rather than strict vertex order? Or is there a recommended workaround for this situation, or something I might be missing in our test setup?

The issue is not whether we're using `equals_exact()` or `equals()`. If you give it a try, you'll see that the comparison fails either way. Despite the name, equals_exact is actually more lenient, since it supports a tolerance parameter.

<details>
<summary>Script to test using equals instead</summary>

```
import shapely
geom1 = shapely.wkt.loads("POLYGON ((1 4, 3 4, 4 4, 3 1, 4 0, 0 0, 1 1, 0 4, 1 4))")
geom2 = shapely.wkt.loads("POLYGON ((1 4, 3 4, 4 4, 4 0, 3 1, 0 0, 1 1, 0 4, 1 4))")
print(geom1.equals(geom2))  # False
print(geom1.equals_exact(geom2, 0.000001))  # False
```
</details>

<details>
<summary>Images to visualize the different results </summary>
image
image
</details>

As you can see above, these geometries actually are different. Why this is happening, I'm not sure. It could either be a bug in Sedona's function or a difference in algorithms or edge case behaviors. If it's a significant bug, we may want to hold off. If it's a reasonable difference in behavior that's still correct, we can add a note in the docs and merge this anyway. Either way, we need to understand what's happening before merging. I encourage you to investigate, though you're not obligated to, of course, and can continue with something else instead.

python/tests/geopandas/test_match_geopandas_series.py
jiayuasu: Removed the mixed-geometry case. The separate per-geometry fixtures remain as the parity coverage.

python/tests/geopandas/test_match_geopandas_series.py
jiayuasu: Updated the parity loop to split the geometry fixtures across ratios 0.0, 0.5, and 1.0 as suggested.

python/tests/geopandas/test_geoseries.py
jiayuasu: Added direct and parity coverage with allow_holes=True using an asymmetric MultiPoint fixture that produces one interior ring. The earlier symmetric fixture no longer reproduces with current JTS and GEOS stacks; tied boundary triangles can be processed in different orders, so the replacement avoids that ambiguity while retaining strict geometry equality.

python/sedona/spark/geopandas/base.py
```diff
@@ -712,19 +712,25 @@ def centroid(self):
...
-        The concave hull of a geometry is a possibly concave geometry that
-        encloses the input geometry.
+        A concave hull is a concave or convex `Polygon` containing all the
+        vertices in each geometry. For two vertices, the hull collapses to a
+        `LineString`; for one vertex, it collapses to a `Point`.
+
+        The hull is constructed by removing border triangles of the Delaunay
+        triangulation while their size is larger than the maximum edge length
+        ratio. The ratio is a fraction of the difference between the longest
+        and shortest edges in the triangulation.
```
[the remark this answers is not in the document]
jiayuasu: Reworded this in fef658e1da3 to state that ratio determines the maximum edge-length threshold by interpolating between the shortest and longest Delaunay edges. This keeps the documented JTS semantics while removing the ambiguous comparison between triangle size and a dimensionless ratio.

PR #2563, opened 2025-12-17
[GH-2565] Fix NULL handling for various aggregation functions in SedonaSpark

### LINKED ISSUE #2565: Aggregation functions in SedonaSpark throw exception when processing `NULL` values
This is a short repro for the problem:

```python
spark.sql(
    """
    WITH t AS (SELECT explode(array(NULL, ST_GeomFromWKT('POINT (1 2)'), NULL)) AS geom)
    SELECT ST_Envelope_Aggr(geom) FROM t
    """
).show()
```

This query fails with the following exception:

```
org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 3.0 failed 1 times, most recent failure: Lost task 0.0 in stage 3.0 (TID 2) (bos-laptop.bbrouter executor driver): java.lang.NullPointerException: Cannot invoke "org.locationtech.jts.geom.Geometry.getEnvelopeInternal()" because "input" is null
	at org.apache.spark.sql.sedona_sql.expressions.ST_Envelope_Aggr.reduce(AggregateFunctions.scala:112)
	at org.apache.spark.sql.sedona_sql.expressions.ST_Envelope_Aggr.reduce(AggregateFunctions.scala:106)
	at org.apache.spark.sql.execution.aggregate.ScalaAggregator.update(udaf.scala:527)
```

Other aggregation functions such as `ST_Union_Aggr`, `ST_Intersection_Aggr` have the same problem. The correct behavior should be ignoring NULL inputs, and return NULL when the inputs are all NULLs.

Kontinuation:

- Yes, and the PR name follows the format `[GH-2565] my subject`. Closes #2565

This PR refactors several spatial aggregate functions to fix `NULL` handling. Impacted functions are

* **ST_Envelope_Aggr**: fixed null handling, simplified the implementation by switching to lightweight intermediate data structure
* **ST_Union_Aggr**: fixed null handling
* **ST_Intersection_Aggr**: fixed null handling
* **ST_Collect_Agg**: return null instead of empty geometry collection when all the inputs are null. This is theoretically a breaking change, but it is a niche case so hopefully it won't impact many users.

## How was this patch tested?

Added new tests and tested locally using a standalone cluster to make sure that we are not damaging the serde in distributed environment.

### DIFF
--- a/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/AggregateFunctions.scala
+++ b/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/AggregateFunctions.scala
@@ -92,35 +92,76 @@ private[apache] class ST_Union_Aggr(bufferSize: Int = 1000)
 }
 
 /**
+ * A helper class to store envelope boundary during aggregation. We use this custom case class
+ * instead of JTS Envelope to work with the Spark Encoder.
+ */
+case class EnvelopeBuffer(minX: Double, maxX: Double, minY: Double, maxY: Double) {
+  def isNull: Boolean = minX > maxX
+
+  def toEnvelope: Envelope = {
+    if (isNull) {
+      new Envelope()
+    } else {
+      new Envelope(minX, maxX, minY, maxY)
+    }
+  }
+
+  def merge(other: EnvelopeBuffer): EnvelopeBuffer = {
+    if (this.isNull) {
+      other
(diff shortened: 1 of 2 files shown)


Kontinuation:
The Python build extension failure is probably caused by the newly released PySpark 4.1.0 not playing well with Windows. See https://github.com/apache/spark/pull/50466

PR #2753, opened 2026-03-13
[CI] Add basic tests action workflow with one job

jbampton:
Apache Airflow uses this same basic check:

https://github.com/apache/airflow/blob/a9a9c5a57fbaaa11499967b39bac11adedea857e/.github/workflows/basic-tests.yml#L270

We don't have much test coverage running on Windows.

Should be quick to run on the GitHub CI

- No:
  - this is a CI update. The PR name follows the format `[CI] my subject`

Minor improvement one more check.

Maybe we can add more basic test jobs in future ?

### DIFF
--- a/.github/workflows/basic-tests.yml
+++ b/.github/workflows/basic-tests.yml
@@ -1,0 +1,41 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: Basic Tests
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - '*'
+
+permissions:
+  contents: read
+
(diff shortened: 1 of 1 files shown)


jiayuasu:
Why do we need to test git clone on Windows?

jbampton:
From Google:

Testing git clone on Windows is essential because of how the operating system handles files differently than Linux or macOS. Since Git was originally designed for Linux, several "Windows-specific" behaviors can cause a clone to fail or corrupt the project if not properly configured. 
The primary reasons for testing on Windows include:
1. Line Ending Inconsistencies

* The Issue: Windows uses Carriage Return + Line Feed (CRLF) for new lines, while Linux and macOS use just Line Feed (LF).
* The Risk: Without testing, a clone might automatically convert all files to CRLF, which can break scripts (like .sh files) intended to run in Docker or cross-platform environments.
* The Fix: Developers often use the git config --global core.autocrlf true setting to manage these conversions automatically. 

2. Case Sensitivity Conflicts

* The Issue: The Windows file system is typically case-insensitive (e.g., README.md and readme.md are the same file), whereas Linux is case-sensitive.
* The Risk: If a repository contains two files with the same name but different casing, cloning on Windows will result in a path collision. One file will overwrite the other, or the clone will fail with an error.
* The Fix: Testing ensures that the repository structure is compatible with the core.ignorecase setting commonly used on Windows. 

3. File Path Length Limits

* The Issue: By default, Windows has a "MAX_PATH" limit of 260 characters.
* The Risk: Deeply nested projects (common in Node.js or Java) may fail to clone because the resulting file paths are too long for Windows to handle.
* The Fix: Testing identifies if you need to enable git config --system core.longpaths true to bypass these limits. [6] 

4. Permission & Tooling Differences

* Permissions: Windows does not use the same POSIX file permissions as Linux. This can lead to "executable" bits being lost or ignored after a clone.
* SSH vs. HTTPS: Windows often requires specific credential managers or configurations for SSH keys that differ from Unix-based systems.
* Network Performance: Features like "Receive Window Auto-Tuning" in Windows can sometimes cause git clone to be significantly slower or more prone to timing out than on other platforms.

PR #2772, opened 2026-03-21
[GH-2230] Implement GeoSeries.minimum_clearance

piyushka-ally:
# [GH-2230] Implement GeoSeries.minimum_clearance

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Part of #2230

Implements `GeoSeries.minimum_clearance()` from the [GH-2230 EPIC](https://github.com/apache/sedona/issues/2230) to expand Sedona's geopandas-compatible API:

**`GeoSeries.minimum_clearance()` (method)**
- Returns the minimum clearance distance of each geometry as a `float64` Series — the smallest distance by which any vertex could be moved to produce an invalid geometry.
- Backed by `ST_MinimumClearance`.
- Handles the JTS/GEOS representation difference: JTS returns `Double.MAX_VALUE` for degenerate geometries (e.g. `Point`, empty); this is converted to `float('inf')` to match geopandas/shapely behaviour.

The method is implemented in `geoseries.py` (Spark SQL logic) and `base.py` (GeoDataFrame delegation + docstring with examples).

## How was this patch tested?

- `tests/geopandas/test_geoseries.py` — hardcoded expected-value tests including GeoDataFrame delegation:
  - `test_minimum_clearance`: unit square → `1.0`, half-unit square → `0.5`, degenerate `MultiPoint` → `inf`
- `tests/geopandas/test_match_geopandas_series.py` — comparison against geopandas:
  - `test_minimum_clearance`: iterates all geometry fixture types (`self.geoms`) and compares against geopandas results

All tests pass locally.

- No, this PR does not affect any public API documentation — it implements an existing stub method that was already part of the planned API surface (tracked in #2230). A docstring with examples is included inline.

### DIFF
--- a/python/sedona/spark/geopandas/base.py
+++ b/python/sedona/spark/geopandas/base.py
@@ -1129,8 +1129,32 @@ class GeoFrame(metaclass=ABCMeta):
         """
         return _delegate_to_geometry_column("minimum_bounding_radius", self)
 
-    # def minimum_clearance(self):
-    #     raise NotImplementedError("This method is not implemented yet.")
+    def minimum_clearance(self) -> ps.Series:
+        """Return a ``Series`` containing the minimum clearance distance,
+        which is the smallest distance by which a vertex of the geometry
+        could be moved to produce an invalid geometry.
+
+        If no minimum clearance exists for a geometry (for example,
+        a single point, or an empty geometry), infinity is returned.
+
+        Examples
+        --------
+        >>> from sedona.spark.geopandas import GeoSeries
+        >>> from shapely.geometry import Polygon, LineString, Point
+        >>> s = GeoSeries(
+        ...     [
+        ...         Polygon([(0, 0), (1, 1), (0, 1), (0, 0)]),
+        ...         LineString([(0, 0), (1, 1), (3, 2)]),
+        ...         Point(0, 0),
+        ...     ]
+        ... )
+        >>> s.minimum_clearance()
+        0    0.707107
+        1    1.414214
+        2         inf
+        dtype: float64
+        """
+        return _delegate_to_geometry_column("minimum_clearance", self)
 
     def normalize(self):
         """Return a ``GeoSeries`` of normalized geometries.
--- a/python/sedona/spark/geopandas/geoseries.py
+++ b/python/sedona/spark/geopandas/geoseries.py
@@ -15,6 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
+import sys
 import typing
 from typing import Any, Union, Literal, List
 
@@ -1077,9 +1078,14 @@ class GeoSeries(GeoFrame, pspd.Series):
             returns_geom=False,
         )
 
-    def minimum_clearance(self):
-        # Implementation of the abstract method.
-        raise NotImplementedError("This method is not implemented yet.")
+    def minimum_clearance(self) -> pspd.Series:
+        spark_col = stf.ST_MinimumClearance(self.spark.column)
+        # JTS returns Double.MAX_VALUE for degenerate geometries (e.g. Point, empty);
+        # convert to float('inf') to match geopandas/shapely behaviour.
+        spark_expr = F.when(
+            spark_col >= sys.float_info.max, F.lit(float("inf"))
+        ).otherwise(spark_col)
+        return self._query_geometry_column(spark_expr, returns_geom=False)
 
     def normalize(self):
         spark_expr = stf.ST_Normalize(self.spark.column)
--- a/python/tests/geopandas/test_geoseries.py
+++ b/python/tests/geopandas/test_geoseries.py
@@ -1614,7 +1614,20 @@ e": "Feature", "properties": {}, "geometry": {"type": "Point", "coordinates": [3
         self.check_pd_series_equal(df_result, expected)
 
     def test_minimum_clearance(self):
-        pass
+        s = GeoSeries(
+            [
+                Polygon([(0, 0), (1, 0), (1, 1), (0, 1)]),
+                Polygon([(0, 0), (0.5, 0), (0.5, 0.5), (0, 0.5)]),
+                MultiPoint([(1, 1), (1, 1)]),
+            ]
+        )
+        expected = pd.Series([1.0, 0.5, float("inf")])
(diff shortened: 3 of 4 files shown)


python/sedona/spark/geopandas/base.py
```diff
@@ -489,9 +489,36 @@ def is_ring(self):
...
+        is counter-clockwise oriented.
+
+        Returns False for non-polygon geometry types (e.g. LineString, Point).
+
+        Notes
+        -----
+        Unlike geopandas, which also supports LinearRing inputs, this
+        implementation uses Sedona's ``ST_IsPolygonCCW`` which only recognises
+        Polygon and MultiPolygon geometries. All other geometry types return
+        ``False``.
```
petern48: We actually had a previous contribution attempt for `is_ccw()`, and encountered this problem, and we agreed not to implement it for now (https://github.com/apache/sedona/pull/2386#discussion_r2422383890), given it's not easy to replicate the full desired behavior. We're not just missing proper behavior for LinearRings (which are rare), but we would have the incorrect behavior for `LineString`s, which are very common. As you can see in the code snippet below, `LineString` can evaluate to True here too, but our existing `ST_IsPolygonCCW()` isn't capable of determining that at the moment.

```python
import shapely
from shapely import LineString
from geopandas import GeoSeries

ls = LineString([[0, 0], [1, 0], [1, 1], [2, 2]])
gs = GeoSeries(ls)
print(gs.is_ccw)
```

In order to implement proper support for `is_ccw()`, I think we would need to investigate doing some more changes in the actual Java code and potentially look into the upstream JTS library. Given that complexity, I would rather skip implementing it entirely for now (or at least in a separate PR if you really want to investigate it, though it would be harder). Could we remove it for now?

piyushka-ally: Yes please.

No worries, I will try to pick some more functions I can implement. Thanks for reviewing my PR.

piyushka-ally (commented):
Please approve. Thanks.

petern48:
> Please approve. Thanks.

You haven't addressed the review feedback that I left (I don't see any new commits pushed). Do you plan to fulfill my request? I don't think this PR is ready to merge without it.

jiayuasu:
Note that all Python CIs are currently failing due to https://github.com/apache/sedona/issues/2774

We cannot merge this PR without all green python CIs. Meanwhile, please fix what Peter suggested.

piyushka-ally:
Hey sorry, I thought this was my other PR. I am working on this one, will push shortly.

piyushka-ally:
Pre-commit hooks pass locally, and I have removed the is_ccw() implementation. Thanks.

python/sedona/spark/geopandas/base.py
```diff
@@ -489,7 +489,6 @@ def is_ring(self):
         """
         return _delegate_to_geometry_column("is_ring", self)
 
-    # @property
     # def is_ccw(self):
```
petern48, line 493:
```suggestion
    # @property
    # def is_ccw(self):
```

Let's undo this deletion too, since we're doing one more iteration anyways.

python/sedona/spark/geopandas/geoseries.py
```diff
@@ -1081,9 +1078,14 @@ def minimum_bounding_radius(self) -> pspd.Series:
...
-    def minimum_clearance(self):
-        # Implementation of the abstract method.
-        raise NotImplementedError("This method is not implemented yet.")
+    def minimum_clearance(self) -> pspd.Series:
+        spark_col = stf.ST_MinimumClearance(self.spark.column)
+        # JTS returns Double.MAX_VALUE for degenerate geometries (e.g. Point, empty);
+        # convert to float('inf') to match geopandas/shapely behaviour.
+        spark_expr = F.when(
+            spark_col >= sys.float_info.max, F.lit(float("inf"))
+        ).otherwise(spark_col)
```
petern48, line 1087: I checked out this code and confirmed this extra conditional logic is necessary to pass tests 👍

(Just a note, no action needed by you)

python/sedona/spark/geopandas/base.py
```diff
@@ -1129,8 +1128,33 @@ def minimum_bounding_radius(self):
...
         return _delegate_to_geometry_column("minimum_bounding_radius", self)
 
-    # def minimum_clearance(self):
-    #     raise NotImplementedError("This method is not implemented yet.")
+    def minimum_clearance(self) -> ps.Series:
+        """Return the minimum clearance of each geometry.
+
+        The minimum clearance is the smallest distance by which a vertex of
+        a geometry could be moved to produce an invalid geometry.  A larger
+        value indicates a more robust geometry.
```
petern48: Could you replace this entire docstring with a copy-paste from the original geopandas one [here](https://github.com/geopandas/geopandas/blob/8762d4899f8c111fd57ac5e652e091ded73bdbaf/geopandas/base.py#L1951-L1982)? This is what we generally prefer to do. For this case, I like that their docstring mentions the infinity edge case behavior, and the examples they give are more helpful than the current ones that only show examples for polygons.

python/tests/geopandas/test_geoseries.py
```diff
@@ -1614,7 +1614,19 @@ def test_minimum_bounding_radius(self):
...
 
     def test_minimum_clearance(self):
-        pass
+        s = GeoSeries(
+            [
+                Polygon([(0, 0), (1, 0), (1, 1), (0, 1)]),
+                Polygon([(0, 0), (0.5, 0), (0.5, 0.5), (0, 0.5)]),
+            ]
+        )
+        expected = pd.Series([1.0, 0.5])
```
```suggestion
                Polygon([(0, 0), (0.5, 0), (0.5, 0.5), (0, 0.5)]),
                MultiPoint([(1, 1), (1, 1)]),
            ]
        )
        expected = pd.Series([1.0, 0.5, float("inf")])
```

I'd like to add this edge case, since it's not covered here or in `test_match_geopandas_series.py`. MultiPoint with identical points should return an inf, as Point would. Tested this locally, and already confirmed this test should pass.

piyushka-ally:
Sure let me work on it

piyushka-ally:
Done please recheck and let me know if any further changes are necessary. Thanks.

Issue #2830, opened 2026-04-08
Improve Geography query support in Apache Sedona


zhangfengcdt:
This project aims to improve the geography (spherical geometry) support to Apache Sedona by building OGC Simple Feature–compatible operators on top of the Google S2 Geometry Java library.

Functions:
- A Geography abstraction layer in Java backed by the S2 Java library
- A working set of OGC-compatible functions (intersects, contains, distance, etc.)
 
Spatial Joins:
- Geography-aware spatial join partitioning using S2 cells in Apache Sedona

Integration: 
- End-to-end integration in Sedona: SQL UDFs, tests, benchmarks
- Documentation and code examples

jiayuasu:
This issue has been mostly completed. The remaining issues are being tracked in separate tickets.

Issue #2810, opened 2026-03-30
checkWKBGeography compares geoWKT against itself instead of geoWKB


jiayuasu:
## Description

In `TestHelper.checkWKBGeography`, the comparison is `compareTo(geoWKT, geoWKT)` instead of `compareTo(geoWKB, geoWKT)`. This means the method always returns equal (comparing the same object), so the test never actually validates that WKB decoding matches the expected WKT.

## Root cause

Two bugs interact:

1. **Typo in `checkWKBGeography`** (`TestHelper.java:177`): `compareTo(geoWKT, geoWKT)` should be `compareTo(geoWKB, geoWKT)`.

2. **`compareTo` fails for empty geometries**: WKB and WKT readers produce incompatible `Geography` subtypes for empty geometries (e.g., `POINT EMPTY`). When the typo is fixed, `WKBReaderTest.EmptyTest` fails because `compareTo` cannot correctly compare these mismatched types.

## Steps to reproduce

Change line 177 of `TestHelper.java` from:
```java
boolean isEqual = compareTo(geoWKT, geoWKT) == 0;
```
to:
```java
boolean isEqual = compareTo(geoWKB, geoWKT) == 0;
```

Run `WKBReaderTest.EmptyTest` — it will fail with `AssertionError`.

## What needs to happen

- Fix `compareTo` to handle the case where WKB and WKT readers produce different `Geography` subtypes for empty geometries
- Fix the typo in `checkWKBGeography`
- Verify all `WKBReaderTest` cases pass with the corrected comparison

PR #2831, opened 2026-04-08
[GH-2830] Improve Geography query support - core

zhangfengcdt:

Implements WKB-based Geography serialization (Option B: WKB with Cached S2) and a full set of Geography ST functions.

**Core architecture:**
  - WKBGeography — stores WKB bytes as primary representation with lazy-parsed JTS, S2, and ShapeIndex caches (double-checked locking for thread safety)
  - GeographyWKBSerializer — WKB serializer with 0xFF format byte, backward-compatible with legacy S2-native format
  - GeographyUDT, implicits.scala, GeometrySerde — switched to WKBSerializer for all serialization paths

**Geography functions (3 new):**
  - Level 1 (JTS): ST_NPoints
  - Level 2 (JTS + Spheroid): ST_Distance
  - Level 3 (S2): ST_Contains

**Docs**: API docs for all 3 new functions in docs/api/sql/geography/

**Note**: Geography-aware spatial join partitioning using S2 cells will be in a separate PR

## How was this patch tested?

  - all unit tests pass in common module (WKBGeographyTest, FunctionTest)
  - GeographyFunctionTest.scala — Spark SQL integration tests covering constructors, structural functions, metrics, predicates, DataFrame API, and serialization round-trips


jiayuasu:
@zhangfengcdt Is this PR ready for review? It has lots of unnecessary content (e.g., the benchmark folder). Please also break this huge PR to several small pieces so we can review piece by piece.

zhangfengcdt:
> @zhangfengcdt Is this PR ready for review? It has lots of unnecessary content (e.g., the benchmark folder). Please also break this huge PR to several small pieces so we can review piece by piece.

I am still working on it. I will clean up the benchmark codes and also for this PR, it will focus on building the core architecture of the WKB-based Geography serialization with cached S2. I will keep a few core ST functions and tests and move other to individual PRs following the merging.

zhangfengcdt:
ST Function Performance: Geography vs Geometry (cached objects, ns/op)

 - ST_NPoints (Level 1 — JTS accessor)


```
  ┌─────────────────────┬───────────┬──────────┬───────┐
  │        Shape        │ Geography │ Geometry │ Ratio │
  ├─────────────────────┼───────────┼──────────┼───────┤
  │ Point               │         2 │        2 │    1x │
  ├─────────────────────┼───────────┼──────────┼───────┤
  │ LineString (16 vtx) │         2 │        2 │    1x │
  ├─────────────────────┼───────────┼──────────┼───────┤
  │ Polygon (16 vtx)    │         2 │        2 │    1x │
  ├─────────────────────┼───────────┼──────────┼───────┤
  │ Polygon (64 vtx)    │         2 │        2 │    1x │
  ├─────────────────────┼───────────┼──────────┼───────┤
  │ Polygon (500 vtx)   │         2 │        2 │    1x │
  └─────────────────────┴───────────┴──────────┴───────┘
```

 - ST_Distance (Level 2 — S2 geodesic distance)

```
  ┌─────────────────────┬───────────┬──────────┬───────┐
  │        Shape        │ Geography │ Geometry │ Ratio │
  ├─────────────────────┼───────────┼──────────┼───────┤
  │ Point               │       269 │       12 │   22x │
  ├─────────────────────┼───────────┼──────────┼───────┤
  │ LineString (16 vtx) │     1,576 │      373 │  4.2x │
  ├─────────────────────┼───────────┼──────────┼───────┤
  │ Polygon (16 vtx)    │     1,419 │      613 │  2.3x │
  ├─────────────────────┼───────────┼──────────┼───────┤
  │ Polygon (64 vtx)    │    69,279 │    3,874 │   18x │
  ├─────────────────────┼───────────┼──────────┼───────┤
  │ Polygon (500 vtx)   │   224,696 │  129,518 │  1.7x │
  └─────────────────────┴───────────┴──────────┴───────┘
```

 - ST_Contains (Level 3 — S2 predicate)

```
  ┌─────────────────────┬───────────┬──────────┬───────┐
  │        Shape        │ Geography │ Geometry │ Ratio │
  ├─────────────────────┼───────────┼──────────┼───────┤
  │ Point               │       284 │        8 │   36x │
  ├─────────────────────┼───────────┼──────────┼───────┤
  │ LineString (16 vtx) │       664 │        8 │   83x │
  ├─────────────────────┼───────────┼──────────┼───────┤
  │ Polygon (16 vtx)    │       684 │        8 │   86x │
  ├─────────────────────┼───────────┼──────────┼───────┤
  │ Polygon (64 vtx)    │       677 │        8 │   87x │
  ├─────────────────────┼───────────┼──────────┼───────┤
  │ Polygon (500 vtx)   │       703 │        8 │   88x │
  └─────────────────────┴───────────┴──────────┴───────┘


```

common/src/main/java/org/apache/sedona/common/S2Geography/WKBGeography.java
```diff
@@ -0,0 +1,229 @@
...
+      geog.getShapeIndexGeography();
+    }
+    return geog;
+  }
+
+  /** Create a WKBGeography from a JTS Geometry by serializing it to WKB. */
+  public static WKBGeography fromJTS(Geometry jts) {
+    org.locationtech.jts.io.WKBWriter writer =
+        new org.locationtech.jts.io.WKBWriter(
+            2, org.locationtech.jts.io.ByteOrderValues.BIG_ENDIAN);
```
paleolimbot: You probably want little endian here?

common/src/main/java/org/apache/sedona/common/S2Geography/WKBGeography.java
```diff
@@ -0,0 +1,229 @@
...
+  /** Returns an S2 Geography, lazily parsed from WKB on first access. */
+  public Geography getS2Geography() {
+    Geography result = s2Geography;
+    if (result == null) {
+      synchronized (this) {
+        result = s2Geography;
+        if (result == null) {
+          try {
+            WKBReader reader = new WKBReader();
+            result = reader.read(wkbBytes);
```
paleolimbot, line 173: I am guessing that this line is very expensive (it is in C++). The recent redo of s2geography in C++ that I just did is mostly about avoiding this line (e.g., by manually iterating over edges and using lower-level S2 primitives.

common/src/main/java/org/apache/sedona/common/S2Geography/WKBGeography.java
```diff
@@ -0,0 +1,229 @@
...
+    }
+    return result;
+  }
+
+  // --- Geography abstract method delegation to lazy S2 ---
+
+  @Override
+  public int dimension() {
+    return getS2Geography().dimension();
+  }
```
paleolimbot, line 241: This one you can get from the first few WKB bytes

common/src/main/java/org/apache/sedona/common/S2Geography/WKBGeography.java
```diff
@@ -0,0 +1,229 @@
...
+
+  @Override
+  public int numShapes() {
+    return getS2Geography().numShapes();
+  }
+
+  @Override
+  public S2Shape shape(int id) {
+    return getS2Geography().shape(id);
+  }
```
paleolimbot, line 257: I optimized this piece with a few hundred lines of code implementing the S2Shape interface on top of WKB. If you have this, you can avoid the `Geography` entirely (because actual functions usually need a ShapeIndex).

common/src/main/java/org/apache/sedona/common/S2Geography/WKBGeography.java
```diff
@@ -0,0 +1,229 @@
...
+
+  @Override
+  public S2Region region() {
+    return getS2Geography().region();
+  }
```
paleolimbot, line 265: I optimized this one for points using the S2PointRegion for points to avoid constructing Geography. That might not matter here (in general when implementing the functions I also made every attempt to avoid accessing the Region to avoid the index build)

common/src/main/java/org/apache/sedona/common/S2Geography/WKBGeography.java
```diff
@@ -0,0 +1,229 @@
...
+
+  @Override
+  public void getCellUnionBound(List<com.google.common.geometry.S2CellId> cellIds) {
+    getS2Geography().getCellUnionBound(cellIds);
+  }
```
paleolimbot, line 274: This one can be avoided for a Point by just returning its cell

common/src/main/java/org/apache/sedona/common/geography/Functions.java
```diff
@@ -75,14 +71,69 @@ private static S2Polygon rectToPolygon(double lngLo, double latLo, double lngHi,
...
+   * Geometry-to-geometry geodesic distance in meters. Uses S2ClosestEdgeQuery for true minimum
+   * distance between any two points on the geometries (not centroid-to-centroid). Consistent with
+   * sedona-db's s2_distance implementation.
+   */
+  public static Double distance(Geography g1, Geography g2) {
+    if (g1 == null || g2 == null) return null;
+    Distance dist = new Distance();
+    double radians = dist.S2_distance(toShapeIndex(g1), toShapeIndex(g2));
+    return radiansToMeters(radians);
+  }
```
paleolimbot, line 118: It was a bit of a rabbit hole, but this is now several hundred lines in C++ to avoid building the shape index in all cases except (1) very long linestrings and (2) polygons

common/src/main/java/org/apache/sedona/common/geography/Functions.java
```diff
@@ -75,14 +71,69 @@ private static S2Polygon rectToPolygon(double lngLo, double latLo, double lngHi,
...
+
+  // ─── Level 3: S2 spherical predicates ────────────────────────────────────
+
+  /** Spherical containment test using S2 boolean operations. */
+  public static boolean contains(Geography g1, Geography g2) {
+    if (g1 == null || g2 == null) return false;
+    Predicates pred = new Predicates();
+    return pred.S2_contains(toShapeIndex(g1), toShapeIndex(g2), s2Options());
+  }
```
paleolimbot, line 130: This one has a similar array of optimizations for small geometries, including for small polygons (I probably should have put the small polygon optimization in the distance path as well but I ran out of will)

paleolimbot (commented):
This seems like it is headed in the right direction!

The `WKBGeography` is the right direction I think (In C++ I call this a GeoArrowGeography and it's slightly more general but the same idea). As written, I am not sure it is able to make anything faster (I am guessing that Spark already handles `filter()`s and won't materialize a Java object unless it will actually be used for something).

This may not be true in Java, but in C++ the S2Polygon and maybe S2Polyline has an internal index for itself and also one for each loop (lazily constructed, but almost always needed for initializing from WKB to figure out the nesting of shells/holes). The optimization of implementing the S2Shape interface directly on WKB is pretty much 100% to avoid those extra index builds (so that for any given function there's at most one shape index per object that is built). I don't know how flexible the S2Shape is in Java (in C++ implementing it was not too bad). I think you would need to do something similar to see improved performance (but maybe the first thing you want to do is get the functions and tests in place).

zhangfengcdt:
> This seems like it is headed in the right direction!
> 
> The `WKBGeography` is the right direction I think (In C++ I call this a GeoArrowGeography and it's slightly more general but the same idea). As written, I am not sure it is able to make anything faster (I am guessing that Spark already handles `filter()`s and won't materialize a Java object unless it will actually be used for something).
> 
> This may not be true in Java, but in C++ the S2Polygon and maybe S2Polyline has an internal index for itself and also one for each loop (lazily constructed, but almost always needed for initializing from WKB to figure out the nesting of shells/holes). The optimization of implementing the S2Shape interface directly on WKB is pretty much 100% to avoid those extra index builds (so that for any given function there's at most one shape index per object that is built). I don't know how flexible the S2Shape is in Java (in C++ implementing it was not too bad). I think you would need to do something similar to see improved performance (but maybe the first thing you want to do is get the functions and tests in place).

@paleolimbot Thanks for the review and they are really helpful.  I have implemented some of the optimizations. Here's what changed and the benchmark results.

- use little endian wkb for fast w/r
- get dim from wkb bytes directly
- add new WkbS2Shape and pre-convert wkb coordinate to S2Point (cached)
- use regions for points
- get cell union for points
- point to point distance fast path
- ShapeIndex is built from WkbS2Shape

```
  ST_NPoints (Level 1 — JTS accessor)

  ┌───────────────────┬────────┬───────┬────────┐
  │       Shape       │ Before │ After │ Change │
  ├───────────────────┼────────┼───────┼────────┤
  │ Point             │      2 │     2 │  ~same │
  ├───────────────────┼────────┼───────┼────────┤
  │ Polygon (16 vtx)  │      2 │     2 │  ~same │
  ├───────────────────┼────────┼───────┼────────┤
  │ Polygon (500 vtx) │      2 │     2 │  ~same │
  └───────────────────┴────────┴───────┴────────┘

  ST_Distance (Level 2 — S2 geodesic)

  ┌─────────────────────┬─────────┬─────────┬───────────────────────────────┐
  │        Shape        │ Before  │  After  │            Change             │
  ├─────────────────────┼─────────┼─────────┼───────────────────────────────┤
  │ Point               │     269 │     245 │ 1.1x faster (point fast path) │
  ├─────────────────────┼─────────┼─────────┼───────────────────────────────┤
  │ LineString (16 vtx) │   1,576 │   1,575 │                         ~same │
  ├─────────────────────┼─────────┼─────────┼───────────────────────────────┤
  │ Polygon (16 vtx)    │   1,419 │   2,248 │                   1.6x slower │
  ├─────────────────────┼─────────┼─────────┼───────────────────────────────┤
  │ Polygon (64 vtx)    │  69,279 │  64,515 │                  1.07x faster │
  ├─────────────────────┼─────────┼─────────┼───────────────────────────────┤
  │ Polygon (500 vtx)   │ 224,696 │ 218,689 │                  1.03x faster │
  └─────────────────────┴─────────┴─────────┴───────────────────────────────┘
```

The major improvement: WkbS2Shape avoids building 3 redundant S2ShapeIndex instances (one per S2Loop, one for S2Polygon, and only the ShapeIndexGeography one is actually used).

```
  ST_Contains true (Level 3 — S2 predicate, point inside polygon)

  ┌─────────────────────┬────────┬───────┬─────────────┐
  │        Shape        │ Before │ After │   Change    │
  ├─────────────────────┼────────┼───────┼─────────────┤
  │ Point               │    284 │   248 │ 1.1x faster │
  ├─────────────────────┼────────┼───────┼─────────────┤
  │ LineString (16 vtx) │    664 │   238 │ 2.8x faster │
  ├─────────────────────┼────────┼───────┼─────────────┤
  │ Polygon (16 vtx)    │    684 │   239 │ 2.9x faster │
  ├─────────────────────┼────────┼───────┼─────────────┤
  │ Polygon (64 vtx)    │    677 │   233 │ 2.9x faster │
  ├─────────────────────┼────────┼───────┼─────────────┤
  │ Polygon (500 vtx)   │    703 │   254 │ 2.8x faster │
  └─────────────────────┴────────┴───────┴─────────────┘
```

For st_contains false case, it becomes slower because S2BooleanOperation.Contains() has different internal execution paths for true vs false results. Not sure if sedona-db also does this (S2Polygon for polygon predicates), which may see similar tradeoffs. But here is the results:

```
  ST_Contains false (Level 3 — S2 predicate, point outside polygon)

  ┌─────────────────────┬────────┬───────┬─────────────┐
  │        Shape        │ Before │ After │   Change    │
  ├─────────────────────┼────────┼───────┼─────────────┤
  │ Point               │    226 │   223 │       ~same │
  ├─────────────────────┼────────┼───────┼─────────────┤
  │ LineString (16 vtx) │    220 │   653 │   3x slower │
  ├─────────────────────┼────────┼───────┼─────────────┤
  │ Polygon (16 vtx)    │    222 │   657 │   3x slower │
  ├─────────────────────┼────────┼───────┼─────────────┤
  │ Polygon (64 vtx)    │    222 │   657 │   3x slower │
  ├─────────────────────┼────────┼───────┼─────────────┤
  │ Polygon (500 vtx)   │    246 │   692 │ 2.8x slower │
  └─────────────────────┴────────┴───────┴─────────────┘

```

common/src/main/java/org/apache/sedona/common/S2Geography/WKBGeography.java
zhangfengcdt: We now avoid this for Point, LineString, and Polygon by implementing WkbS2Shape (an S2Shape on pre-converted S2Point arrays) and building the ShapeIndex directly from it.

common/src/main/java/org/apache/sedona/common/S2Geography/WKBGeography.java
zhangfengcdt: The new class WkbS2Shape implements S2Shape for Point, LineString, and Polygon.

common/src/main/java/org/apache/sedona/common/S2Geography/WKBGeography.java
zhangfengcdt: region() now returns new S2PointRegion(extractPoint()) for Point WKB

common/src/main/java/org/apache/sedona/common/geography/Functions.java
zhangfengcdt: Now, it has the similar logic for the ShapeIndex is now built from WkbS2Shape instead of S2 Geography. For large polygons, it runs slightly faster.

zhangfengcdt:
> Cool!
> 
> In s2geography there's quite a bit of code to avoid the shape index build at all for ST_Distance() and each individual predicate, which may bring your results closer to on par. Notably, there's the possible intersection check which should help the "point is not in polygon" case.

Thanks for pointing to the s2geography optimizations. I implemented the following two optimization for now:

1. point container to false directly early stop
2. pointTarget for distance: avoids building ShapeIndex for the point side

I also tried to implement the MayIntersect that might help with the contains false case, but looks like the MayIntersects path (~700ns) is more expensive than the cached ShapeIndex path (~220ns) in a single call. 

In this PR, I was thinking to optimize the hot path, which may benefit more with cached ShapeIndex (good for joins). For the code path optimizations, adding mayIntersects pre-filters may be better; potentially adding checking covering intersection pre-filtering as well. I'd prefer to add them in following PRs with a configurable flags to avoid the hot path gets slower. 

There seems to be tradeoffs we need to make default and also custom sedona configs for users to tune these optimizations since the geography has too much overhead compared to the geometry functions. 

What do you think?

zhangfengcdt:
> That sounds good to me! I the work I did in C++ was mostly to expose all of the possible optimizations but I haven't tuned them yet (and the tuning is probably much different in Java than it is in C++). The tuning/extreme optimization is also much better done in the presence of integration testing against BigQuery/PostGIS (which is in the works on the SedonaDB side). The point optimizations are nice because they truly are easy outs (and point/polygon is I think common).
> 
> One of the things that may have affected my initial implementation is that SedonaDB benchmarks mostly have one Scalar argument (where the cost of computing the covering is amortized over many more rows). I didn't implement a S2LatLngRect bound comparison but I think computing that might be cheaper.

Yes, one scalar case is a very popular case and the ShapeIndex cache and/or covering will help a lot in performance. The join also benefit from the cache and covering.

zhangfengcdt:
@jiayuasu Could you take a look again? I think the core is ready and I'd create a couple of following up PRs for different geography ST functions once this core PR is merged.

I'd also like to add a few optimization configs once all of the st functions are merged so we can measure the performance under different optimizations in all.

Thanks!

docs/api/sql/geography/Geography-Functions/ST_Contains.md
```diff
@@ -0,0 +1,45 @@
...
+
+Introduction: Tests whether geography A fully contains geography B using S2 spherical boolean operations. Returns true if every point of B is inside or on the boundary of A.
+
+Format:
+
+`ST_Contains (A: Geography, B: Geography)`
+
+Return type: `Boolean`
+
+Since: `v1.9.0`
```
jiayuasu, line 30: Let's make it 1.9.1. I don't think the geography function will come in on time with 1.9.0

common/src/main/java/org/apache/sedona/common/S2Geography/GeographyWKBSerializer.java
```diff
@@ -0,0 +1,84 @@
...
+package org.apache.sedona.common.S2Geography;
+
+import java.io.IOException;
+import java.util.Arrays;
+import org.locationtech.jts.io.ByteOrderValues;
+
+/**
+ * Serializer for Geography objects using WKB as the primary format. Supports backward-compatible
+ * reading of legacy S2-native serialized data. Format discrimination by first byte: 0xFF for WKB
+ * format [0xFF][4-byte SRID big-endian][WKB payload], or 1-10 for legacy S2-native format.
```
jiayuasu, line 28: I suggest we remove the backward compatibility of the S2 native format unless it helps with the performance. This will reduce the complexity of the codebase. The S2 SerDe format is never announced and we don't expect active users of this feature.

common/src/main/java/org/apache/sedona/common/geography/Functions.java
```diff
@@ -75,14 +71,91 @@ private static S2Polygon rectToPolygon(double lngLo, double latLo, double lngHi,
...
+    }
+    return new ShapeIndexGeography(g);
+  }
+
+  private static S2BooleanOperation.Options s2Options() {
+    return new S2BooleanOperation.Options();
+  }
+
+  /** Mean Earth radius in meters (WGS84 mean radius). */
+  private static final double EARTH_RADIUS_METERS = 6371008.8;
```
jiayuasu, line 156: There is another place in Sedona codebase (Distance Spheroid / Sphere) also hard code the value of earth radius. can you unify them together?

docs/api/sql/geography/Geography-Functions/ST_Contains.md
```diff
@@ -0,0 +1,45 @@
...
+
+SQL Example
+
+```sql
+SELECT ST_Contains(
```
jiayuasu, line 35: Can you add SVG visuals for this too? See Geometry ST_Contains: https://sedona.apache.org/latest/api/sql/Predicates/ST_Contains/

jiayuasu:
There is a bug in the geography test helper which never actually validates the correctness of the code: https://github.com/apache/sedona/issues/2810

Is it still true in this PR?

common/src/main/java/org/apache/sedona/common/S2Geography/WKBGeography.java
```diff
@@ -0,0 +1,310 @@
...
+              | ((wkbBytes[4] & 0xFF) << 24);
+    } else {
+      raw =
+          ((wkbBytes[1] & 0xFF) << 24)
+              | ((wkbBytes[2] & 0xFF) << 16)
+              | ((wkbBytes[3] & 0xFF) << 8)
+              | (wkbBytes[4] & 0xFF);
+    }
+    return raw & 0xFF;
+  }
```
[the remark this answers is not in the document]
zhangfengcdt: Replaced raw & 0xFF with (typeInt & 0xffff) % 1000 (matches WKBReader.java pattern)

common/src/main/java/org/apache/sedona/common/S2Geography/WkbS2Shape.java
```diff
@@ -0,0 +1,267 @@
...
+          this.containsOriginValue = false;
+          break;
+        }
+
+      case 2: // LineString
+        {
+          this.dim = 1;
+          int numCoords = buf.getInt(5);
+          this.vertices = readVertices(buf, 9, numCoords);
+          this.totalEdges = Math.max(0, numCoords - 1);
```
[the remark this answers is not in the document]
zhangfengcdt: Added EWKB SRID-flag detection (0x20000000) that shifts coordinate/count offsets by 4 bytes

common/src/main/java/org/apache/sedona/common/S2Geography/WKBGeography.java
```diff
@@ -0,0 +1,310 @@
...
+
+  /** Extract the S2Point from a Point WKB without full S2 parse. */
+  public S2Point extractPoint() {
+    boolean le = (wkbBytes[0] == 0x01);
+    ByteBuffer bb =
+        ByteBuffer.wrap(wkbBytes).order(le ? ByteOrder.LITTLE_ENDIAN : ByteOrder.BIG_ENDIAN);
+    double lon = bb.getDouble(5);
+    double lat = bb.getDouble(13);
+    return S2LatLng.fromDegrees(lat, lon).toPoint();
```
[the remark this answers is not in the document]
zhangfengcdt: Added explicit Z/M rejection

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/JoinQueryDetector.scala
```diff
@@ -209,6 +199,16 @@ class JoinQueryDetector(sparkSession: SparkSession) extends SparkStrategy {
...
+            case ST_Contains(Seq(leftShape, rightShape)) =>
+              Some(
+                JoinQueryDetection(
+                  left,
+                  right,
+                  leftShape,
+                  rightShape,
+                  SpatialPredicate.CONTAINS,
+                  false,
+                  extraCondition))
```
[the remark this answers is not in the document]
zhangfengcdt: The spatial join on geography type/udfs will be a follow up pr, so we will just fall back to spark (sedona)'s default here for now.

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/FunctionResolver.scala
```diff
@@ -82,14 +82,10 @@ object FunctionResolver {
...
-              "  (" + function.sparkInputTypes.mkString(", ") + ")"
-            }
-            .mkString("\n")
-          throw new IllegalArgumentException(
-            "Ambiguous function call. Candidates are: \n" + candidateTypesMsg)
+          // Multiple candidates match equally (e.g., null inputs matching both Geometry and
+          // Geography overloads). Prefer the first candidate — for null inputs all overloads
+          // return null, so the choice is irrelevant.
+          ambiguousMatches.head._1
         }
```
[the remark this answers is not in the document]
zhangfengcdt: Silent first-candidate selection only fires when every argument is NullType; otherwise throws IllegalArgumentException with a candidate-list message.

common/src/main/java/org/apache/sedona/common/S2Geography/WkbS2Shape.java
```diff
@@ -0,0 +1,267 @@
...
+
+  // For polygon containsOrigin — computed eagerly at construction for polygons
+  private final boolean containsOriginValue;
+
+  public WkbS2Shape(byte[] wkb) {
+    boolean le = (wkb[0] == 0x01);
+    ByteBuffer buf =
+        ByteBuffer.wrap(wkb).order(le ? ByteOrder.LITTLE_ENDIAN : ByteOrder.BIG_ENDIAN);
+    int wkbType = buf.getInt(1) & 0xFF;
+
```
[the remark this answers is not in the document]
zhangfengcdt: this should be addressed now by using the same WKBReader.java pattern

docs/api/sql/geography/Geography-Functions/ST_Contains.md
zhangfengcdt: Yes, all three new md files now state 1.9.1.

common/src/main/java/org/apache/sedona/common/S2Geography/GeographyWKBSerializer.java
zhangfengcdt: I have removed the s2-native backward compatibility.

docs/api/sql/geography/Geography-Functions/ST_Contains.md
zhangfengcdt: Yes, I have added the svg illustrations to all three st functions added.

zhangfengcdt:
> There is a bug in the geography test helper which never actually validates the correctness of the code: #2810
> 
> Is it still true in this PR?

Yes, the bug was still present. I just fixed it in this PR.

The TestHelper.java now shows "compareTo(geoWKB, geoWKT)" and compareTo — added early-return for same-class empty geometries before the kind comparison.

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/JoinQueryDetector.scala
```diff
@@ -54,40 +54,47 @@ case class JoinQueryDetection(
...
+    shape.dataType.isInstanceOf[GeographyUDT]
+
   private def getJoinDetection(
       left: LogicalPlan,
       right: LogicalPlan,
       predicate: ST_Predicate,
       extraCondition: Option[Expression] = None): Option[JoinQueryDetection] = {
     predicate match {
-      case ST_Contains(Seq(leftShape, rightShape)) =>
+      case ST_Intersects(Seq(leftShape, rightShape)) =>
```
jiayuasu, line 72: Why does ST_Contains get changed here?

jiayuasu (changes requested):

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/strategy/join/JoinQueryDetector.scala
jiayuasu: And other predicates do not seem to change? @zhangfengcdt

zhangfengcdt: Yes, the ST_Contains moved out of getJoinDetection because it was converted to InferredExpression in this PR (dual-dispatch JTS + S2 Geography). That removed it from the ST_Predicate class hierarchy, so it no longer is type-checked.

Once we do the same geography function changes for other predicates, we will move them as well.

jiayuasu (approved):

PR #2849, opened 2026-04-22
[GH-2830] Add Geography dual-dispatch to ST_AsText

zhangfengcdt:

  - Adds Geography dual-dispatch to ST_AsText — now returns WKT for both Geometry and Geography inputs.
  - Follow-up to #2831; L1 tier (Geography path delegates to JTS via getJTSGeometry().toText()).

## How was this patch tested?
  - mvn -pl common -am test -Dtest=FunctionTest
  - New Spark SQL case in GeographyFunctionTest

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/geography/Functions.java
+++ b/common/src/main/java/org/apache/sedona/common/geography/Functions.java
@@ -85,6 +85,12 @@ public class Functions {
     return toJTS(g).getNumPoints();
   }
 
+  /** Return the WKT text representation of a geography. */
+  public static String asText(Geography g) {
+    if (g == null) return null;
+    return toJTS(g).toText();
+  }
+
   // ─── Level 2: JTS + S2 geodesic metrics ──────────────────────────────────
 
   /**
--- a/spark/common/src/test/scala/org/apache/sedona/sql/geography/GeographyFunctionTest.scala
+++ b/spark/common/src/test/scala/org/apache/sedona/sql/geography/GeographyFunctionTest.scala
@@ -23,7 +23,8 @@ import org.apache.sedona.sql.TestBaseScala
 import org.apache.spark.sql.functions.{col, lit}
 import org.apache.spark.sql.sedona_sql.expressions.{st_constructors, st_functions, st_predicates}
 import org.junit.Assert.{assertEquals, assertNotNull, assertTrue}
-import org.locationtech.jts.geom.Geometry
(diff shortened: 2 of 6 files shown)


spark/common/src/test/scala/org/apache/sedona/sql/geography/GeographyFunctionTest.scala
```diff
@@ -87,6 +87,15 @@ class GeographyFunctionTest extends TestBaseScala {
...
     }
+
+    it("ST_AsText") {
+      val row = sparkSession
+        .sql("SELECT ST_AsText(ST_GeogFromWKT('POINT (1 2)', 4326)) AS wkt")
+        .first()
+      val wkt = row.getString(0)
+      // S2 round-trip may introduce sub-nanometer floating-point drift
+      assertTrue(s"Expected POINT near (1, 2), got: $wkt", wkt.startsWith("POINT ("))
+    }
```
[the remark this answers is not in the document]
zhangfengcdt: Now, ST_AsText case — parses the WKT and asserts point.getX == 1.0, point.getY == 2.0 within 1e-9

common/src/main/java/org/apache/sedona/common/geography/Functions.java
```diff
@@ -85,6 +85,12 @@ public static int nPoints(Geography g) {
     return toJTS(g).getNumPoints();
   }
 
+  /** Return the WKT text representation of a geography. */
+  public static String asText(Geography g) {
+    if (g == null) return null;
+    return toJTS(g).toText();
```
jiayuasu, line 91: @zhangfengcdt This toJTS function seem to SerDe WKB one more time. Is it possible to avoid it?

zhangfengcdt: Good question! toJTS(g) on WKBGeography does one WKB → JTS parse and caches the result in jtsGeometry (WKBGeography.java:116). Subsequent calls — both later asText invocations on the same row and any other JTS-path function (ST_NPoints, ST_GeometryType, ST_Centroid, ST_NumGeometries) — reuse the cached Geometry for free.

The alternative — g.toString() on the Geography — actually goes through WKBGeography.toString() getS2Geography().toString(), which populates the other cache (s2Geography), so it's worse when the current implementation.

A true zero-parse path (WKB → WKT walker) would need custom code since JTS doesn't expose it.

jiayuasu (approved):

PR #2852, opened 2026-04-22
[GH-2830] Adds Geography dual-dispatch to ST_Centroid

zhangfengcdt:

  - Adds Geography dual-dispatch to ST_Centroid — returns a Geography point for Geography inputs (planar centroid, computed in projected lon/lat space, not on the sphere).
 - Use S2-spherical dispatcher S2 returns weighted (non-unit) centroids; final step normalizes via S2PAoint.normalize. Returns null if the centroid is undefined (empty input or vectors that cancel).

## How was this patch tested?
  - mvn clean test -pl common -am -Dtest=FunctionTest
  - New Spark SQL case in GeographyFunctionTest
  - mvn spotless:apply run before commit; no format violations remain.


jiayuasu:
@zhangfengcdt Are you sure this should go back to JTS and use planar geometry to get the centroid? Does S2 have something spherical?

zhangfengcdt:
> @zhangfengcdt Are you sure this should go back to JTS and use planar geometry to get the centroid? Does S2 have something spherical?

Yes, you're right! I switched it to S2, need a bit more work on that though - S2 returns a non-unit weighted vector, so the result is normalized via S2Point.normalize before being wrapped as SinglePointGeography.

jiayuasu (approved):

Issue #2877, opened 2026-05-01
Add a native Box2D type for bounding boxes


jiayuasu:
**Phase 1 complete.** All sub-issues are closed. The Box2D type, its UDT, the SQL surface, cross-language bindings (Python, Flink, Scala DataFrame API), and GeoParquet writer recognition all landed in master.

## Summary of what shipped

- `Box2D` value class + `Box2DUDT` (struct-backed `xmin/ymin/xmax/ymax: double`)
- SQL: `ST_Box2D`, `ST_MakeBox2D`, `ST_Extent` (aggregate), `ST_GeomFromBox2D`, `ST_AsText(box2d)`, accessor overloads `ST_XMin/XMax/YMin/YMax(box2d)`
- Python bindings (PySpark `Box2DType`, all wrappers in `st_functions` / `st_constructors` / `st_aggregates`)
- Scala DataFrame API wrappers in `st_functions` / `st_constructors` / `st_aggregates`
- Flink bindings (scalar + aggregate + `Box2DTypeSerializer`)
- GeoParquet 1.1: Box2D columns are recognized as bbox covering columns in metadata (both explicit `geoparquet.covering[.geom]=<col>` and auto-detected `<geom>_bbox`)

### Foundation

- [x] **#2878** — Box2D value class + Box2DUDT (struct-backed) + UDT registration + Python `Box2DType`

### SQL surface

- [x] **#2881** — `ST_Box2D(geom)` scalar
- [x] **#2882** — `ST_XMin / XMax / YMin / YMax(box2d)` accessor overloads
- [x] **#2883** — `ST_MakeBox2D(p1, p2)` scalar constructor
- [x] **#2884** — `ST_Extent(geom)` aggregate
- [x] **#2885** — `ST_GeomFromBox2D` and `ST_AsText(box2d)`

### Storage

- [x] **#2886** — Box2D columns recognized as GeoParquet 1.1 bbox covering columns in metadata

### Cross-language bindings

- [x] **#2887** — Python bindings for the Phase 1 surface
- [x] **#2888** — Flink bindings for the Phase 1 surface
- [x] **#2891** — Scala DataFrame API wrappers (folded into #2904 once it surfaced as a hard dependency for the Python bindings)

## Deferred follow-ups

These were explicitly scoped out of Phase 1 and will need their own issues if/when prioritized:

[...]

PR #2878, opened 2026-05-01
[GH-2877] Add Box2D type and Box2DUDT

jiayuasu:
## Summary

Adds the `Box2D` value type and its UDT, the foundation for the bbox work tracked in #2877. Functions (`ST_Box2D`, `ST_MakeBox2D`, `ST_Extent`, accessor overloads, casts) follow in subsequent PRs.

- `common/.../geometryObjects/Box2D.java` — planar 2D bounding box. Always a valid finite bbox; absence of a bbox is represented by SQL NULL at the column level (PostGIS-compatible). `xmin > xmax` is intentionally **not** used as an in-band empty marker so it remains free for a future antimeridian-wraparound semantics on geography bboxes (cf. `apache/sedona-db`'s `WraparoundInterval`).
- `spark/common/.../UDT/Box2DUDT.scala` — struct-backed UDT with `sqlType = struct<xmin, ymin, xmax, ymax>` (all `double`, non-nullable). Struct-backed (not binary-backed) so values round-trip natively to Parquet and align zero-copy with GeoParquet 1.1 bbox covering columns.
- `spark/common/.../UDT/UdtRegistratorWrapper.scala` — register `Box2D ↔ Box2DUDT`.
- `python/sedona/spark/...` — matching `Box2DType` UDT and `Box2D` value class so a Box2D column materialized in PySpark resolves cleanly.
- `Box2DUDTSuite` covers UDT registration, JSON schema round-trip, serde round-trip, case-object equality, and Parquet write/read of a Box2D column.

Field names (`xmin/ymin/xmax/ymax`) match the GeoParquet 1.1 spec and `apache/sedona-db`'s GeoParquet writer for direct cross-engine interop.

## Test plan

- [x] CI builds the `common` and `spark/common` modules cleanly.
- [x] JSON schema round-trip via `Box2DUDTSuite`.
- [x] Box2D struct serde round-trip via `Box2DUDTSuite`.
- [x] Parquet write/read of a Box2D column via `Box2DUDTSuite`.
- [x] Python `Box2DType` round-trips through `serialize` / `deserialize` (smoke-tested locally).
- [x] `UdtRegistratorWrapper.registerAll()` registers Box2D (asserted in `Box2DUDTSuite`).

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/geometryObjects/Box2D.java
+++ b/common/src/main/java/org/apache/sedona/common/geometryObjects/Box2D.java
@@ -27,8 +27,11 @@ import org.locationtech.jts.geom.GeometryFactory;
 import org.locationtech.jts.geom.Polygon;
 
 /**
- * Planar 2D bounding box with min/max X and Y. Empty boxes are encoded as {@code xmin > xmax ||
- * ymin > ymax} (JTS Envelope convention), which makes union/expand a no-op against the empty value.
+ * Planar 2D bounding box with min/max X and Y. Always a valid finite bbox; absence of a bbox (e.g.
+ * bbox of an empty geometry, extent over zero rows) is represented by SQL NULL at the column level
+ * rather than by an in-band sentinel. This matches PostGIS behavior and leaves {@code xmin > xmax}
+ * free for a future antimeridian-wraparound semantics on geography bboxes (cf. sedona-db's {@code
+ * WraparoundInterval}).
  */
 public final class Box2D implements Serializable {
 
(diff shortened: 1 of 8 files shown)


jiayuasu:
@zhangfengcdt @paleolimbot what do you think of this?

common/src/main/java/org/apache/sedona/common/geometryObjects/Box2D.java
```diff
@@ -0,0 +1,137 @@
...
+import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.geom.GeometryFactory;
+import org.locationtech.jts.geom.Polygon;
+
+/**
+ * Planar 2D bounding box with min/max X and Y. Always a valid finite bbox; absence of a bbox (e.g.
+ * bbox of an empty geometry, extent over zero rows) is represented by SQL NULL at the column level
+ * rather than by an in-band sentinel. This matches PostGIS behavior and leaves {@code xmin > xmax}
+ * free for a future antimeridian-wraparound semantics on geography bboxes (cf. sedona-db's {@code
+ * WraparoundInterval}).
```
paleolimbot, line 31: I think you will be happier "allowing" this from the start and erroring when you see `xmin > xmax` anywhere it matters for a computation (e.g., `expandToInclude()`. In other words, the Box2D always allows antimeridian crossing-like behaviour (but not everything has to be implemented yet). I think this will make it easier to avoid accidental incorrect behaviour when those operations do support it.

paleolimbot (approved):
I can't speak to the Spark details but the definition looks good to me!

It also matches the GeoArrow naming and definition of the box type, which is what we'll match this with in SedonaDB: https://geoarrow.org/format.html#box

zhangfengcdt:
Would be clearer if we just name it BOXUDT, with the possibility to extend to z and m dimensions later on (if needed)? The parquet bbox does not limit to 2D scenario.

jiayuasu:
> Would be clearer if we just name it BOXUDT, with the possibility to extend to z and m dimensions later on (if needed)? The parquet bbox does not limit to 2D scenario.

@zhangfengcdt Yes, there will be a BOX3D type. This is to maintain compatibility with PostGIS box2d and box3d

zhangfengcdt:
> > Would be clearer if we just name it BOXUDT, with the possibility to extend to z and m dimensions later on (if needed)? The parquet bbox does not limit to 2D scenario.
> 
> @zhangfengcdt Yes, there will be a BOX3D type. This is to maintain compatibility with PostGIS box2d and box3d

Got it, make sense to me.

zhangfengcdt (approved):

PR #2962, opened 2026-05-16
[GH-2617] Reorder raster function signatures

### LINKED ISSUE #2617: docs: Rearrange order of function headers in Raster Operation docs to be more intuitive
While looking at https://github.com/apache/sedona/pull/2616, I noticed many function headers in the [Raster Operations docs](https://github.com/apache/sedona/blob/aeb88221871a1b2f5f4604bbb36a683093c38e15/docs/api/sql/Raster-operators.md?plain=1#L1030) are listed in order from most to least function arguments. In general, I think it's better to order it from least to most because that reading order is more intuitive to see what parameters are optional and often puts the most commonly used function header first.

Here's an example of the current order of RS_Count, for example.
```
RS_Count(raster: Raster, band: Integer = 1, excludeNoDataValue: Boolean = true)

RS_Count(raster: Raster, band: Integer = 1)

RS_Count(raster: Raster)
```

I think it's better to instead list it like this:

```
RS_Count(raster: Raster)

RS_Count(raster: Raster, band: Integer = 1)

RS_Count(raster: Raster, band: Integer = 1, excludeNoDataValue: Boolean = true)
```

List of functions that I noticed had this problem (there could be more):
- RS_AddBand
- RS_Count
- RS_Resample
- RS_SetBandNoDataValue
- RS_ZonalStats
- RS_ZonalStatsAll
- etc. (there are certainly more)

Interestingly, some functions like `RS_NormalizeAll ` and `RS_Interpolate` are already in the desired order.

omribz156:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #2617

This reorders raster function format signatures from the shortest form to the longest form, so optional parameters are introduced progressively.

The patch keeps the change docs-only and updates the raster operator/band accessor pages where the reversed ordering still existed.

## How was this patch tested?

- `git diff --check`
- Static format-block check that the changed `RS_*` signatures are ordered by increasing argument count

This was implemented with Codex assistance, with the final diff reviewed before posting.

### DIFF
--- a/docs/api/sql/Raster-Band-Accessors/RS_Count.md
+++ b/docs/api/sql/Raster-Band-Accessors/RS_Count.md
@@ -30,11 +30,11 @@ Introduction: Returns the number of pixels in a given band. If band is not speci
 
 Format:
 
-`RS_Count(raster: Raster, band: Integer = 1, excludeNoDataValue: Boolean = true)`
+`RS_Count(raster: Raster)`
 
 `RS_Count(raster: Raster, band: Integer = 1)`
 
-`RS_Count(raster: Raster)`
+`RS_Count(raster: Raster, band: Integer = 1, excludeNoDataValue: Boolean = true)`
 
 Return type: `Long`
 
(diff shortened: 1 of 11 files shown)


jiayuasu (approved):
Thank you!

PR #2995, opened 2026-05-24
Fix docs broken link report

thuantip:
## Summary

- Replace root-relative `/sedonadb` homepage links with the canonical SedonaDB URL.
- Exclude the historical DataSysLab URL from Lychee checks because it currently returns 522.
- Convert the Chinese `python/pyproject.toml` bare URL into a Markdown link so the trailing full-width parenthesis is not captured.

Fixes #2963.

## Verification

- git diff --check


jiayuasu (approved):
Thank you!

PR #3018, opened 2026-05-31
[GH-3017] Implement rotate

### LINKED ISSUE #3017: Geopandas: Implement `rotate`

oglego:

- Yes, and the PR name follows the format `[GH-3017] Implement rotate`. Closes #3017 

Implement rotate

## How was this patch tested?

Added tests, verified success of all tests with:

```python
pytest -v tests/geopandas/test_geoseries.py
```

117 passed, 44108 warnings in 140.69s (0:02:20)

Verified formatting with:

```bash
make check
```

### DIFF
--- a/python/sedona/spark/geopandas/base.py
+++ b/python/sedona/spark/geopandas/base.py
@@ -1362,6 +1362,56 @@ class GeoFrame(metaclass=ABCMeta):
     # def transform(self, transformation, include_z=False):
     #     raise NotImplementedError("This method is not implemented yet.")
 
+    def rotate(self, angle, origin="center", use_radians=False):
+        """Return a ``GeoSeries`` with rotated geometries.
+
+        See http://shapely.readthedocs.io/en/latest/manual.html#shapely.affinity.rotate
+        for details.
+
+        Parameters
+        ----------
+        angle : float
+            The angle of rotation can be specified in either degrees (default)
+            or radians by setting use_radians=True. Positive angles are
+            counter-clockwise and negative are clockwise rotations.
+        origin : string, Point, or tuple (x, y)
+            The point of origin can be a keyword 'center' for the bounding box
+            center (default), 'centroid' for the geometry's centroid, a Point
+            object or a coordinate tuple (x, y).
+        use_radians : boolean
+            Whether to interpret the angle of rotation as degrees or radians
+
+        Examples
+        --------
+        >>> from shapely.geometry import Point, LineString, Polygon
+        >>> s = geopandas.GeoSeries(
+        ...     [
+        ...         Point(1, 1),
+        ...         LineString([(1, -1), (1, 0)]),
+        ...         Polygon([(3, -1), (4, 0), (3, 1)]),
+        ...     ]
+        ... )
+        >>> s
+        0                         POINT (1 1)
+        1              LINESTRING (1 -1, 1 0)
+        2    POLYGON ((3 -1, 4 0, 3 1, 3 -1))
+        dtype: geometry
+
+        >>> s.rotate(90)
+        0                                          POINT (1 1)
+        1                      LINESTRING (1.5 -0.5, 0.5 -0.5)
+        2    POLYGON ((4.5 -0.5, 3.5 0.5, 2.5 -0.5, 4.5 -0.5))
+        dtype: geometry
+
+        >>> s.rotate(90, origin=(0, 0))
+        0                       POINT (-1 1)
+        1              LINESTRING (1 1, 0 1)
+        2    POLYGON ((1 3, 0 4, -1 3, 1 3))
+        dtype: geometry
+
+        """
+        return _delegate_to_geometry_column("rotate", self, angle, origin, use_radians)
+
     def force_2d(self):
         """Force the dimensionality of a geometry to 2D.
 
--- a/python/sedona/spark/geopandas/base.py
+++ b/python/sedona/spark/geopandas/base.py
@@ -1384,7 +1384,8 @@ class GeoFrame(metaclass=ABCMeta):
         Examples
         --------
         >>> from shapely.geometry import Point, LineString, Polygon
-        >>> s = geopandas.GeoSeries(
+        >>> from sedona.spark.geopandas import GeoSeries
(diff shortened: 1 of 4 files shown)


oglego:
Hey @petern48,

I have tried to follow along with the instructions on ticket 2230 and PR example 2231 but if I missed anything on this please let me know!  Thank you for putting those instructions together they were very helpful!

python/sedona/spark/geopandas/base.py
```diff
@@ -1362,6 +1362,56 @@ def segmentize(self, max_segment_length):
...
+            The point of origin can be a keyword 'center' for the bounding box
+            center (default), 'centroid' for the geometry's centroid, a Point
+            object or a coordinate tuple (x, y).
+        use_radians : boolean
+            Whether to interpret the angle of rotation as degrees or radians
+
+        Examples
+        --------
+        >>> from shapely.geometry import Point, LineString, Polygon
+        >>> s = geopandas.GeoSeries(
```
```suggestion
        >>> from sedona.spark.geopandas import GeoSeries
        >>> s = GeoSeries(
```
we should update the import for sedona instead of geopandas

python/tests/geopandas/test_match_geopandas_series.py
```diff
@@ -956,6 +956,17 @@ def test_segmentize(self):
...
 
+    def test_rotate(self):
+        for geom in self.geoms:
+            # Sedona converts empty geometries to None, skip them
+            if not gpd.GeoSeries(geom).is_empty.any():
+                sgpd_result = GeoSeries(geom).rotate(45)
+                gpd_result = gpd.GeoSeries(geom).rotate(45)
+                self.check_sgpd_equals_gpd(sgpd_result, gpd_result)
+                sgpd_result = GeoSeries(geom).rotate(45, origin=(0, 0))
+                gpd_result = gpd.GeoSeries(geom).rotate(45, origin=(0, 0))
```
petern48: can we update these to an origin other than `(0, 0)`, just to make sure it works? maybe something like `(1, 2)`.

can we also add add a case for testing `"centroid"` and a `Point()` object as input for origin field to make sure they work?

python/tests/geopandas/test_geoseries.py
```diff
@@ -1847,6 +1847,44 @@ def test_segmentize(self):
...
+            Point(1, 1),
+            LineString([(1, -1), (1, 0)]),
+            Polygon([(3, -1), (4, 0), (3, 1)]),
+            None,
+        ]
+        s = GeoSeries(geoms)
+        gpd_s = gpd.GeoSeries(geoms)
+
+        # Test default (degrees, origin='center')
+        self.check_sgpd_equals_gpd(s.rotate(90), gpd_s.rotate(90))
```
petern48: `test_geoseries.py` should be testing the hard-coded expected outputs instead of comparing it directly to the result of what geopandas returns (that's what test_match_geopandas_series.py` is for. Could you update all of these examples to compare to the hard-coded expected output geometries?

petern48 (commented):
Thanks for working on this. I've done an initial scan and left some feedback

oglego:
@petern48 thank you for reviewing this and for the feedback on it I really appreciate it!  I've gone back through the tests and have tried to update them using the suggestions you have provided - if I missed anything on them or need to be more explicit in some of the expected results just let me know.  Thanks again!

python/tests/geopandas/test_match_geopandas_series.py
```diff
@@ -956,6 +956,19 @@ def test_segmentize(self):
...
 
+    @pytest.mark.parametrize("angle", [0, 45, 90, 180])
+    @pytest.mark.parametrize(
+        "origin_kwargs",
+        [{}, {"origin": (1, 2)}, {"origin": "centroid"}, {"origin": "point"}],
+    )
+    def test_rotate(self, angle, origin_kwargs):
+        for geom in self.geoms:
+            # Sedona converts empty geometries to None, skip them
+            if not gpd.GeoSeries(geom).is_empty.any():
```
jiayuasu: This condition skips every fixture group because each list in `self.geoms` starts with at least one empty geometry, so the parametrized cases pass without running any rotate comparisons. Could we filter empty geometries within each group, or build non-empty rotate fixtures, and replace `origin="point"` with an actual `Point` origin or an explicit invalid-origin case?

python/sedona/spark/geopandas/geoseries.py
```diff
@@ -1160,6 +1160,36 @@ def transform(self, transformation, include_z=False):
...
+                spark_expr = stf.ST_Rotate(self.spark.column, angle, origin_x, origin_y)
+            elif origin == "centroid":
+                origin_x = stf.ST_X(stf.ST_Centroid(self.spark.column))
+                origin_y = stf.ST_Y(stf.ST_Centroid(self.spark.column))
+                spark_expr = stf.ST_Rotate(self.spark.column, angle, origin_x, origin_y)
+            else:
+                raise ValueError(
+                    f"origin must be 'center', 'centroid', a Point, or (x, y) tuple, got {origin!r}"
+                )
+        elif isinstance(origin, tuple) and len(origin) == 2:
```
jiayuasu: GeoPandas/Shapely also accepts a two-element list for `origin`, for example `origin=[0, 0]`, but this branch only accepts tuples. Since this wrapper is matching GeoPandas semantics, can we accept any two-element coordinate sequence here and add a list-origin regression test?

jiayuasu (commented):
Review notes on the rotate implementation and coverage.

python/tests/geopandas/test_match_geopandas_series.py
oglego: ahhh, thank you for reviewing this and catching that!  I'll get that updated and fixed!

oglego:
@jiayuasu I looked around at the other tests and noticed that `test_union_all` used a list comprehension to help with filtering so I ended up doing something similar with filtering out the empty geometries, let me know if there is a better way of doing this!  I think I've got everything updated now with your feedback, thanks again for reviewing this!

jiayuasu (approved):
LGTM. Thank you!

Issue #987, opened 2023-08-25
Shpefile attributes do not produce values in the column after joining.


PrathameshDhapodkar:
## Expected behavior
value| current_timestamp | network_operator_name | dl_load_date | isDishNrCell | isSimDish | nrNci | AOI_ID | Cluster_ID
(nestedjsonvalue) | 8/22/2023 5:44:11 PM | Digicel | 8/22/2023 | FALSE | TRUE | 3569856325 | ALB | ALB-01-Downtown

## Actual behavior
value| current_timestamp | network_operator_name | dl_load_date | isDishNrCell | isSimDish | nrNci | AOI_ID | Cluster_ID
(nestedjsonvalue) | 8/22/2023 5:44:11 PM | Digicel | 8/22/2023 | FALSE | TRUE | 3569856325 |  | 

## Steps to reproduce the problem
actual dataframe is a streaming dataset running on spark cluster.

1. Create spark session
2. get shape file from location(s3 here)
code:
  def getAoiShapeDf: DataFrame  = {
    val aoiShapefileLocation = "s3://bucket/opensource_loc/top_shp/aoi_oto/"
    val aoiShapeRdd = ShapefileReader.readToGeometryRDD(session.sparkContext, aoiShapefileLocation)
    aoiShapeRdd.CRSTransform("epsg:4326", "epsg:5070", false)
    val aoiShapeDf = Adapter.toDf(aoiShapeRdd, session)
    aoiShapeDf
  }

  def getClusterShapeDf: DataFrame = {
    val clusterShapefileLocation = "s3://bucket/opensource_loc/top_shp/cluster_oto/"
    val clusterShapeRdd = ShapefileReader.readToGeometryRDD(session.sparkContext, clusterShapefileLocation)
    clusterShapeRdd.CRSTransform("epsg:4326", "epsg:5070", false)
    val clusterShapeDf = Adapter.toDf(clusterShapeRdd, session)
    clusterShapeDf
  }

3. join shape file dataframe with actual dataframe on ST_Contains join condition.
code:
  def enrichWithAoi(dataframe:DataFrame,clientLatColumn: String, clientLongColumn: String): DataFrame = {
    val networkAoiShape = broadcast(this.getAoiShapeDf.select("geometry","AOI_ID"))
    val ueDataWithGeom = dataframe.withColumn("aoiGeoPoint",
[...]

PR #985, opened 2023-08-24
[SEDONA-375] Add RS_SetBandNoDataValue

furqaankhan:
## Did you read the Contributor Guide?

- Yes, I have read [Contributor Rules](https://sedona.apache.org/latest-snapshot/community/rule/) and [Contributor Development Guide](https://sedona.apache.org/latest-snapshot/community/develop/)

## Is this PR related to a JIRA ticket?

- Yes, the URL of the associated JIRA ticket is https://issues.apache.org/jira/browse/SEDONA-375. The PR name follows the format `[SEDONA-XXX] my subject`.

## What changes were proposed in this PR?

- Add RS_SetBandNoDataValue

## How was this patch tested?

- Passed added tests

## Did this PR include necessary documentation updates?

- Yes, I have updated the documentation update.

PR #3064, opened 2026-06-19
[DOCS] Release post for SedonaDB 0.4.0

paleolimbot:

- No:
  - this is a documentation update. The PR name follows the format `[DOCS] my subject`

Added a (draft) post for the forthcoming SedonaDB release candidate

## How was this patch tested?

Just docs


docs/blog/posts/intro-sedonadb-0-4.md
```diff
@@ -0,0 +1,398 @@
...
+Support for `.group_by()`, `.agg()`, `.distinct()`, and `.distinct_on()` were also added in 0.4.0 and more are in the works!
+
+In addition to data frame operators, we increasingly realized that our hard-won library of 170+ spatial functions was difficult to explore and use (despite improved [SQL reference documentation](https://sedona.apache.org/sedonadb/latest/reference/sql/)!). Following the pattern of [Pandas-style datatype-specific accessors](https://pandas.pydata.org/docs/reference/series.html#accessors), you can now write expressions as chains with inline documentation helping you as you go.
+
+
+```python
+countries.select(
+    countries.name, geometry=countries.geometry.geo.centroid().geo.buffer(0.1)
+).limit(4)
+```
```
paleolimbot, line 114: For this we could also have a screen cap video of what this looks like to type (where the functions and their parameters are shown as you go)

docs/blog/posts/intro-sedonadb-0-4.md
```diff
@@ -0,0 +1,398 @@
...
+
+## Packaging for conda-forge
+
+We're excited to announce that sedonadb is now available on conda-forge! Users of the conda ecosystem can now install SedonaDB with:
+
+```shell
+conda install -c conda-forge sedonadb
+```
+
+Thank you to [p-vdp](https://github.com/p-vdp) for driving this work!
```
paleolimbot, line 52: @p-vdp Did I get this right / is there anything else that should go in this section?

docs/blog/posts/intro-sedonadb-0-4.md
```diff
@@ -0,0 +1,398 @@
...
+
+Thanks to [edzer](https://github.com/edzer), [benbovy](https://github.com/benbovy), [jorisvandenbossche](https://github.com/jorisvandenbossche), and many years of [s2geography](https://github.com/paleolimbot/s2geography) and [r-spatial/s2](https://github.com/r-spatial/s2) contributors that formed the basis for these implementations.
+
+## GPU-Accelerated Spatial Join
+
+```bash
+docker run -it --rm --gpus all -p 8888:8888 apache/sedona:sedonadb-latest
+```
```
paleolimbot, line 299: @pwrliang @jiayuasu Would either of you like to fill in this section? I can give it a try based on the docs but I think you both have a better handle on one good motivating example for this feature.

docs/blog/posts/intro-sedonadb-0-4.md
```diff
@@ -0,0 +1,398 @@
...
+
+
+
+
+    <pyarrow._parquet.ParquetSchema object at 0x127948e00>
+    required group field_id=-1 arrow_schema {
+      optional binary field_id=-1 name (String);
+      optional binary field_id=-1 continent (String);
+      optional binary field_id=-1 geometry;
+    }
```
paleolimbot, line 351: This has a bug that we need to fix...it should be writing a Geography type here ( https://github.com/apache/sedona-db/issues/983 )

docs/blog/posts/intro-sedonadb-0-4.md
```diff
@@ -0,0 +1,398 @@
...
+
+Thank you to [Kontinuation](https://github.com/Kontinuation), [james-willis](https://github.com/james-willis), [oglego](https://github.com/oglego), and [sapienza88](https://github.com/sapienza88) for these contributions!
+
+## Raster Infrastructure
+
+While we're not ready to announce that SedonaDB supports raster data, SedonaDB contributors dedicated significant time laying the foundation for first-class raster and ND-array data support, drawing the best from [Sedona Spark's Raster SQL](https://sedona.apache.org/latest/api/sql/Raster-Functions/), [PostGIS Raster Support](https://postgis.net/docs/RT_reference.html), [GDAL](https://gdal.org/), and [Zarr](https://zarr.dev/) with vectorized execution and SedonaDB's ground-up spatial support. We look forward to building this feature in earnest with the community over the next few months!
+
+Thank you to [Kontinuation](https://github.com/Kontinuation) and [james-willis](https://github.com/james-willis) for designing and driving this functionality!
```
paleolimbot: @james-willis @Kontinuation @jiayuasu I made a passing effort at this paragraph but I'm happy to put whatever here if any of you have suggestions!

james-willis: Revised draft — now reads a **real, public ERA5 rainfall Zarr pyramid** (EPSG:3857, anonymous; the cube is chunked by `year` *and* spatially), uses the `.rst` DataFrame accessor, and shows real output. Verified end-to-end against current `main`.

### N-Dimensional Rasters and Zarr

Geospatial raster data is increasingly a *datacube*: climate reanalyses, satellite time series, and model outputs all stack extra axes — `time`, `year`, `band` — on top of the spatial grid. In 0.4.0, SedonaDB's raster type goes natively **N-dimensional**, and the new `sedonadb-zarr` extension reads [Zarr](https://zarr.dev/) groups (v2 or v3) — local or in cloud object storage — straight into a queryable raster column.

Point SedonaDB at a Zarr datacube and explore its shape without reading a single pixel:

```python
import sedona.db
import sedonadb_zarr

sd = sedona.db.connect()
sd.register(sedonadb_zarr.ZarrExtension())

# A public ERA5 rainfall pyramid (Zarr, anonymous). Reading + inspecting
# dimensions is a metadata-only round-trip — no pixel bytes fetched.
url = "https://weathermapdata.rdrn.me/era5_2015_2020_l5.zarr/2"
spec = sedonadb_zarr.Zarr().with_options({"arrays": ["rain_ok"]})
cube = sd.read(url, format=spec)

cube.select(
    cube.raster.rst.num_dimensions().alias("ndim"),
    cube.raster.rst.dim_names().alias("dims"),
    cube.raster.rst.shape().alias("shape"),
    cube.raster.rst.srid().alias("srid"),
).show(1)
```

```text
┌───────┬──────────────┬───────────────┬────────┐
│  ndim ┆     dims     ┆     shape     ┆  srid  │
│ int32 ┆     list     ┆      list     ┆ uint32 │
╞═══════╪══════════════╪═══════════════╪════════╡
│     3 ┆ [year, y, x] ┆ [1, 128, 128] ┆   3857 │
└───────┴──────────────┴───────────────┴────────┘
```

`sedonadb-zarr` emits **one row per Zarr chunk**, so the storage layout *is* the data layout. This cube tiles each year into a `4×4` grid and chunks one year per chunk, so it loads as `16 × 6 = 96` rows — `cube.count()` confirms — each a single year of one spatial tile. Inspecting dimensions touches only the group schema: no pixel bytes.

`RS_Slice` collapses a named axis. Slicing the `year` dimension hands back a 2-D `[y, x]` rainfall field:

```python
sliced = cube.select(plane=cube.raster.rst.slice("year", 0))
sliced.select(
    dims=sliced.plane.rst.dim_names(),
    shape=sliced.plane.rst.shape(),
).show(1)
```

```text
┌────────┬────────────┐
│  dims  ┆    shape   │
╞════════╪════════════╡
│ [y, x] ┆ [128, 128] │
└────────┴────────────┘
```

`RS_Slice` needs pixels, so SedonaDB resolves each row's Zarr chunk on demand — you never call a loader yourself.

> *Note:* pixel-reading operations like `RS_Slice` fetch a chunk's bytes on demand, and do so **eagerly** when the operator runs. We're separately making slice and other "crop" operators *lazy* — a lightweight view over the chunk, so bytes aren't retrieved until their values are consumed ([#813](https://github.com/apache/sedona-db/pull/813)).

And because each chunk is a georeferenced row, you can see a Zarr's layout on a map without decoding a pixel. `RS_Envelope` turns each chunk into its footprint; reproject to lon/lat and the `4×4` chunk grid draws straight onto a map:

```python
from lonboard import viz   # in a notebook

f = sd.funcs
chunks = cube.select(geom=f.st_transform(cube.raster.rst.envelope(), "EPSG:4326"))

# Draw outlines only, so the basemap shows through the chunk grid.
viz(
    chunks,
    polygon_kwargs=dict(
        filled=False, stroked=True,
        get_line_color=[236, 64, 160], line_width_min_pixels=2,
    ),
)
```

*(Figure to attach: the chunk grid drawn as a `4×4` lattice over a world basemap.)*

For the full walkthrough — load a cube, inspect its dimensions, slice a plane, map the chunks, and hand a plane to NumPy — see [Working with Zarr and NDArray data in SedonaDB](LINK).

We're excited about what shipped here — and we're just getting started. There's more user-facing functionality for N-dimensional rasters and Zarr on the way, and we'd love your input on where it goes next. If you're working with datacubes or cloud-native raster data, [open an issue](https://github.com/apache/sedona-db/issues/new/choose) and tell us what you're building and what you need.

docs/blog/posts/intro-sedonadb-0-4.md
pwrliang: Hi, this is my release notes:
---
SedonaDB now introduces hardware acceleration via an integrated **GPU Spatial Join Library**. This feature significantly boosts the performance of compute-intensive spatial joins by offloading highly parallel filtering and refinement operations to the GPU.

### Key Capabilities & Enhancements

* **Ray Tracing (RT) Core Acceleration:** Repurposes dedicated GPU RT cores to accelerate the bounding-box filtering stage of spatial queries and Point-in-Polygon (PIP) tests. This delivers massive performance gains on complex spatial joins (e.g., `intersects`, `contains`). The evaluation of PIP queries is heavily optimized to exploit RT cores, while other geometric operations run on CUDA cores.
* **GPU-Optimized Storage Layout:** Unlike conventional GPU databases that load entire datasets into device memory, SedonaDB only loads geometries in Well-Known Binary (WKB) format to the GPU during query execution. This allows large queries to run efficiently even with limited device memory. The WKB data is subsequently converted into a GPU-friendly format, maximizing memory throughput and enabling parallel random access directly on the device.
* **CPU Fallback:** Currently, only a subset of spatial predicates are [supported](https://github.com/apache/sedona-db/blob/main/docs/gpu-acceleration.md). When executing an unsupported spatial join, the engine automatically falls back to the CPU implementation.

---

### Prerequisites & Deployment

By default, the GPU feature is disabled and is not included in the standard published python packages.

**Hardware Requirements:**

* An NVIDIA GPU with a compute capability of $\ge$ 7.5. A GPU without RT cores (e.g., A100, H100) should also work.

#### Quick Start with Docker

We provide an official Docker image to easily try this feature with a single command:

```bash
docker run -it --rm --gpus all -p 8888:8888 apache/sedona:sedonadb-latest

```

> ⚠️ **Note:** This pre-built image supports GPU models with compute capabilities **7.5, 8.6, and 8.9**.

For other GPU models, we encourage users to build the image from source to avoid time-consuming Just-In-Time (JIT) compilation:

```bash
docker build -f docker/sedonadb-gpu.dockerfile --build-arg CMAKE_CUDA_ARCHITECTURES="<your GPU compute capability>" -t sedonadb-gpu .

```

---

### Usage

Launching the container provides a **JupyterLab** instance. From there, you can connect to SedonaDB and enable GPU acceleration using the following configuration:

```python
import sedonadb

ctx = sedonadb.connect()

# Enable the GPU feature
ctx.sql("SET gpu.enable = true") 

# Increase the batch size to feed sufficient data to the GPU
ctx.sql("SET datafusion.execution.batch_size = 100000") 

```

docs/blog/posts/intro-sedonadb-0-4.md
paleolimbot: Just kidding, this was user error (was outputting GeoParquet 1.0 with plain types)

docs/blog/posts/intro-sedonadb-0-4.md
james-willis: @paleolimbot — revised this draft above. It now reads a real public ERA5 rainfall Zarr pyramid (EPSG:3857) instead of the placeholder cube, uses the `.rst` DataFrame accessor, and includes real output. The two pieces it needed both landed in sedona-db — CF/rioxarray CRS (#985) and the zlib codec (#987) — so it runs end-to-end against current `main`.

paleolimbot: I edited this a bit for length/clarity but feel free to suggest changes!

docs/blog/posts/intro-sedonadb-0-4.md
paleolimbot: I edited this a bit for length/clarity but feel free to suggest changes!

jiayuasu (approved):
Please fix the pre commit

PR #3108, opened 2026-07-15
[GH-3166] Resolve raster sample positions the way Java AWT does in the Python reader

### LINKED ISSUE #3166: Python raster reader does not resolve sample positions the way Java AWT does
## Expected behavior

`SedonaRaster.as_numpy()` should return the samples that `java.awt.image` reads from the same sample model and data buffer. The Python sample models in `python/sedona/spark/raster/sample_model.py` are ports of the Java AWT ones, so a sample of band `b` at `(x, y)` should resolve exactly the way `Raster.getSample(x, y, b)` does:

```java
// ComponentSampleModel
data.getElem(bankIndices[b], y * scanlineStride + x * pixelStride + bandOffsets[b])
// DataBufferInt and friends
public int getElem(int bank, int i) { return bankdata[bank][i + offsets[bank]]; }
```

## Actual behavior

The Python implementation disagrees with Java AWT in five ways. Rasters affected either come back with silently wrong pixel values or fail to be read at all.

**1. `DataBuffer.offsets` is ignored by all four sample models.** Nothing in `sample_model.py` reads `data_buffer.offsets`; the models index `data_buffer.bank_data[i]` directly. Java resolves a sample as `bankdata[bank][i + offsets[bank]]`. Every sample of a bank with a non-zero offset is therefore read from the wrong position. `DataBufferSerializer` writes these offsets and `raster_serde._read_data_buffer` deserializes them, so they do arrive on the Python side and are then dropped.

**2. The fast paths reshape a whole bank instead of a `width * height` window.** `ComponentSampleModel.as_numpy` narrows the bank only when the band offset is non-zero, and `PixelInterleavedSampleModel.as_numpy` never narrows it. A bank that holds more samples than the image needs — which is exactly what a non-zero buffer offset implies, but also happens with a zero band offset over a padded bank — raises:

```
ValueError: cannot reshape array of size 6 into shape (2,2)
```

**3. `ComponentSampleModel`'s fast path indexes band offsets by bank number.** It uses `self.band_offsets[bank_index]` where the band position `k` is meant, so a raster whose `bank_indices` is not the identity mapping reads bands at each other's offsets. The slow path already uses `band_offsets[k]`.

**4. `PixelInterleavedSampleModel`'s slow path assumes band offsets fall inside one pixel stride.** It slices a single pixel out of the bank and then indexes it with the band offsets:

```python
pixel = bank_data[begin:end][self.band_offsets]   # end = begin + num_bands
```

Band offsets are positions within a scanline, not within one pixel, so a layout such as `pixelStride = 4` with `bandOffsets = {3, 1, 0}` raises:

```
IndexError: index 3 is out of bounds for axis 0 with size 3
```

**5. `SinglePixelPackedSampleModel` masks and shifts signed samples.** Java extracts a band with `(bits & bitMasks[b]) >>> bitOffsets[b]`, an unsigned shift. The Python code applies `>>` to the signed sample, so a mask that covers the sign bit — the alpha mask `0xFF000000` of a standard ARGB raster — yields sign-extended negative values instead of `0..255`.

**6. `MultiPixelPackedSampleModel` reads one sample past the end of every scanline.** The `pos += 1; value = bank_data[pos]` at the bottom of its loop runs even when the scanline is already complete, so a bank that ends where the last scanline ends raises:

```
IndexError: index 1 is out of bounds for axis 0 with size 1
```

Its running bit shift also diverges from Java when `data_bit_offset` is not a multiple of `num_bits`, the case where a pixel spans two data elements. For every `data_bit_offset` that is a multiple of `num_bits`, the loop agrees with Java (checked over 1323 parameter combinations).

**7. `MultiPixelPackedSampleModel` shifts samples as unsigned values, by the distance the sample model implies.** Java reads a sample through `DataBuffer.getElem()`, which returns a signed `int`, and shifts it with `>>`: an arithmetic shift whose distance an `int` shift takes modulo 32. Two things follow from that.

A `data_bit_offset` that is not a multiple of `num_bits` leaves a pixel straddling two samples, which makes the distance negative. Java masks it into a large positive one and shifts the sample's top bits down; numpy shifts an unsigned array by a negative count to zero:

```java
new MultiPixelPackedSampleModel(DataBuffer.TYPE_INT, 4, 1, 8, 1, 4)
// bank {0x89ABCDEF, 0x12345678} -> [154, 188, 222, 248]
// the fourth pixel's distance is 32 - 28 - 8 = -4, which java masks to 28
// python reads [154, 188, 222, 0]
```

This is not limited to samples with the sign bit set: the same layout over `{0x12345678, 0}` reads `[35, 69, 103, 1]` in Java and `[35, 69, 103, 0]` in Python.

And the mask, which Java derives as `(1 << numberOfBits) - 1` on an `int`, is **zero** for a pixel occupying a whole 32 bit sample, so Java accepts the layout and reads every pixel of such a raster as zero:

```java
new MultiPixelPackedSampleModel(DataBuffer.TYPE_INT, 3, 2, 32, 4, 0)  // accepted
// getSample over a populated bank -> [[0, 0, 0], [0, 0, 0]]
```

Python computes `(1 << 32) - 1 == 0xFFFFFFFF` and returns the backing words instead.

**8. `SinglePixelPackedSampleModel` derives a bit offset of `-1` for a zero bit mask.** Java accepts a zero mask, leaves its bit offset at 0 and reads the band as zero — `getBitOffsets()` returns `[16, 8, 0, 0]` for `{0xFF0000, 0xFF00, 0xFF, 0}`. `(v & -v).bit_length() - 1` gives `-1` for `v = 0`, and shifting by it reads the band as zero by accident on numpy 1.x but raises on numpy 2.x:

```
OverflowError: Python integer -1 out of bounds for uint32
```

None of these are covered today: every layout `RS_MakeRasterForTesting` builds uses zero buffer offsets, and its band offsets are either an identity mapping or within the pixel stride, so `python/tests/raster/test_serde.py` passes throughout. Defect 8 is also numpy-version dependent, so it would stay hidden while CI pins `numpy<2`.

The layouts are easier to build directly than to produce through the serializer, and the sample models are the unit under test either way:

```python
import numpy as np
from sedona.spark.raster.data_buffer import DataBuffer
from sedona.spark.raster.sample_model import (
    ComponentSampleModel,
    MultiPixelPackedSampleModel,
    PixelInterleavedSampleModel,
    SinglePixelPackedSampleModel,
)

# 1 + 2: BandedSampleModel(TYPE_INT, 2, 2, 2, {0, 1}, {0, 0})
#        over DataBufferInt(banks of 6 samples, 4, {2, 2})
sm = ComponentSampleModel(DataBuffer.TYPE_INT, 2, 2, 1, 2, [0, 1], [0, 0])
db = DataBuffer(
    DataBuffer.TYPE_INT,
    [np.arange(10, 16, dtype=np.int32), np.arange(20, 26, dtype=np.int32)],
    4,
    [2, 2],
)
sm.as_numpy(db)
# ValueError: cannot reshape array of size 6 into shape (2,2)
# java.awt.image reads [[[12, 13], [14, 15]], [[22, 23], [24, 25]]]

# 4: PixelInterleavedSampleModel(TYPE_INT, 2, 2, 4, 8, {3, 1, 0})
sm = PixelInterleavedSampleModel(DataBuffer.TYPE_INT, 2, 2, 4, 8, [3, 1, 0])
db = DataBuffer(DataBuffer.TYPE_INT, [np.arange(10, 26, dtype=np.int32)], 16, [0])
sm.as_numpy(db)
# IndexError: index 3 is out of bounds for axis 0 with size 3
# java.awt.image reads
# [[[13, 17], [21, 25]], [[11, 15], [19, 23]], [[10, 14], [18, 22]]]

# 5: SinglePixelPackedSampleModel(TYPE_INT, 2, 1, 2, {0xFF000000, 0xFF0000, 0xFF00, 0xFF})
sm = SinglePixelPackedSampleModel(
    DataBuffer.TYPE_INT, 2, 1, 2, [0xFF000000 - (1 << 32), 0xFF0000, 0xFF00, 0xFF]
)
db = DataBuffer(
    DataBuffer.TYPE_INT,
    [np.array([0x8090A0B0, 0xFFFFFFFF], dtype=np.uint32).astype(np.int32)],
    2,
    [0],
)
sm.as_numpy(db).tolist()
# [[[-128, -1]], [[144, 255]], [[160, 255]], [[176, 255]]]  <- alpha sign extended
# java.awt.image reads [[[128, 255]], [[144, 255]], [[160, 255]], [[176, 255]]]

# 6: MultiPixelPackedSampleModel(TYPE_BYTE, 8, 1, 1, 1, 0) over a bank of one byte
sm = MultiPixelPackedSampleModel(DataBuffer.TYPE_BYTE, 8, 1, 1, 1, 0)
db = DataBuffer(DataBuffer.TYPE_BYTE, [np.array([0xA9], dtype=np.uint8)], 1, [0])
sm.as_numpy(db)
# IndexError: index 1 is out of bounds for axis 0 with size 1
# java.awt.image reads [[[1, 0, 1, 0, 1, 0, 0, 1]]]

# 7: MultiPixelPackedSampleModel(TYPE_INT, 3, 2, 32, 4, 0)
sm = MultiPixelPackedSampleModel(DataBuffer.TYPE_INT, 3, 2, 32, 4, 0)
db = DataBuffer(
    DataBuffer.TYPE_INT,
    [np.array([0x11223344, 0x55667788, 0x99AABBCC, 0, 0xDEADBEEF, 0x0F0F0F0F,
               0x12345678, 0], dtype=np.uint32).astype(np.int32)],
    8,
    [0],
)
sm.as_numpy(db).tolist()
# [[[287454020, 1432778632, -1716864052], [-559038737, 252645135, 305419896]]]
# java.awt.image reads [[[0, 0, 0], [0, 0, 0]]]

# 8: SinglePixelPackedSampleModel(TYPE_INT, 2, 1, 2, {0xFF0000, 0xFF00, 0xFF, 0})
sm = SinglePixelPackedSampleModel(
    DataBuffer.TYPE_INT, 2, 1, 2, [0x00FF0000, 0x0000FF00, 0x000000FF, 0]
)
sm.bit_offsets  # [16, 8, 0, -1], java reports [16, 8, 0, 0]
db = DataBuffer(
    DataBuffer.TYPE_INT,
    [np.array([0x8090A0B0, 0xFFFFFFFF], dtype=np.uint32).astype(np.int32)], 2, [0],
)
sm.as_numpy(db)
# numpy 2.x: OverflowError: Python integer -1 out of bounds for uint32
# java.awt.image reads [[[144, 255]], [[160, 255]], [[176, 255]], [[0, 0]]]
```

Every "java.awt.image reads" line above was produced by handing the same sample model and data buffer to `java.awt.image` and reading the samples back with `Raster.getSample`.

## Settings

Sedona version = 1.9.0 and master

Apache Spark version = 3.5.x (not specific to a Spark version)

Apache Flink version = n/a

API type = Python

Scala version = 2.12

JRE version = 1.17

Python version = 3.12

Environment = Standalone

ShiroKSH:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #3166

The Python sample models did not resolve sample positions the way `java.awt.image` does. The band offset mix-up in `ComponentSampleModel`'s fast path was one of several defects of that kind, all of which are fixed here:

1. **Band offsets were indexed by bank instead of by band** in `ComponentSampleModel`'s fast path, so a raster whose bank mapping is not the identity returned samples from the wrong positions. The slow path already used the band position.
2. **The offsets of the `DataBuffer` were never applied**, by any of the four sample models. `DataBuffer.getElem(bank, i)` resolves to `bankdata[bank][i + offsets[bank]]` in Java, while the Python code read `bankdata[bank][i]`, so every sample of a bank with a non-zero offset came from the wrong position. `DataBuffer.bank_samples()` now applies the offset in one place, and the sample models index the array it returns.
3. **The two fast paths reshaped a whole bank** instead of a bounded `width * height` window. A bank holding more samples than the image needs, which is what a non-zero buffer offset implies, failed with `cannot reshape array of size N into shape (h,w)`.
4. **`PixelInterleavedSampleModel`'s slow path looked band offsets up within a single pixel stride** (`bank_data[begin:begin + num_bands][band_offsets]`). Band offsets are positions within a scanline and may reach past the pixel they belong to, which raised `IndexError` on such layouts.
5. **`SinglePixelPackedSampleModel` masked and shifted signed samples.** Java extracts bands with `>>>`, so a mask covering the sign bit — the alpha mask of an ARGB raster, `0xFF000000` — produced sign-extended, negative band values instead of `0..255`.
6. **`MultiPixelPackedSampleModel` always read one sample past the end of a scanline**, which is out of bounds when the last scanline ends at the end of the bank. Its running bit shift also diverged from Java for a `data_bit_offset` that is not a multiple of `num_bits`, the case where a pixel spans two data elements; for offsets that are a multiple of `num_bits` the old loop and this one agree, checked over 1323 parameter combinations.

7. **`MultiPixelPackedSampleModel` shifted samples as unsigned values, by the distance the sample model implies.** Java shifts the sample as the signed `int` that `DataBuffer.getElem()` returns, using `>>`, and takes the distance of an `int` shift modulo 32. Two consequences were missed: a `data_bit_offset` that is not a multiple of `num_bits` leaves a pixel straddling two samples and makes that distance negative, where Java shifts the sample's top bits down (`MultiPixelPackedSampleModel(TYPE_INT, 4, 1, 8, 1, 4)` over `{0x89ABCDEF, 0x12345678}` reads `[154, 188, 222, 248]`, and the last pixel came out as `0`); and the mask, derived as `(1 << numberOfBits) - 1` on an `int`, is *zero* for a pixel occupying a whole 32 bit sample, so Java reads every pixel of such a raster as zero while the port returned the backing words.
8. **`SinglePixelPackedSampleModel` derived a bit offset of `-1` for a zero bit mask,** which Java accepts and reads as a zero band. Shifting by `-1` read the band as zero by accident on numpy 1.x and raised `OverflowError: Python integer -1 out of bounds for uint32` on numpy 2.x.

Both packed models now resolve each pixel on its own, the way Java does, and the strided paths index with numpy instead of looping over every pixel in Python.

The two layouts behind defect 7 are quirks of Java's `int` arithmetic rather than sample addressing, and neither can hold a pixel that survives a round trip through Java itself — `setSample` is a no-op for a 32 bit pixel, and it truncates a straddling pixel to the bits that fit. Reading them any other way would put `as_numpy()` at odds with `RS_Value`, `RS_AsGeoTiff` and everything else that reads through `Raster.getSamples`, without recovering the value the caller wrote, so they are reproduced as Java reads them and a `UserWarning` names the layout when one is encountered.

## How was this patch tested?

New unit tests in `python/tests/raster/test_sample_model.py`: 22 cases across all four sample models. Every expected value was produced by handing the same sample model and data buffer to `java.awt.image` and reading the samples back with `Raster.getSample`, so the tests record what Java AWT reads rather than what the implementation happens to produce. 21 of the 22 fail without this change.

The warning is covered too: it fires for both quirk layouts, and reading the packed layout `RS_MakeRasterForTesting` builds today raises nothing under `warnings.simplefilter("error")`.

Both packed models were additionally swept over **7580 layouts** recorded from `java.awt.image`: every `num_bits` that divides the sample size for byte, ushort and int samples, **every `data_bit_offset` from 0 to two samples wide** so that pixels straddling two samples are covered, zero and non-zero bank offsets, and mask sets covering zero masks, whole-sample masks and masks over the sign bit. All 7580 match, under numpy 1.26 and 2.5 with numpy warnings raised as errors. 5295 of them do not match on master.

The layouts `RS_MakeRasterForTesting` builds today (`BandedSampleModel`, `PixelInterleavedSampleModel`, `PixelInterleavedSampleModelComplex`, `ComponentSampleModel`, `SinglePixelPackedSampleModel`, `MultiPixelPackedSampleModel`) were replayed through the new code with their Java-produced bank data and read back unchanged. All of them use zero buffer offsets, an identity bank mapping or in-range band offsets, which is why `tests/raster/test_serde.py` did not catch any of these defects.

Checked under numpy 1.26 and numpy 2.5.

### DIFF
--- a/python/sedona/spark/raster/sample_model.py
+++ b/python/sedona/spark/raster/sample_model.py
@@ -80,9 +80,9 @@ class ComponentSampleModel(SampleModel):
         if self.scanline_stride == self.width and self.pixel_stride == 1:
             # Fast path: no gaps between pixels
             band_arrs = []
-            for bank_index in self.bank_indices:
+            for k, bank_index in enumerate(self.bank_indices):
                 bank_data = data_buffer.bank_data[bank_index]
-                offset = self.band_offsets[bank_index]
+                offset = self.band_offsets[k]
                 if offset != 0:
                     bank_data = bank_data[offset : (offset + self.width * self.height)]
                 band_arr = bank_data.reshape(self.height, self.width)
--- a/python/sedona/spark/raster/sample_model.py
+++ b/python/sedona/spark/raster/sample_model.py
@@ -17,6 +17,7 @@
 
 from abc import ABC, abstractmethod
 from typing import List
+from warnings import warn
 
 import numpy as np
 
@@ -194,8 +195,7 @@ class MultiPixelPackedSampleModel(SampleModel):
         samples = data_buffer.bank_samples()
         bits_per_value = samples.dtype.itemsize * 8
 
-        # Resolve every pixel on its own, the way Java does: this sample model requires
-        # num_bits to divide the size of a data element, so no pixel spans two elements.
+        # Resolve every pixel on its own, the way Java does
         pixel_bits = self.data_bit_offset + np.arange(self.width) * self.num_bits
         cols = pixel_bits // bits_per_value
         shifts = bits_per_value - (pixel_bits % bits_per_value) - self.num_bits
@@ -210,10 +210,27 @@ class MultiPixelPackedSampleModel(SampleModel):
         # multiple of num_bits, which leaves a pixel straddling two samples and gives a
         # negative distance here, shifts the top bits of the sample down rather than
         # shifting the whole sample out, and a pixel occupying a whole 32 bit sample gets a
-        # zero mask and reads as zero.
-        values = samples[positions].astype(np.int32)
-        shifts = (shifts & 31).astype(np.int32)
+        # zero mask and reads as zero. Neither layout can hold a pixel that survives a round
+        # trip through Java, so warn about them rather than read them some other way.
         bit_mask = np.int32((1 << (self.num_bits % 32)) - 1)
-        pixels = (values >> shifts) & bit_mask
+        if bit_mask == 0:
+            warn(
+                "This raster packs one pixel per 32 bit sample. java.awt.image derives the "
+                "bit mask for it as `(1 << 32) - 1`, which is zero on an int, so Java reads "
+                "every pixel of such a raster as zero and writes to it are no-ops. "
+                "Returning zeroes to match."
+            )
+        elif (shifts < 0).any():
+            warn(
+                f"This raster's data bit offset ({self.data_bit_offset}) is not a multiple "
+                f"of its {self.num_bits} bits per pixel, so some pixels straddle two "
+                "samples. java.awt.image shifts those by a negative distance, which it "
+                "takes modulo 32, reading the top bits of the sample instead; Java's own "
+                "writes to those pixels are lossy in the same way. Returning what Java "
+                "reads."
+            )
+
+        values = samples[positions].astype(np.int32)
+        pixels = (values >> (shifts & 31).astype(np.int32)) & bit_mask
 
         return pixels.astype(samples.dtype).reshape(1, self.height, self.width)
--- a/python/sedona/spark/raster/sample_model.py
+++ b/python/sedona/spark/raster/sample_model.py
@@ -40,6 +40,7 @@ class SampleModel(ABC):
     data_type: int
     width: int
     height: int
+    scanline_stride: int
 
     def __init__(self, sample_model_type, data_type, width, height):
         self.sample_model_type = sample_model_type
@@ -52,6 +53,16 @@ class SampleModel(ABC):
         raise NotImplementedError(
             "Abstract method as_numpy was not implemented by subclass"
         )
+
+    def _pixel_positions(self, pixel_stride: int, offset: int = 0) -> np.ndarray:
+        """Sample position of every pixel, as a (height, width) array of indices.
+
+        The positions are relative to the start of a bank, so they index the arrays
+        returned by :meth:`DataBuffer.bank_samples`.
+        """
+        rows = np.arange(self.height) * self.scanline_stride + offset
+        cols = np.arange(self.width) * pixel_stride
+        return rows[:, np.newaxis] + cols[np.newaxis, :]
 
 
 class ComponentSampleModel(SampleModel):
@@ -77,33 +88,27 @@ class ComponentSampleModel(SampleModel):
         self.band_offsets = band_offsets
 
     def as_numpy(self, data_buffer: DataBuffer) -> np.ndarray:
-        if self.scanline_stride == self.width and self.pixel_stride == 1:
-            # Fast path: no gaps between pixels
-            band_arrs = []
-            for k, bank_index in enumerate(self.bank_indices):
-                bank_data = data_buffer.bank_data[bank_index]
-                offset = self.band_offsets[k]
-                if offset != 0:
-                    bank_data = bank_data[offset : (offset + self.width * self.height)]
-                band_arr = bank_data.reshape(self.height, self.width)
-                band_arrs.append(band_arr)
-            return np.array(band_arrs)
-        else:
-            # Slow path
-            band_arrs = []
-            for k in range(len(self.bank_indices)):
-                bank_index = self.bank_indices[k]
-                bank_data = data_buffer.bank_data[bank_index]
-                offset = self.band_offsets[k]
-                band_pixel_data = []
-                for y in range(self.height):
-                    for x in range(self.width):
-                        pos = offset + y * self.scanline_stride + x * self.pixel_stride
-                        band_pixel_data.append(bank_data[pos])
-                arr = np.array(band_pixel_data).reshape(self.height, self.width)
-                band_arrs.append(arr)
+        contiguous = self.scanline_stride == self.width and self.pixel_stride == 1
+        num_samples = self.width * self.height
 
-            return np.array(band_arrs)
+        band_arrs = []
+        for k, bank_index in enumerate(self.bank_indices):
+            # The band offset is relative to the offset of the bank the band lives in, and
+            # bank_indices/band_offsets are both indexed by band, not by bank.
+            samples = data_buffer.bank_samples(bank_index)
+            offset = self.band_offsets[k]
+            if contiguous:
+                # Fast path: the samples of a band follow each other. The bank may still
+                # hold more samples than the band needs, so slice a bounded window.
+                band_arr = samples[offset : (offset + num_samples)].reshape(
+                    self.height, self.width
+                )
+            else:
+                # Slow path: gaps between pixels or scanlines
+                band_arr = samples[self._pixel_positions(self.pixel_stride, offset)]
+            band_arrs.append(band_arr)
+
+        return np.array(band_arrs)
 
 
 class PixelInterleavedSampleModel(SampleModel):
@@ -121,26 +126,23 @@ class PixelInterleavedSampleModel(SampleModel):
 
     def as_numpy(self, data_buffer: DataBuffer) -> np.ndarray:
         num_bands = len(self.band_offsets)
-        bank_data = data_buffer.bank_data[0]
+        samples = data_buffer.bank_samples()
         if (
             self.pixel_stride == num_bands
             and self.scanline_stride == self.width * num_bands
             and self.band_offsets == list(range(0, num_bands))
         ):
-            # Fast path: no gapping in between band data, no band reordering
-            arr = bank_data.reshape(self.height, self.width, num_bands)
+            # Fast path: no gapping in between band data, no band reordering. The bank may
+            # still hold more samples than the image needs, so slice a bounded window.
+            num_samples = self.width * self.height * num_bands
+            arr = samples[:num_samples].reshape(self.height, self.width, num_bands)
             return np.transpose(arr, [2, 0, 1])
         else:
-            # Slow path
-            pixel_data = []
-            for y in range(self.height):
-                for x in range(self.width):
-                    begin = y * self.scanline_stride + x * self.pixel_stride
-                    end = begin + num_bands
-                    pixel = bank_data[begin:end][self.band_offsets]
-                    pixel_data.append(pixel)
-            arr = np.array(pixel_data).reshape(self.height, self.width, num_bands)
-            return np.transpose(arr, [2, 0, 1])
+            # Slow path. Band offsets are positions within a scanline, so they are not
+            # bound to the pixel they belong to and may reach past its pixel stride.
+            positions = self._pixel_positions(self.pixel_stride)
+            band_arrs = [samples[positions + offset] for offset in self.band_offsets]
+            return np.array(band_arrs)
 
 
 class SinglePixelPackedSampleModel(SampleModel):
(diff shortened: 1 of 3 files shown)


python/sedona/spark/raster/sample_model.py
```diff
@@ -80,9 +80,9 @@ def as_numpy(self, data_buffer: DataBuffer) -> np.ndarray:
         if self.scanline_stride == self.width and self.pixel_stride == 1:
             # Fast path: no gaps between pixels
             band_arrs = []
-            for bank_index in self.bank_indices:
+            for k, bank_index in enumerate(self.bank_indices):
                 bank_data = data_buffer.bank_data[bank_index]
-                offset = self.band_offsets[bank_index]
+                offset = self.band_offsets[k]
```
jiayuasu: Could we calculate the effective start here (`data_buffer.offsets[bank_index] + self.band_offsets[k]`) and always slice a `width * height` window? I may be missing an invariant, but a component model can have a zero band offset with a padded backing bank; then the current branch seems to pass the whole bank to `reshape()`. The slow path may need the buffer offset too. A zero-offset case in the new test would help confirm it.

jiayuasu (commented):
One question on the offset handling.

jiayuasu:
Confirmed against the JDK and fixed in https://github.com/apache/sedona/commit/642de290053000f1025626901fdb4644fa2bf7ec.

[P1] MultiPixelPackedSampleModel(TYPE_INT, 3, 2, 32, 4, 0) is accepted by Java and reads every pixel as zero, because bitMask = (1 << 32) - 1 is evaluated on an int whose shift count is taken modulo 32. The mask is now derived as (1 << (num_bits % 32)) - 1, which reproduces that and leaves every other width alone (8 on byte and 16 on ushort still mask normally).

[P2] Java leaves the bit offset of a zero mask at 0 and reads the band as zero; getBitOffsets() returns [16, 8, 0, 0] for {0xFF0000, 0xFF00, 0xFF, 0}. The offset is now 0 for a zero mask. Worth noting the crash was numpy-2 only: on numpy 1.26 the -1 shift happened to yield the right answer, so this would not have shown up in CI.

Rather than patch just these two, I swept both packed models over 1004 layouts recorded from java.awt.image: every num_bits that divides the sample size for byte, ushort and int samples, aligned and unaligned data bit offsets, zero and non-zero bank offsets, and mask sets covering zero masks, whole-sample masks and masks over the sign bit. All 1004 match on numpy 1.26 and 2.5, with numpy warnings raised as errors; 510 of them do not match on master. No third case of this kind is left.

One thing the sweep turned up: Java rejects non-contiguous bit masks in the SinglePixelPackedSampleModel constructor (Mask 195 must be contiguous), so those layouts cannot reach the reader and need no handling.

Two new regression tests cover P1 and P2 (test_multi_pixel_packed_whole_sample_pixels_read_zero, test_single_pixel_packed_zero_mask_reads_zero), taking their expected values from the same JDK run.

jiayuasu:
Confirmed and fixed in 9553cc6. JDK 17 returns `[154, 188, 222, 248]`, the port returned `[154, 188, 222, 0]`.

Correction to my previous comment first: the claim that the 1004-case sweep covered unaligned data bit offsets was wrong. Its offsets were `{0, numBits, 3 * numBits, size, size + numBits}`, every one of them a multiple of `num_bits`, so it never built a pixel straddling two samples. The sweep could not have caught this, and I should not have described it as covering unaligned offsets.

Root cause is narrower than the shift distance alone. Java reads a sample through `DataBuffer.getElem()`, which returns a signed `int`, then applies `>>`: an arithmetic shift whose distance an `int` shift takes modulo 32. So the -4 distance becomes 28 and the sample's top bits move down, sign extended. The port shifted an unsigned array by a negative count, which numpy zeroes. Samples are now read as `int32` and shifted by `shifts & 31`, which is Java's width and Java's masking.

It is not only about the sign bit, so I added both variants as regression cases: the same layout over `{0x12345678, 0}` reads `[35, 69, 103, 1]` in Java, and the last pixel was `0` here too. Test is `test_multi_pixel_packed_pixel_straddling_two_samples`, parameterized over both banks.

Keeping the offsets in scope rather than narrowing the claim, since the reader has to handle whatever a serialized Java raster contains. The sweep now enumerates every `data_bit_offset` from 0 to two samples wide instead of a handful:

| | layouts | mismatched vs java.awt.image |
|---|---|---|
| master | 7580 | 5295 |
| previous head 642de29 | 7580 | 1291 |
| this head 9553cc6 | 7580 | 0 |

Verified on numpy 1.26.4 and 2.5.1 with numpy warnings raised as errors. The unit tests are at 22 cases, 21 of which fail without this branch.

jiayuasu:
Added the warning in 3fce157. `UserWarning` via `warnings.warn`, matching how this package already reports its geometry speedup fallback, one message per quirk:

> This raster packs one pixel per 32 bit sample. java.awt.image derives the bit mask for it as `(1 << 32) - 1`, which is zero on an int, so Java reads every pixel of such a raster as zero and writes to it are no-ops. Returning zeroes to match.

> This raster's data bit offset (4) is not a multiple of its 8 bits per pixel, so some pixels straddle two samples. java.awt.image shifts those by a negative distance, which it takes modulo 32, reading the top bits of the sample instead; Java's own writes to those pixels are lossy in the same way. Returning what Java reads.

The trigger is the negative shift distance itself, not the alignment of the data bit offset, because whether a pixel straddles also depends on the width: 4 bit pixels at offset 2 in a 32 bit sample fit until x reaches 8. Checking the distance keeps it exact in both directions.

Behaviour, all three verified rather than assumed:

- reading the packed layout `RS_MakeRasterForTesting` builds today emits nothing, asserted in the test under `warnings.simplefilter("error")` so a future false positive fails CI
- 500 reads of one odd raster produce 1 warning, since Python's default filter dedups per message and location; distinct layouts warn separately
- values did not move: the 7580 case sweep against `java.awt.image` is still exact

Worth knowing where this does and does not help: inside a pandas UDF the warning lands on executor stderr, which is easy to miss. It is aimed at driver-side reads and notebooks, so it is a signpost for whoever is debugging an odd raster interactively rather than a guard rail on a cluster.

PR #3114, opened 2026-07-17
[GH-3112] RS_AsRaster fills uncovered pixels with noDataValue

### LINKED ISSUE #3112: RS_AsRaster: pixels not covered by the geometry get value 0 instead of noDataValue
## Expected behavior

When `RS_AsRaster` is called with a `noDataValue`, pixels not covered by the geometry should be set to that value, so the output distinguishes "burned" from "background". This is what PostGIS `ST_AsRaster` does (`nodataval` is assigned to cells not touched by the geometry), and the GDAL equivalent (`gdal_rasterize -init <nodata> -a_nodata <nodata>`, or `rasterio.features.rasterize(fill=nodata)`).

## Actual behavior

The `noDataValue` is only attached as band metadata. Background pixels keep Java's zero-initialized default, so they come out as `0` — a perfectly valid pixel value — while the band claims its nodata is something else.

A polygon with a hole makes the consequence concrete. Repro on a 7x6 grid (Sedona 1.9.0, Spark 4.0), burning 1 with nodata 9:

```sql
  ST_GeomFromWKT('POLYGON ((100.2 499.5, 113.8 499.5, 113.8 482.4, 100.2 482.4, 100.2 499.5),
                           (104.1 496.9, 110.2 496.9, 110.2 487.8, 104.1 487.8, 104.1 496.9))'),
  RS_MakeEmptyRaster(1, 'B', 7, 6, 100.0, 500.0, 2.0, -3.0, 0.0, 0.0, 0),
  'B', false, 1.0, 9.0, false), 1)
```

Sedona returns (reshaped to 6x7, `RS_BandNoDataValue` reports 9.0):

```
1 1 1 1 1 1 1
1 1 0 0 0 1 1
1 1 0 0 0 1 1
1 1 0 0 0 1 1
1 1 1 1 1 1 1
1 1 1 1 1 1 1
```

PostGIS and GDAL/rasterio return:

```
1 1 1 1 1 1 1
1 1 9 9 9 1 1
1 1 9 9 9 1 1
1 1 9 9 9 1 1
1 1 1 1 1 1 1
1 1 1 1 1 1 1
```

The pixel selection itself is identical across engines here (all rings are axis-aligned), so the divergence is purely the background fill. The hole shows why this can't be worked around by choosing a tighter extent: the uncovered region is interior to the geometry, and it reads back as valid data `0` rather than nodata.

## Root cause

`RasterConstructors.asRaster` burns the geometry into the `WritableRaster` returned by `Rasterization.rasterize` (whose backing Java array is zero-initialized), then only records the nodata as metadata via `setBandNoDataValue` — nothing ever writes the background ([RasterConstructors.java#L133-L149 @ 0299fa8](https://github.com/apache/sedona/blob/0299fa8a380a/common/src/main/java/org/apache/sedona/common/raster/RasterConstructors.java#L133-L149)):

```java
List<Object> objects =
    Rasterization.rasterize(geom, raster, pixelType, value, useGeometryExtent, allTouched);

WritableRaster writableRaster = (WritableRaster) objects.get(0);
...
if (noDataValue != null) {
  resultRaster = RasterBandEditors.setBandNoDataValue(resultRaster, 1, noDataValue);
}
```

This is entirely Sedona-side (no geotools constraint): the fix is to initialize the raster's pixels to `noDataValue` before burning whenever one is provided.

## Impact

Any downstream consumer that respects nodata (stats, map algebra, export to GeoTIFF and reading in other tools) treats the entire background as valid zeros. When 0 is a legitimate data value there is no way to tell background from data, despite the caller having explicitly declared a nodata value.

Affected versions: reproduced on 1.9.0 (released jars); the same code is present on current master (`0299fa8`).

james-willis:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #3112

`RS_AsRaster` previously only attached `noDataValue` as band metadata; pixels not covered by the geometry kept the allocation default of `0`, a valid pixel value indistinguishable from data. Now `Rasterization` fills the freshly-allocated raster with the `noDataValue` before burning the geometry, so uncovered pixels (including interior holes) read back as the declared nodata — matching PostGIS `ST_AsRaster` (`nodataval` fills untouched cells) and the `gdal_rasterize -init <nodata> -a_nodata <nodata>` idiom. The fill happens before burning (not after) because a burn `value` of `0` would otherwise be indistinguishable from background.

**Omitted `noDataValue` now inherits the reference band's nodata** (used for both the band metadata and the background fill) instead of leaving an all-zero background with no nodata metadata. If the reference band has no nodata value — or its nodata is not representable in the output pixel type — the call errors with guidance to pass an explicit representable `noDataValue` to override (`null` opts out of a nodata value, and is reachable only through the programmatic API / the widest overload).

**Representability validation.** A `noDataValue` or burn `value` that cannot be represented exactly in the output pixel type is now rejected up front instead of silently coerced: an out-of-range or fractional value on an integer band, a value that does not round-trip through 32-bit float on an `F` band, or a non-finite `NaN`/`±Infinity`. Silent coercion would store a different number than the caller set (so the background read back as data), and for some inputs previously crashed deep in GeoTools. A burn `value` equal to the resolved `noDataValue` is also rejected, since every covered pixel would then read as nodata.

Because this validation lives in the shared rasterization path, **`RS_SetValues` likewise now rejects a non-representable or non-finite `value` for the target band** rather than silently coercing it — a consistency change to that function. `RS_Clip` is unaffected (its fixed mask marker is always representable). `RS_SetValues` still rasterizes its geometry mask with an explicit `null` nodata and reads the result purely as a `0`/non-`0` mask, so its masking behavior is otherwise unchanged.

Existing test expectations that asserted `0` backgrounds alongside a non-zero `noDataValue` are updated to expect the nodata — those expectations were pinning the bug from #3112.

## How was this patch tested?

- New `testAsRasterBackgroundIsNoDataValue`: a polygon with an interior hole where the hole and the outside must read back as the nodata; a `value = 0` burn that must stay distinguishable from the nodata background; the geometry-extent output path; and the null-nodata default keeping `0`. The test was written first and fails on master.
- Additional tests cover: signed-zero (`-0.0`) fill; representability rejection of out-of-range/fractional integer, non-exact-float, and non-finite values; the inherited-nodata error (no nodata to inherit, and inherited value not representable in the output type); and the burn-`value == noDataValue` guard.
- The full `common` module raster suite and the Spark `rasteralgebraTest` / `rasterIOTest` suites pass. `RS_SetValues`, `RS_Clip`, and the zonal-stats path (which pass `null` nodata and rely on `0`-backed masks) are unaffected in their masking behavior.


common/src/main/java/org/apache/sedona/common/raster/Rasterization.java
```diff
@@ -85,6 +99,23 @@ protected static List<Object> rasterize(
...
+   * covered by the geometry read back as that value instead of the allocation default of 0. A null
+   * or zero background keeps the zero-initialized allocation as-is.
+   */
+  private static void fillBackground(WritableRaster writableRaster, Double backgroundValue) {
+    if (backgroundValue == null || backgroundValue == 0) {
+      return;
+    }
+    int width = writableRaster.getWidth();
+    double[] row = new double[width];
+    Arrays.fill(row, backgroundValue);
```
jiayuasu, line 122: Would it be worth normalizing `backgroundValue` to `pixelType` before writing it here? I may be missing validation elsewhere, but the integer sample conversion seems able to produce a pixel value that differs from the nodata metadata.

For example:

```sql
WITH r AS (
  SELECT RS_AsRaster(
    ST_GeomFromWKT('POLYGON ((0 3, 1 3, 1 2, 0 2, 0 3))'),
    RS_MakeEmptyRaster(1, 'B', 3, 3, 0.0, 3.0, 1.0, -1.0, 0.0, 0.0, 0),
    'B', false, 1.0, -1.0, false
  ) AS rast
)
SELECT RS_BandAsArray(rast, 1), RS_BandNoDataValue(rast, 1)
FROM r;
```

On this branch I get background pixels of `255`, while `RS_BandNoDataValue` returns `-1.0`. Changing `-1.0` to `300.0` gives background pixels of `44` with nodata metadata `300.0`.

That would leave the background counted as data for those inputs. PostGIS looks like it clamps/truncates nodata to the band pixel type. Could we either convert once and use the converted value for both the fill and metadata, or reject values that the pixel type cannot represent?

docs/api/sql/Raster-Operators/RS_AsRaster.md
```diff
@@ -26,7 +26,7 @@ Introduction: `RS_AsRaster` converts a vector geometry into a raster dataset by
 * `pixelType`: Defines data type of the output raster. This can be one of the following, D (double), F (float), I (integer), S (short), US (unsigned short) or B (byte).
 * `allTouched` (Since: `v1.7.1`): Decides the pixel selection criteria. If set to `true`, the function selects all pixels touched by the geometry, else, selects only pixels whose centroids intersect the geometry. Defaults to `false`.
 * `value`: The value to be used for assigning pixels covered by the geometry. Defaults to using `1.0` if not provided.
-* `noDataValue`: Used for assigning the no data value of the resultant raster. Defaults to `null` if not provided.
+* `noDataValue`: Used for assigning the no data value of the resultant raster; pixels not covered by the geometry are set to this value. Defaults to `null` if not provided, in which case uncovered pixels are `0` and the raster has no no data value.
```
jiayuasu: One separate compatibility question: do we want omission to continue meaning `null` here? I may be reading the PostGIS behavior too literally, but its scalar overloads declare `nodataval DEFAULT 0`, and omission is observably different from passing `NULL`.

This PostGIS query shows the distinction:

```sql
WITH g AS (
  SELECT ST_GeomFromText(
    'POLYGON ((0 0, 3 0, 3 3, 0 3, 0 0),
              (1 1, 2 1, 2 2, 1 2, 1 1))'
  ) AS geom
),
r AS (
  SELECT
    ST_AsRaster(geom, 1.0, -1.0, '8BUI') AS omitted,
    ST_AsRaster(
      geom, 1.0, -1.0, '8BUI', 1.0, NULL::double precision
    ) AS explicit_null
  FROM g
)
SELECT
  ST_BandNoDataValue(omitted, 1),
  ST_Value(omitted, 1, 2, 2, false),
  ST_BandNoDataValue(explicit_null, 1),
  ST_Value(explicit_null, 1, 2, 2, false)
FROM r;
```

I get `0 | 0 | NULL | 0`: both backgrounds contain zero, but only the omitted case marks zero as nodata.

The comparable Sedona check is:

```sql
WITH inputs AS (
  SELECT
    ST_GeomFromWKT(
      'POLYGON ((0 0, 3 0, 3 3, 0 3, 0 0),
                (1 1, 2 1, 2 2, 1 2, 1 1))'
    ) AS geom,
    RS_MakeEmptyRaster(
      1, 'B', 3, 3, 0.0, 3.0, 1.0, -1.0, 0.0, 0.0, 0
    ) AS ref
),
r AS (
  SELECT
    RS_AsRaster(geom, ref, 'B') AS omitted,
    RS_AsRaster(geom, ref, 'B', false, 1.0, 0.0) AS explicit_zero
  FROM inputs
)
SELECT
  RS_BandNoDataValue(omitted, 1),
  RS_BandNoDataValue(explicit_zero, 1)
FROM r;
```

That gives `NULL | 0`. I can see a compatibility reason to preserve the existing Sedona default, so this may be better handled separately. If matching PostGIS defaults is the goal, though, the shorter overloads would need to pass `0`, while explicit `NULL` could keep the current no-nodata behavior.

james-willis: I'm not sure what the project's position on matching behavior should be.

I guess my opinion (without regard to prior art) is that the default should be the NDV of `ref` and if `ref` has no NDV we should error or return null. But of course this is a breaking change.

The current behavior of setting the pixel to 0 but not setting the NDV to match is not great. But maybe it is the least bad solution.

common/src/main/java/org/apache/sedona/common/raster/Rasterization.java
james-willis: I would vote for rejecting values that the pixel type cannot represent. I do not like permissive type coercion, it is hard for users to debug.

docs/api/sql/Raster-Operators/RS_AsRaster.md
paleolimbot: I agree with James' assessment of what the ideal behaviour is, although I also don't know the degree to which a nodata value is typically lifted from a reference (are there other functions where this happens? Or does GDAL do this in its command line tool for any operations?). In general I think GDAL's command line forces you to re-specify the nodata value frequently, perhaps this is intentional (to maybe maximize a safe nodata value that won't conflict with an actual value) or perhaps this is historical.

It's possible to phase raster API changes in Sedona over a few versions...opt-in to new behaviour -> opt-out of new behaviour -> removed. This has some testing overhead during the transition but I think it will be positive for our Raster API to emerge as more consistent and easier to use (without just going through and breaking functions before we're sure).

common/src/main/java/org/apache/sedona/common/raster/Rasterization.java
jiayuasu: agree with this decision. Please fix then.

docs/api/sql/Raster-Operators/RS_AsRaster.md
jiayuasu: I’m okay changing the default behavior while the raster APIs are still evolving. When noDataValue is omitted, I’d lean toward using the reference band’s nodata value for both the output metadata and the unburned pixels. If the reference band has no nodata value, returning an error/null seems preferable to inventing one.

I also checked GDAL 3.13.1. It does not have an exact equivalent of Sedona’s reference-raster creation model. When updating an existing raster, GDAL preserves its nodata metadata and leaves pixels outside the geometry unchanged. For a new raster, nodata metadata and initialization are separate options: noData/-a_nodata sets metadata, while initValues/-init fills the pixels.

james-willis: do we consider sedona raster API to be "still evolving"? I know that is our stance in sedonadb but not sure if that is how we feel in sedona spark.

Let me know what we want for a migration story here and I will implement.

jiayuasu: Yes I still consider Sedona's raster API to be still evolving although I want to avoid unnecessary major change. But I think this change is fine. So no migration story is needed I guess?

common/src/main/java/org/apache/sedona/common/raster/Rasterization.java
james-willis: done

common/src/main/java/org/apache/sedona/common/utils/RasterUtils.java
```diff
@@ -647,6 +647,75 @@ public static boolean isDataTypeIntegral(int dataTypeCode) {
...
+        // 32-bit float represents NaN and the infinities; reject only a finite value whose
+        // magnitude overflows the float range (it would silently become an infinity). Rounding a
+        // value to the nearest float is the inherent, expected behavior of a float band.
+        if (Double.isFinite(noDataValue) && Math.abs(noDataValue) > Float.MAX_VALUE) {
+          throw new IllegalArgumentException(
+              String.format(
+                  "noDataValue %s is not representable in pixel type '%s' (32-bit float, valid range %s..%s)",
+                  noDataValue, pixelType, -Float.MAX_VALUE, Float.MAX_VALUE));
+        }
+        return noDataValue;
```
jiayuasu: Could we round-trip this through `float` before returning it? This passes validation:

`RasterConstructors.asRaster(geom, ref, "F", false, 1d, 0.1d, false)`

but then fails while constructing the nodata metadata with `Range [0.1 .. 0.1] is not valid`. Returning `(double) (float) noDataValue` makes the stored pixel and metadata agree. A test with `0.1` should catch it.

common/src/main/java/org/apache/sedona/common/raster/Rasterization.java
```diff
@@ -85,6 +99,23 @@ protected static List<Object> rasterize(
     return objects;
   }
 
+  /**
+   * Fills every pixel of the freshly-allocated raster with the background value, so pixels never
+   * covered by the geometry read back as that value instead of the allocation default of 0. A null
+   * or zero background keeps the zero-initialized allocation as-is.
+   */
+  private static void fillBackground(WritableRaster writableRaster, Double backgroundValue) {
+    if (backgroundValue == null || backgroundValue == 0) {
```
jiayuasu: `backgroundValue == 0` also matches `-0.0`. For `F`/`D`, that leaves the zero-initialized pixels as `+0.0` while the nodata metadata remains `-0.0`. `RS_Count(..., true)` uses `Double.compare`, so the hole is counted as data (9 instead of 8 in the 3x3 polygon-with-hole repro). Could we either normalize signed zero before both uses, or skip only positive zero here?

docs/api/sql/Raster-Operators/RS_AsRaster.md
james-willis: This approach matches what Postgis but does I wonder if we should just provide a different parameter for the background value that is not coupled to the no data value.

Theres no reason that the background has to be the NDV and it more cleanly solves some of the question in this thread.


Imo the only downside to this is that it deviates from postgis

jiayuasu:
@james-willis address conflict?

james-willis:
> @james-willis address conflict?

I had a comment on the thread about no data values vs unburned value and how to represent. I want to resolve on that question.

common/src/main/java/org/apache/sedona/common/utils/RasterUtils.java
james-willis: 0.1d doesn't make any sense as a no data value for a float. Floats cannot represent 0.1 so you'll never have 0.1 in the data array.

james-willis: all we could do here is make a better error message.

common/src/main/java/org/apache/sedona/common/raster/Rasterization.java
james-willis: done

docs/api/sql/Raster-Operators/RS_AsRaster.md
james-willis: @paleolimbot @jiayuasu PTAL ^

jiayuasu:
@james-willis I am fine with the PostGIS behavior which treats the NoData value and unburned value the same. And that's the behavior we have been following.

docs/api/sql/Raster-Operators/RS_AsRaster.md
james-willis: jia indicated below he does not want to separate NDV from non-burned values. 

I will confirm we have the desired behavior and merge main. 

Then this will be ready for another review

paleolimbot: > jia indicated below he does not want to separate NDV from non-burned values.

That's fine with me too...we might want to make sure there's a way to "unset" the NoData value (is there?), which would be a workaround if somebody needed this for some reason.

common/src/main/java/org/apache/sedona/common/raster/RasterConstructors.java
```diff
@@ -130,8 +131,35 @@ public static GridCoverage2D asRaster(
       boolean useGeometryExtent)
       throws FactoryException {
 
+    // Reject a burn value that cannot be represented in the target pixel type, for the same reason
+    // the noDataValue is rejected below: silently coercing an out-of-range or fractional value (for
+    // example 265 -> 9 in an unsigned 8-bit band) would store a different number than the caller
+    // asked to burn. Unlike noDataValue this is a primitive double, so it is always validated.
+    RasterUtils.assertRepresentable(value, pixelType, "value");
```
jiayuasu: I think reusing the nodata check here is too strict. `value` is the burn value, not the nodata sentinel, so normal conversion to the band type seems expected. For example, PostGIS accepts this:

```sql
WITH r AS (
  SELECT ST_AsRaster(
    ST_GeomFromText('POLYGON((0 0,2 0,2 2,0 2,0 0))'),
    2, 2, '32BF'::text, 0.1::float8, 0.5::float8
  ) AS rast
)
SELECT ST_Value(rast, 1, 1, 1, false) FROM r;
-- 0.10000000149011612
```

GDAL does the same Float32 rounding. Could we keep the exact check for `noDataValue`, but use a separate burn-value check? Rejecting out-of-range integer burns may still be useful; rejecting `0.1` for `F` seems surprising.

common/src/main/java/org/apache/sedona/common/raster/RasterConstructors.java
```diff
@@ -130,8 +131,35 @@ public static GridCoverage2D asRaster(
...
+    // background would read back as data. Validating once here keeps the background fill and the
+    // nodata metadata below using the same value.
+    if (noDataValue != null) {
+      noDataValue = RasterUtils.assertNoDataValueRepresentable(noDataValue, pixelType);
+    }
+
+    // If the burn value equals the noDataValue, every covered pixel reads back as nodata, so the
+    // geometry would vanish from the output. Reject rather than produce an all-nodata raster; this
+    // also catches the default burn value of 1 colliding with an inherited nodata of 1.0.
+    if (noDataValue != null && Double.compare(value, noDataValue) == 0) {
```
jiayuasu, line 153: I am not sure we need to reject this combination. PostGIS allows the burn value and nodata value to be equal:

```sql
WITH r AS (
  SELECT ST_AsRaster(
    ST_GeomFromText('POLYGON((0 0,2 0,2 2,0 2,0 0))'),
    2, 2, '32BF'::text, 1::float8, 1::float8
  ) AS rast
)
SELECT
  ST_Value(rast, 1, 1, 1, false),
  ST_Value(rast, 1, 1, 1, true)
FROM r;
-- 1 | NULL
```

That is an all-nodata raster, but it can be intentional. This check also makes the shorter overload fail whenever the reference nodata is `1`, since the default burn value is also `1`. I would lean toward allowing the caller to choose this.

common/src/main/java/org/apache/sedona/common/utils/RasterUtils.java
```diff
@@ -647,6 +647,114 @@ public static boolean isDataTypeIntegral(int dataTypeCode) {
...
+          throw new IllegalArgumentException(
+              String.format(
+                  "%s %s is not supported for pixel type '%s' (NaN and infinite values are not supported); use a finite value",
+                  argName, value, pixelType));
+        }
+        return value;
+    }
+
+    // Integer pixel types: reject out-of-range and fractional values (Math.rint also rejects NaN).
+    if (value < min || value > max || value != Math.rint(value)) {
```
jiayuasu, line 793: One signed-zero case still slips through here. `-0.0` passes the integer check because ordinary comparisons treat it as equal to `+0.0`, but integer storage cannot preserve the sign. This reproduces it:

```java
GridCoverage2D ref = RasterConstructors.makeEmptyRaster(
    1, "D", 3, 3, 0, 3, 1, -1, 0, 0, 0);
Geometry geom = Constructors.geomFromWKT(
    "POLYGON ((0 0, 3 0, 3 3, 0 3, 0 0), (1 1, 2 1, 2 2, 1 2, 1 1))", 0);
RasterConstructors.asRaster(geom, ref, "B", false, 1d, -0.0d, false);
```

For `B` this currently fails later with `Ranges [0..255] and [0..0] overlap`; `US` does the same, while `S` and `I` silently store `+0.0`. Since we are rejecting unrepresentable nodata values, could we reject negative zero for integer pixel types here? It should remain valid for `F` and `D`.

common/src/main/java/org/apache/sedona/common/raster/RasterConstructors.java
james-willis: what case could this possibly serve? To me it seems better to reject because its clearly a mistake.

paleolimbot: Sticter is better to start (if somebody actually wants to do this they can file an issue)

common/src/main/java/org/apache/sedona/common/utils/RasterUtils.java
paleolimbot: I don't think distinguishing -0.0 and +0.0 as a nodata value is something we should claim to support (it is almost certainly not distinguished anywhere else)

common/src/main/java/org/apache/sedona/common/raster/RasterConstructors.java
jiayuasu: ok fine by me

common/src/main/java/org/apache/sedona/common/raster/RasterConstructors.java
james-willis: Ok. I did this. I'm not sure about it still. People can always cast some literal to float.

common/src/main/java/org/apache/sedona/common/utils/RasterUtils.java
james-willis: I did this. not sure what it matters. Haven't bothered to think about what -0 means

jiayuasu (approved):

PR #3119, opened 2026-07-17
[GH-3118] Rasterize line segments by exact cell traversal instead of fixed-step sampling

### LINKED ISSUE #3118: Line rasterization uses fixed-step sampling instead of exact cell traversal, dropping boundary pixels (allTouched and LINESTRING)
## Expected behavior

Rasterizing a `LINESTRING`, and rasterizing a polygon with `allTouched = true`, should burn every pixel the geometry touches — this is what the [`RS_AsRaster` docs](https://sedona.apache.org/latest/api/sql/Raster-Operators/RS_AsRaster/) promise ("selects all pixels touched by the geometry") and what GDAL (`gdal_rasterize -at`, `rasterio.features.rasterize(all_touched=True)`) produces.

## Actual behavior

Pixels that the geometry edge crosses over a short distance are silently not burned.

A `LINESTRING` on a 6x6 unit raster:

```
RS_AsRaster of LINESTRING (3.97 1.57, 0.31 3.24):

Sedona                 GDAL / rasterio
0 0 0 0 0 0            0 0 0 0 0 0
0 0 0 0 0 0            0 0 0 0 0 0
1 0 0 0 0 0            1 0 0 0 0 0
0 1 1 0 0 0            1 1 1 1 0 0
0 0 0 1 0 0            0 0 0 1 0 0
0 0 0 0 0 0            0 0 0 0 0 0
```

Sedona drops pixels (row 3, col 0) and (row 3, col 3), which the line clearly passes through.

The same happens to polygon boundaries under `allTouched = true`. For `POLYGON ((3.7 1.28, 0.92 5.23, 4.26 4.15, 3.7 1.28))` on the same grid, Sedona omits pixel (row 3, col 4) — the boundary crosses it with a chord of 0.19 pixel-widths — where GDAL burns it.

The raster here is square with unit pixels, so this is unrelated to the non-square-pixel selection bug (a separate report).

```sql
-- LINESTRING: pixels (3,0) and (3,3) are missing
  ST_GeomFromWKT('LINESTRING (3.97 1.57, 0.31 3.24)'),
  'B', false, 1.0, 0.0, false), 1);

-- Polygon with allTouched = true: pixel (3,4) is missing
  ST_GeomFromWKT('POLYGON ((3.7 1.28, 0.92 5.23, 4.26 4.15, 3.7 1.28))'),
  'B', true, 1.0, 0.0, false), 1);
```

Comparison reference (Python):

```python
import numpy as np, rasterio.features
from affine import Affine
from shapely.wkt import loads

wkt = "LINESTRING (3.97 1.57, 0.31 3.24)"
print(rasterio.features.rasterize(
    [(loads(wkt), 1)], out_shape=(6, 6), fill=0,
    transform=Affine(1, 0, 0, 0, -1, 6), all_touched=True, dtype="uint8"))
```

## Root cause

`Rasterization.drawLineBresenham` is not a Bresenham line walk despite the name: it samples the segment at fixed 0.2-pixel steps and burns the pixel under each sample ([Rasterization.java#L204-L256 @ 0299fa8](https://github.com/apache/sedona/blob/0299fa8a380a/common/src/main/java/org/apache/sedona/common/raster/Rasterization.java#L204-L256)):

```java
drawLineBresenham(params, x0, y0, x1, y1, value, 0.2);
...
int steps = (int) Math.ceil(distance / stepSize);   // stepSize = 0.2
double stepX = dx / steps, stepY = dy / steps;
for (int i = 0; i <= steps; i++) {
  params.writableRaster.setSample((int) Math.floor(x), (int) Math.floor(y), 0, value);
  x += stepX; y += stepY;
}
```

A pixel is burned only if a sample lands in it, so any pixel the edge crosses with a chord shorter than the 0.2-pixel step can fall between two consecutive samples and be missed. This path is used for `LINESTRING` geometries (independent of `allTouched`) and for polygon-boundary pixels when `allTouched = true`, so it affects `RS_AsRaster`, `RS_Clip`, `RS_ZonalStats`/`RS_ZonalStatsAll`, and `RS_SetValues`. Introduced with the `allTouched` parameter in SEDONA-707 (#1788) and present on current master.

The fix is exact grid traversal (integer DDA / Amanatides–Woo), visiting each cell the segment enters. Reducing the step size only narrows the miss window; it cannot close it.

## Settings

Sedona version = 1.9.0 (also reproduced on current master)

Apache Spark version = 4.0.0

API type = SQL / Python

Scala version = 2.13

JRE version = 17

Python version = 3.10

james-willis:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #3118

`Rasterization` burned line/ring segments by sampling them at fixed 0.2-pixel steps and flooring each sample to a cell (`drawLineBresenham`). A cell is only burned if a sample lands inside it, so a cell the segment crosses over a chord shorter than the step could fall between two samples and be missed. This replaced the sampler with exact grid traversal (Amanatides–Woo): the segment is walked from one cell-boundary crossing to the next, so every cell it enters is burned. This fixes both `LINESTRING` rasterization and polygon-boundary selection under `allTouched = true` (which affects `RS_AsRaster`, `RS_Clip`, `RS_ZonalStats`/`RS_ZonalStatsAll`, and `RS_SetValues`).

This PR is stacked on #3113 (the non-square-pixel scanline fix); review that one first.

Small known deviation from GDAL: when a vertex lands exactly on a grid line, the geometry touches a cell only at a corner or edge. This exact traversal burns that cell (a superset consistent with the documented "all pixels touched by the geometry"), whereas GDAL omits it. This only occurs at exact grid-aligned vertices (never on general input) and GDAL's own behaviour there is inconsistent, so it is not pursued here — tracked in #3120.

## How was this patch tested?

- New `testAsRasterTouchedPixelsExactTraversal`: a polygon boundary and a `LINESTRING` that each clip a pixel over a sub-pixel chord, on a square unit grid; expected matrices from GDAL (`rasterio.features.rasterize`, all_touched=True). Written first, failed on the sampler.
- The seeded 100-case `allTouched` fuzz added on #3113 (`python/tests/sql/test_rasterize_parity.py`) matches GDAL exactly with this fix; its two strict-`xfail` markers (the allTouched polygon corpus and the LineString case) become passing assertions here.
- Existing `RS_AsRaster`/`RS_Clip`/`RS_ZonalStats`/`RS_SetValues` expectations that had encoded the sampler's missed boundary pixels are updated; each change was verified against GDAL/rasterio. `RS_ZonalStats` allTouched count on the edge-case geometry goes 1832 → 1842 (the 10 boundary pixels GDAL also selects). Full `common` module suite passes (1147 tests).


paleolimbot (approved):
I am not very familiar with raster algorithms in Sedona Java and this could use 👀 from someone who is, but this seems like it enables us to remove some known failures and is well tested. Thank you!

common/src/main/java/org/apache/sedona/common/raster/Rasterization.java
```diff
@@ -200,58 +200,74 @@ private static void rasterizeLineString(
...
-      // Only write if within raster bounds
-      if (rasterX >= 0
-          && rasterX < params.writableRaster.getWidth()
-          && rasterY >= 0
-          && rasterY < params.writableRaster.getHeight()) {
-        params.writableRaster.setSample(rasterX, rasterY, 0, value);
+      // On an exact corner crossing (segment passes through a lattice point) step both axes at
+      // once, so only the two cells the segment actually passes through are burned rather than the
+      // off-diagonal pair as well.
+      if (tMaxX < tMaxY) {
```
jiayuasu: Could this tie comparison account for floating-point drift? I may be missing an intended corner convention, but I can reproduce coordinate-order-dependent output when the segment crosses multiple grid corners.

Reproducer on this head:
```java
GridCoverage2D raster =
    RasterConstructors.makeEmptyRaster(1, "d", 20, 20, 0, 20, 1, -1, 0, 0, 0);
Geometry forward =
    Constructors.geomFromWKT("LINESTRING (1.25 18.75, 3.75 6.25)", 0);
Geometry reverse =
    Constructors.geomFromWKT("LINESTRING (3.75 6.25, 1.25 18.75)", 0);

double[] a = MapAlgebra.bandAsArray(
    RasterConstructors.asRaster(forward, raster, "d", false, 1d, 0d, false), 1);
double[] b = MapAlgebra.bandAsArray(
    RasterConstructors.asRaster(reverse, raster, "d", false, 1d, 0d, false), 1);
assertArrayEquals(a, b, 0d);
```

The assertion fails at band element 82 on `fabfe22`; the same test passes on base `93c7324`. My suspicion is that accumulated `tMax` values that represent the same crossing can differ by a few ULPs, causing one axis to step first and burn a corner-adjacent cell. Would a tolerance/ULP-aware tie check, or computing each crossing without accumulation, be safer?

james-willis: good call. I computed each crossing without the accumulation to avoid the ULP rounding drift.

jiayuasu: Recomputing fixes the original case, but exact equality between the two divisions can still be direction-dependent. Another repro on `55d13e6` is:

```text
50x50 raster, origin (0, 0), scale (1, -1)
LINESTRING (
  13.157894736842104 -20.385964912280702,
  23.157894736842106 -11.052631578947368
)
```

Rasterizing this forward and reversed changes band element 720 from 0 to 1.

One way to avoid the division/tolerance issue is to compare the segment against the next cell corner with the JTS robust orientation predicate:

```java
int crossingOrder;
if (stepX == 0) {
  crossingOrder = 1;
} else if (stepY == 0) {
  crossingOrder = -1;
} else {
  double gridX = stepX > 0 ? x + 1.0 : x;
  double gridY = stepY > 0 ? y + 1.0 : y;
  int orientation =
      CGAlgorithmsDD.orientationIndex(x0, y0, x1, y1, gridX, gridY);
  crossingOrder = -orientation * stepX * stepY;
}

if (crossingOrder < 0) {
  x += stepX;
} else if (crossingOrder > 0) {
  y += stepY;
} else {
  x += stepX;
  y += stepY;
}
```

I tried this locally; the six new traversal tests and this repro pass.

jiayuasu (approved):

PR #3211, opened 2026-07-29
[GH-3053] Add Flink 2.2 support

### LINKED ISSUE #3053: Support Apache Flink 2.x (currently capped at Flink 1.19)
### Problem

Sedona's Flink module is locked to Flink 1.x — 1.9.0 targets Flink 1.19, and the
latest dependency work (#2985) only moves to `flink-table-planner_2.12` 1.20.4,
still Flink 1.x and still Scala-suffixed. There's no path to run Sedona on Flink
2.x today.

This blocks teams standardizing on Flink 2.x — e.g. pipelines that need 2.x-era
Kafka source / deserialization APIs (Protobuf ingestion) can't drop below 2.0,
so Sedona's spatial functions are simply off the table. My use case is streaming
map-matching / spatial joins of vehicle positions against an indexed road network
— a spatial-index workload that doesn't reduce to point-level shapely calls.

### Scope

The geometry core (`sedona-common`) is Flink-agnostic; the break is confined to
`sedona-flink`:

- **Scala-free artifacts** — `sedona-flink_2.12` / `flink-table-planner_2.12` →
  the unsuffixed Flink 2.x table artifacts.
- **Removed legacy interfaces** — `SourceFunction` / `SinkFunction` / SinkV1 are
  gone in 2.0; move to Source/Sink v2.
- **Serialization** — Geometry `TypeInformation` / serializer registration likely
  needs updating for Flink 2.x.
- **Build** — a Flink 2.x variant alongside the 1.x line.

### Questions

1. Is Flink 2.x support already on anyone's radar / in progress?
2. Would a Flink-2.x compatibility module be accepted as a contribution (kept
   side-by-side with 1.x during transition), or is there a preferred shape first?

Happy to test against a real Flink 2.2 streaming workload if someone picks it up.

### Environment
- Sedona 1.9.0 · Flink target 2.2.0 · streaming Table/DataStream, geospatial map-matching

joshmanis618-stack:
## Summary

Fixes the two API breaks that block Sedona's `flink` module from compiling/running against Flink 2.x, and adds an opt-in `sedona-flink-2.2` profile so both lines are tested in CI.

- **Kryo registration**: `ExecutionConfig.registerTypeWithKryoSerializer` was removed in Flink 2.x; registration moved to `SerializerConfigImpl` via `getSerializerConfig()`.
- **TypeSerializerSnapshot signature change**: `resolveSchemaCompatibility` now takes a `TypeSerializerSnapshot<T>` instead of a `TypeSerializer<T>`. Fixed across all six type serializers (`Geometry`, `Geometry[]`, `Box2D`, `Box3D`, `Geography`, and a new `GeometryDoublePairTypeSerializer`).
- **`ST_MinimumBoundingRadius`**: its bare `@DataTypeHint(value = "RAW")` relied on Flink auto-deriving a Kryo serializer for `Pair<Geometry, Double>`, which throws under Flink 2.2.1. Added an explicit `GeometryDoublePairTypeSerializer`, matching how every other RAW hint in `Functions.java` already avoids auto-derivation.
- Added a `sedona-flink-2.2` Maven profile (`-Dflink=2.2 -Dscala=2.12`), mirroring the existing `sedona-spark-*` version-profile convention, pinning `flink.version` to 2.2.1.
- Added a Flink 2.2.1 row to the `java.yml` CI matrix and updated `docs/setup/flink/platform.md`.

## Why this doesn't need to wait for a breaking/2.0 release

When I filed #3053 I expected this to require migrating off legacy `SourceFunction`/`SinkFunction` (removed in Flink 2.0) and moving to unsuffixed Flink 2.x table artifacts. Neither turned out to be necessary in practice:

- The `flink` module doesn't use `SourceFunction`/`SinkFunction` anywhere.
- `flink-table-planner_2.12` is still published for Flink 2.2.1, so the existing Scala-suffixed artifact naming keeps working unchanged.

The actual break was narrower: just the two API changes above. **This PR is purely additive** — Flink 1.19 remains the default and is unaffected; `-Dflink=2.2 -Dflink.version=2.2.1` is opt-in. Since no existing behavior changes, I don't think this needs to wait for a breaking Sedona 2.0 release — happy to target the upcoming 1.9.1 if that works for you.

## Testing

Full `flink`/`flink-shaded` test suite (331 tests, 0 failures/errors) passes against both:
- Default Flink 1.19.0
- `-Dflink=2.2 -Dscala=2.12 -Dflink.version=2.2.1`

`mvn spotless:apply` run before submitting.

Closes #3053


joshmanis618-stack:
I did not touch the component that failed in this test, and it built fully on my local machine.

jiayuasu:
Thanks. I will look into the PR today

jiayuasu (commented):
Thanks for adding Flink 2.2 support. I think we should target this change for Sedona 2.0.0 and make Flink 1.19 the minimum supported version starting with that release. Flink 1.18 and earlier are no longer supported by the Apache Flink community, so Sedona 2.0.0 is a reasonable boundary for removing compatibility with those versions.

Before merging, please make the following changes:

1. Remove the `sedona-flink-2.2` Maven profile. Unlike the Spark profiles, it only assigns `flink.version`; there are no version-specific modules or source shims. Builds can use `-Dflink.version=2.2.1` directly.

2. Avoid casting `getSerializerConfig()` to the internal `SerializerConfigImpl`. With Flink 1.19 as the minimum, we can consistently use the public `pipeline.serialization-config` API introduced in Flink 1.19.

3. Update the platform and Maven-coordinate documentation to state that Sedona 2.0.0 requires Flink 1.19 or later. Please also record this minimum-version change in the Sedona 2.0.0 release notes.

4. Add `skipLibPostalTests: 'true'` to the new Flink matrix row. Otherwise it downloads and runs the unrelated 1.3 GB libpostal suite in a second matrix combination.

5. Document the checkpoint/savepoint compatibility impact of changing `ST_MinimumBoundingRadius` from a Kryo-backed RAW type to `GeometryDoublePairTypeSerializer`. Fresh stateless queries work, but stateful Table/SQL jobs containing this result cannot restore state written with the previous serializer. Since this is targeting Sedona 2.0.0, the compatibility break can be documented as part of the major release.

I will handle broader Flink-version coverage and separating the Sedona Spark and Sedona Flink CI workflows in a follow-up.

I verified that the profile-free Flink 2.2.1 build passes all 333 Flink tests with:

```bash
mvn -pl flink,flink-shaded -am test \
  -Dscala=2.12 -Dflink.version=2.2.1
```

joshmanis618-stack:
I think I’ve addressed your request

flink/src/main/java/org/apache/sedona/flink/SedonaContext.java
```diff
@@ -69,4 +60,40 @@ public static StreamTableEnvironment create(
...
+            kryoRegistration(GeometryCollection.class, GeometrySerde.class),
+            kryoRegistration(Circle.class, GeometrySerde.class),
+            kryoRegistration(Envelope.class, GeometrySerde.class),
+            kryoRegistration(Quadtree.class, SpatialIndexSerde.class),
+            kryoRegistration(STRtree.class, SpatialIndexSerde.class));
+
+    Configuration configuration = new Configuration();
+    configuration.set(PipelineOptions.SERIALIZATION_CONFIG, kryoRegistrations);
+    env.getConfig()
+        .getSerializerConfig()
```
jiayuasu: `ExecutionConfig.getSerializerConfig()` is annotated `@Internal` in both Flink 1.19 and 2.2, so this still depends on an internal API even though the `SerializerConfigImpl` cast is gone. Please apply the configuration through the public environment API instead:

```java
env.configure(configuration);
```

This also uses the environment’s user classloader rather than the current thread’s context classloader. I tested this exact replacement locally against Flink 1.19 and 2.2.1; `ModuleTest` passed all 7 tests with both versions.

jiayuasu (changes requested):
Please address the inline comment.

flink/src/main/java/org/apache/sedona/flink/SedonaContext.java
joshmanis618-stack: i think ive resolved your request

flink/src/main/java/org/apache/sedona/flink/SedonaContext.java
```diff
@@ -69,4 +60,40 @@ public static StreamTableEnvironment create(
...
+            kryoRegistration(MultiPolygon.class, GeometrySerde.class),
+            kryoRegistration(GeometryCollection.class, GeometrySerde.class),
+            kryoRegistration(Circle.class, GeometrySerde.class),
+            kryoRegistration(Envelope.class, GeometrySerde.class),
+            kryoRegistration(Quadtree.class, SpatialIndexSerde.class),
+            kryoRegistration(STRtree.class, SpatialIndexSerde.class));
+
+    Configuration configuration = new Configuration();
+    configuration.set(PipelineOptions.SERIALIZATION_CONFIG, kryoRegistrations);
+    env.configure(configuration);
```
jiayuasu: Could we keep the thread context class loader here? The one-argument overload uses the environment class loader in PyFlink, so the current PyFlink job fails with `ClassNotFoundException: org.locationtech.jts.geom.Point`. The public two-argument overload is available in both Flink 1.19 and 2.2:

```suggestion
    env.configure(configuration, Thread.currentThread().getContextClassLoader());
```

jiayuasu (approved):

joshmanis618-stack:
Thanks for the help, I look forward to using the changes when they land. Have a great week.

PR #3241, opened 2026-08-06
[GH-3240] Run Makefile prek commands through uv

### LINKED ISSUE #3240: Possible Docs and Makefile improvements involving prek from copilot reviews
## Pull request overview

Copilot reviewed 3 out of 3 changed files in this pull request and generated no new comments.

<details>
<summary>Suppressed comments (3)</summary>

**Makefile:68**
* `make update-deps` invokes `prek` directly (and then suppresses failures). If `prek` is only installed inside the uv-managed environment, this will be a no-op on fresh setups. Run it via `uv run` to ensure the hook update actually executes against the locked environment.
```
update-deps: check-install ## Update pre-commit hooks and dependency locks
	prek auto-update --freeze || :
	uv lock --upgrade
```
**Makefile:64**
* `make update` invokes `prek` directly. Since `prek` is declared as a dev dependency (installed into `.venv` by `uv sync`) and the docs describe uv-managed execution, this target will fail unless `prek` is already available on the user's PATH. Prefer running it through `uv run` so it uses the project-managed environment.

This issue also appears on line 66 of the same file.
```
update: install
	prek auto-update --freeze
```
**docs/setup/compile.md:192**
* This paragraph states that the Makefile+uv workflow guarantees `prek` runs in the locked environment, but the current Makefile targets invoke `prek` directly (not via `uv run`). Either update the Makefile to use `uv run prek ...` consistently, or soften this wording to avoid implying the environment is guaranteed.
```
The `Makefile` handles environment isolation and dependency resolution automatically using `uv`. For maximum convenience, you should interact with `prek` entirely through the following make shortcuts:

```
</details>

_Originally posted by @Copilot in https://github.com/apache/sedona/pull/3081#pullrequestreview-4838329446_

nightcityblade:

- Yes, I have read the [Contributor Rules](https://sedona.apache.org/latest/community/rule/) and [Contributor Development Guide](https://sedona.apache.org/latest/community/develop/).

- Yes, and the PR name follows the `[GH-XXX] my subject` format. Closes #3240.

- Run every Makefile `prek` invocation through `uv run`, including check, installation, and update targets.
- Ensure those targets consistently use the project-managed development environment, matching the existing documentation.

## How was this patch tested?

- `make -n check check-from-ref check-last check-stage update update-deps`
- `uv run prek run --files Makefile`
- `uv run prek run check-makefiles-tabs --hook-stage manual --files Makefile`
- `uv run prek run insert-license --files Makefile`
- `git diff --check`

The auto-update targets were inspected with `make -n` rather than executed because running them would intentionally update hook and dependency locks.

- No, this PR does not affect any public API, so no documentation change is needed. The existing documentation already describes `uv`-managed execution; this patch makes the Makefile honor that behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,19 +18,19 @@
 .PHONY: check check-from-ref check-install check-last check-stage clean docsbuild docsinstall install run-docs sync-no-dev update update-deps
 
 check: install ## run all pre-commit checks
-	prek run --all-files
+	uv run prek run --all-files
 
 check-from-ref: install ## will run prek checks for all changes since you branched off
-	prek run --from-ref master
+	uv run prek run --from-ref master
 
 check-install: ## checks for a uv installation
 	@command -v uv >/dev/null 2>&1 || (echo 'uv not found, install via: curl -LsSf https://astral.sh/uv/install.sh | sh' && exit 1)
 
 check-last: install ## will run pre-commit checks on last commit only
-	prek --last-commit
+	uv run prek --last-commit
 
 check-stage: install ## runs check on currently staged changes
-	prek
+	uv run prek
 
 clean:
 	@echo "Cleaning up generated files... (TODO)"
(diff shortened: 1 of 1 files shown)


jiayuasu (approved):

jiayuasu:
Thank you!

PR #3279, opened 2026-08-24
[GH-3271] Enhance GeoDataFrame and GeoSeries to handle authoritative …

### LINKED ISSUE #3271: GeoPandas: preserve authoritative no-CRS metadata for local geometry columns
## Problem

Sedona's distributed GeoPandas layer uses field metadata to distinguish a known CRS from a geometry column whose CRS must be inferred from its embedded SRID. A locally constructed geometry column with `crs=None` currently has no CRS metadata, so it is indistinguishable from a raw distributed geometry column with unknown provenance.

Reading `.crs` for such a metadata-free column performs a distributed `ST_SRID(first(...))` aggregation. APIs that inspect both operands' CRS, including binary geometry predicates, can therefore launch two eager Spark actions before returning an otherwise lazy result.

In a local Spark benchmark, comparing two CRS-less local operands submitted four Spark jobs before result materialization. Recording authoritative no-CRS metadata reduced that count to zero. The one-time constructor cost increased by approximately 2.3 ms for `GeoSeries` and 4.9 ms for `GeoDataFrame` in a warm one-row benchmark.

## Expected behavior

Sedona should represent three distinct states:

1. a known CRS stored as CRS metadata;
2. a known absence of CRS stored as explicit empty CRS metadata; and
3. an unknown raw or legacy distributed geometry column with no CRS metadata, which retains the existing embedded-SRID fallback.

The implementation should:

- record explicit no-CRS metadata for locally owned `GeoSeries` and `GeoDataFrame` geometry columns;
- preserve locally supplied per-column CRS values and synchronize their embedded SRIDs;
- propagate authoritative CRS state through `set_geometry`, WKT/WKB and file constructors, copies, selections, and all-null geometry columns;
- preserve public `crs=None` without erasing an embedded EWKB SRID;
- leave raw Spark and pandas-on-Spark geometry columns without metadata on the existing SRID-inference path;
- preserve Series names, indexes, active-geometry selection, and laziness; and
- add action-count, metadata propagation, raw-column fallback, and compatibility coverage across supported Spark and GeoPandas versions.

## Performance acceptance

- Constructing a binary predicate result from two locally created CRS-less operands should launch no Spark jobs solely for CRS discovery.
- Metadata-free raw distributed operands should continue to infer their embedded SRIDs and warn when they differ.
- The returned predicate and geometry plans should remain distributed and should not add Python execution nodes.

Part of #2230.

## Sedona version

2.0.0-SNAPSHOT

## API type

Python / distributed GeoPandas API

minhpham1810:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #3271

Addresses [issue #3271](https://github.com/apache/sedona/issues/3271): *"GeoPandas: preserve authoritative no-CRS metadata for local geometry columns"*

The issue reports that locally constructed `GeoSeries`/`GeoDataFrame` objects with `crs=None` have no CRS metadata, making them indistinguishable from raw distributed geometry columns of unknown provenance. This causes `.crs` access to trigger expensive distributed `ST_SRID` aggregation jobs to discover the CRS is absent.

### Changes

**`geoseries.py`**:
- Added `is_locally_owned` flag to distinguish locally-constructed objects (Python list, WKT/WKB, plain pandas/geopandas objects) from wrapped distributed structures
- Added `_record_no_crs_metadata()` helper method that stamps empty CRS metadata (empty string in metadata dict) using `_query_geometry_column()` with `crs_override=None`
- Stamps no-CRS metadata in `__init__` when `crs=None` and `is_locally_owned=True`
- Stamps no-CRS metadata in `_create_from_select()` for `from_wkt()`/`from_wkb()` paths (freshly built geometries with no embedded SRID)

**`geodataframe.py`**:
- Added `is_locally_owned` flag with the same logic as `GeoSeries`
- Added consolidated no-CRS stamping block at end of `__init__` that checks if geometry column exists, is locally owned, and has no CRS metadata
- Fixed `set_geometry()` array-like branch to construct `GeoSeries` directly from local data (avoid pre-wrapping in bare `pspd.Series`)
- Added import of `read_crs_metadata` helper

**Tests**:
- `test_geoseries.py`: Added 5 new tests covering local construction paths, raw column preservation, embedded SRID preservation, and no Spark/Python plan for CRS discovery
- `test_geodataframe.py`: Added 3 new tests covering named geometry column path and Spark/Python plan verification

## How was this patch tested?

- All 4 modified files pass `ast.parse` syntax validation
- Black formatting applied (where possible; test file formatting left untouched per existing repo state)
- Existing test suite patterns followed; no additional test execution possible in this environment (no Spark/geopandas available)
- Manual code review against implementation requirements and existing codebase patterns

The tests are structured to run with full Sedona build (Maven) once Spark/geopandas environment is available.

The changes are internal implementation details that preserve existing public behavior while fixing the underlying performance issue. The public `.crs` property semantics remain unchanged: locally constructed objects now correctly report `crs=None` without triggering distributed SRID lookups.

### DIFF
--- a/python/sedona/spark/geopandas/geodataframe.py
+++ b/python/sedona/spark/geopandas/geodataframe.py
@@ -851,20 +851,15 @@ class GeoDataFrame(GeoFrame, pspd.DataFrame):
                 "or by providing a DataFrame with column name 'geometry'",
             )
 
-        # A locally owned geometry column that still carries no CRS metadata
-        # at this point genuinely has no CRS (it was never derived from a raw
-        # or Sedona-managed source with its own CRS state) -- record that
+        # Locally owned geometry columns that still carry no CRS metadata at
+        # this point genuinely have no CRS (they were never derived from a
+        # raw or Sedona-managed source with its own CRS state) -- record that
         # explicitly so later `.crs` reads don't fall back to a distributed
-        # SRID lookup just to discover there isn't one.
-        if crs is None and is_locally_owned and self._geometry_column_name is not None:
-            active_geometry = self.geometry
-            has_crs_metadata, _ = read_crs_metadata(
-                active_geometry._internal.data_fields[0]
-            )
-            if not has_crs_metadata:
-                self[self._geometry_column_name] = (
-                    active_geometry._record_no_crs_metadata()
-                )
+        # SRID lookup just to discover there isn't one. This covers every
+        # geometry column, not just the active one, since inactive geometry
+        # columns can later become active via set_geometry().
+        if crs is None and is_locally_owned:
+            self._stamp_local_no_crs_metadata()
 
     # ============================================================================
     # GEOMETRY COLUMN MANAGEMENT
@@ -1759,6 +1754,60 @@ class GeoDataFrame(GeoFrame, pspd.DataFrame):
             data_fields=output_fields,
         )
         return PandasOnSparkDataFrame(result_internal)
+
+    def _stamp_local_no_crs_metadata(self) -> None:
+        """Record an authoritative "no CRS" state on every geometry column
+        that doesn't carry CRS metadata yet.
+
+        Built as a single projection over the whole frame, mirroring
+        ``_serialize_geometry_columns``: touching each geometry column
+        through a separately projected Series (e.g. ``self[col] =
+        col_series._record_no_crs_metadata()``) would cross pandas-on-Spark
+        anchors, and with ``compute.ops_on_diff_frames`` enabled could join
+        on duplicate indexes and multiply rows.
+        """
+        internal = self._internal.resolved_copy
+
+        needs_stamp = False
+        output_columns = []
+        output_fields = []
+        for spark_column, spark_name, field in zip(
+            internal.data_spark_columns,
+            internal.data_spark_column_names,
+            internal.data_fields,
+        ):
+            if (
+                isinstance(field.spark_type, GeometryType)
+                and not read_crs_metadata(field)[0]
+            ):
+                needs_stamp = True
+                field = with_crs_metadata(field, None).copy(name=spark_name)
+                spark_column = spark_column.alias(spark_name, metadata=field.metadata)
+            output_columns.append(spark_column)
+            output_fields.append(field)
+
+        if not needs_stamp:
+            return
+
+        output_sdf = internal.spark_frame.select(
+            *[
+                scol_for(internal.spark_frame, name)
+                for name in internal.index_spark_column_names
+            ],
+            *output_columns,
+            scol_for(internal.spark_frame, NATURAL_ORDER_COLUMN_NAME),
+        )
+        result_internal = internal.copy(
+            spark_frame=output_sdf,
+            index_spark_columns=[
+                scol_for(output_sdf, name) for name in internal.index_spark_column_names
+            ],
+            data_spark_columns=[
+                scol_for(output_sdf, name) for name in internal.data_spark_column_names
+            ],
+            data_fields=output_fields,
+        )
+        self._update_internal_frame(result_internal)
 
     @staticmethod
     def _validate_serialization_kwargs(method_name: str, kwargs: dict) -> None:
--- a/python/sedona/spark/geopandas/geodataframe.py
+++ b/python/sedona/spark/geopandas/geodataframe.py
@@ -57,7 +57,11 @@ from pyspark.sql.types import (
 )
 from pyspark.pandas.utils import log_advice
 
-from sedona.spark.geopandas._crs import copy_crs_metadata, with_crs_metadata
+from sedona.spark.geopandas._crs import (
+    copy_crs_metadata,
+    read_crs_metadata,
+    with_crs_metadata,
+)
 from sedona.spark.geopandas._explode import expand_geometry_column
 from sedona.spark.geopandas._typing import Label
 from sedona.spark.geopandas.base import GeoFrame
@@ -708,6 +712,23 @@ class GeoDataFrame(GeoFrame, pspd.DataFrame):
         from sedona.spark.geopandas import GeoSeries
         from pyspark.sql import DataFrame as SparkDataFrame
 
+        # Data built from a Python-native structure (dict, list of records,
+        # pandas/geopandas DataFrame, ...) is locally owned: we can safely
+        # record an authoritative "no CRS" state for its geometry column
+        # below when the caller didn't pass one. Wrapping an existing Sedona
+        # or raw distributed structure instead inherits whatever CRS state it
+        # already had.
+        is_locally_owned = not isinstance(
+            data,
+            (
+                GeoDataFrame,
+                GeoSeries,
+                PandasOnSparkDataFrame,
+                SparkDataFrame,
+                PandasOnSparkSeries,
+            ),
+        )
+
         if isinstance(data, (GeoDataFrame, GeoSeries)):
             if crs:
                 data.crs = crs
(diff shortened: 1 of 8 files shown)


python/sedona/spark/geopandas/geodataframe.py
```diff
@@ -830,6 +851,21 @@ def __init__(
...
+        # or Sedona-managed source with its own CRS state) -- record that
+        # explicitly so later `.crs` reads don't fall back to a distributed
+        # SRID lookup just to discover there isn't one.
+        if crs is None and is_locally_owned and self._geometry_column_name is not None:
+            active_geometry = self.geometry
+            has_crs_metadata, _ = read_crs_metadata(
+                active_geometry._internal.data_fields[0]
+            )
+            if not has_crs_metadata:
+                self[self._geometry_column_name] = (
```
jiayuasu: I think this assignment can cross pandas-on-Spark anchors. `_record_no_crs_metadata()` builds a projected frame, so assigning its Series back here goes through different-frame alignment.

A small reproducer is:

```python
with ps.option_context("compute.ops_on_diff_frames", False):
    sgpd.GeoDataFrame(
        {"zone": ["a"], "geometry": [Point(0, 0)]}
    )
```

The Spark 3.4 CI run currently raises `ValueError: Cannot combine ... different dataframe` at this line. With cross-frame operations enabled, duplicate indexes may be joined and multiply rows instead. Could we attach the alias and updated `InternalField` through the existing internal frame, for example with `with_new_spark_column`, rather than assigning a derived Series?

python/sedona/spark/geopandas/geoseries.py
```diff
@@ -662,6 +673,26 @@ def __init__(
...
+
+    def _record_no_crs_metadata(self, inplace: bool = False) -> "GeoSeries":
+        """Record authoritative "no CRS" metadata without touching geometry bytes.
+
+        Unlike ``set_crs``, this never calls ``ST_SetSRID``, so any SRID
+        already embedded in the geometry (e.g. from WKB/EWKB input) survives
+        untouched even though the public ``.crs`` becomes authoritatively
+        ``None`` instead of falling back to a distributed SRID lookup.
+        """
+        result = self._query_geometry_column(
```
jiayuasu, line 695: Could this path preserve valid falsy names?

```python
s = sgpd.GeoSeries([Point(0, 0)], name=0)
assert s.name == 0

s = sgpd.GeoSeries([Point(0, 0)], name="")
assert s.name == ""
```

Both names appear to become `None` because `_query_geometry_column(..., keep_name=True)` later checks `if keep_name and self.name`. A metadata-only update that keeps the existing label would avoid that; alternatively, the helper could test `self.name is not None`.

python/sedona/spark/geopandas/geoseries.py
```diff
@@ -662,6 +673,26 @@ def __init__(
 
         if crs is not None:
             self.set_crs(crs, inplace=True, allow_override=True)
+        elif is_locally_owned:
```
jiayuasu, line 684: I think a CRS carried by a local `GeometryArray` is lost before this branch:

```python
arr = gpd.GeoSeries([Point(0, 0)], crs=4326).array
s = sgpd.GeoSeries(arr)
assert s.crs.to_epsg() == 4326
```

Only a `gpd.GeoSeries` is checked for an inherited CRS above; a `GeometryArray`, or a plain `pd.Series` backed by one, is converted to object and then stamped as authoritative `None` here. Could we read `data.crs` / `data.array.crs` before conversion and use that value when setting the metadata and embedded SRID?

python/sedona/spark/geopandas/geodataframe.py
```diff
@@ -830,6 +851,21 @@ def __init__(
                 "or by providing a DataFrame with column name 'geometry'",
             )
 
+        # A locally owned geometry column that still carries no CRS metadata
+        # at this point genuinely has no CRS (it was never derived from a raw
+        # or Sedona-managed source with its own CRS state) -- record that
+        # explicitly so later `.crs` reads don't fall back to a distributed
+        # SRID lookup just to discover there isn't one.
+        if crs is None and is_locally_owned and self._geometry_column_name is not None:
```
jiayuasu: Could we apply this to every locally created geometry column rather than only the active one?

```python
gdf = sgpd.GeoDataFrame({
    "geometry": [Point(0, 0)],
    "secondary": [Point(1, 1)],
})
has_metadata, _ = read_crs_metadata(
    gdf["secondary"]._internal.data_fields[0]
)
assert has_metadata
```

The secondary column still appears metadata-free, so `gdf.set_geometry("secondary").crs` takes the distributed SRID fallback. For pandas/GeoPandas inputs with different per-column CRSs, those inactive CRS values can also be lost during the object conversion. One option would be to capture CRS by column position before conversion, then update every geometry field with either its CRS or explicit no-CRS metadata.

python/sedona/spark/geopandas/geoseries.py
```diff
@@ -588,6 +588,17 @@ def __init__(
...
             crs = data.crs
 
+        # Locally owned data (a Python list, WKT/WKB, a plain pandas Series, a
+        # bare geopandas.GeoSeries, ...) is distinct from wrapping an existing
+        # Sedona-managed or raw distributed structure: only in the former case
+        # do we know, right now, whether a CRS is set, so only there do we
+        # need to record an authoritative "no CRS" state below when the user
+        # didn't pass one. Wrapping an existing structure just inherits
+        # whatever CRS state (known, known-absent, or unknown) it already had.
+        is_locally_owned = not isinstance(
```
jiayuasu, line 606: Could the file-constructor path be handled explicitly? `GeoSeries.from_file` passes `df.geometry` here, so a CRS-less file is classified as distributed and remains metadata-free.

A regression check could be:

```python
s = sgpd.GeoSeries.from_file(crsless_geojson, format="geojson")
assert read_crs_metadata(s._internal.data_fields[0]) == (True, None)
```

At the moment `from_file` also reads `df.crs` eagerly, which can run SRID inference, and a later `s.crs` can repeat it. Could the reader attach the authoritative known or known-absent state directly from the file metadata instead?

python/tests/geopandas/test_geoseries.py
```diff
@@ -6678,3 +6679,104 @@ def test_crs_metadata_propagation(self):
...
+        """Accessing .crs on a locally constructed GeoSeries with no CRS
+        metadata should not launch any Spark jobs - it should read metadata
+        directly without SRID inference."""
+        s = sgpd.GeoSeries([Point(1, 1), Point(2, 2)])
+        
+        # Capture the number of pending jobs before access
+        before_jobs = 0
+        try:
+            _ = s.crs
+            after_jobs = len(self.session._sc._jvm.scala.concurrent.FutureCache.getPythonPlans(s))
```
jiayuasu: I may be missing a fixture, but `TestBase` exposes `spark` and `sc`, not `session`. That would make this line raise `AttributeError`, and the broad `except Exception` then turns the result into zero jobs, so the test still passes without measuring anything. `before_jobs` is also unused.

Could we use a Spark job group/status tracker around the operation instead, without swallowing instrumentation errors? For example, assign a unique job group, construct the required binary predicate from two local CRS-less operands, and assert `self.sc.statusTracker().getJobIdsForGroup(group)` stays empty before materialization. The companion plan test could inspect `queryExecution().optimizedPlan()` and check for `BatchEvalPython`, `ArrowEvalPython`, and `PythonUDF`.

python/tests/geopandas/test_geodataframe.py
```diff
@@ -1041,3 +1056,45 @@ def test_to_arrow(self):
 
 def check_geodataframe(df):
     assert isinstance(df, GeoDataFrame)
+
+    def test_local_construction_no_spark_job_for_crs_discovery(self):
```
jiayuasu, line 1187: It looks like these two tests are nested inside the top-level `check_geodataframe` helper, so pytest does not collect them.

This should show the issue:

```bash
pytest --collect-only -q python/tests/geopandas/test_geodataframe.py |
  grep local_construction_no
```

Only the earlier named-geometry test is collected. Could we move both methods into `TestGeoDataFrame`, then use the same real job-count and query-plan assertions as the GeoSeries coverage?

jiayuasu (commented):
I left a few inline questions with small reproducers and possible fixes.

python/sedona/spark/geopandas/geoseries.py
```diff
@@ -4376,7 +4407,13 @@ def _create_from_select(
         ps_series = first_series(PandasOnSparkDataFrame(internal))
         name = None if name == SPARK_DEFAULT_SERIES_NAME else name
         ps_series.rename(name, inplace=True)
-        return GeoSeries(ps_series, index, crs=crs)
+        result = GeoSeries(ps_series, index, crs=crs)
+        if crs is None:
+            # ST_GeomFromWKB/WKT always produce fresh geometries with no
```
jiayuasu: Could we soften this rationale? `ST_GeomFromWKB` can preserve an SRID carried by EWKB.

For example:

```python
geom = shapely.set_srid(Point(0, 0), 4326)
ewkb = shapely.to_wkb(geom, include_srid=True)
s = sgpd.GeoSeries.from_wkb([ewkb])
assert s._internal.spark_frame.select(
    stf.ST_SRID(s.spark.column)
).first()[0] == 4326
```

The metadata-only stamping still seems right. Suggested wording: WKT and ordinary WKB are generally SRID-less, while EWKB may carry an embedded SRID; recording public `crs=None` must not rewrite it.

jiayuasu:
@minhpham1810 I've taken over this PR and made some updates given its complexity. Everything looks good now. Thank you for your contribution!

PR #3292, opened 2026-08-27
[GH-3174] Align ST_IsPolygonCW/ST_IsPolygonCCW with PostGIS polygon orientation semantics

### LINKED ISSUE #3174: Align polygon orientation predicates with PostGIS in Sedona 2.0.0
## Expected behavior

In Sedona 2.0.0, or another major release, `ST_IsPolygonCW` and `ST_IsPolygonCCW` should align with the full PostGIS polygonal-component semantics:

- Recursively inspect polygonal components inside geometry collections.
- Ignore non-polygonal components.
- Return `true` when an input has no polygonal components, including points, line strings, and empty geometry collections.
- Return `NULL` for SQL `NULL` inputs across all engines.

This follows the documented PostGIS behavior:

- https://postgis.net/docs/ST_IsPolygonCW.html
- https://postgis.net/docs/ST_IsPolygonCCW.html

## Actual behavior

Sedona currently returns `false` for non-polygonal inputs such as points and line strings. It also returns `false` for geometry collections instead of recursively evaluating their polygonal members.

Changing these results from `false` to `true` is user-visible and potentially breaking. PR #3173 should therefore remain scoped to fixing empty `Polygon` and `MultiPolygon` handling without changing established results for other geometry types.

The Flink wrappers also currently use primitive `boolean` return types, so matching PostGIS SQL null propagation requires changing them to nullable `Boolean`.

```sql
-- PostGIS returns true from both predicates because there are no polygonal components.
SELECT
    ST_IsPolygonCW(ST_GeomFromWKT('POINT (0 0)')),
    ST_IsPolygonCCW(ST_GeomFromWKT('POINT (0 0)'));

SELECT
    ST_IsPolygonCW(ST_GeomFromWKT('LINESTRING (0 0, 1 0, 0 0)')),
    ST_IsPolygonCCW(ST_GeomFromWKT('LINESTRING (0 0, 1 0, 0 0)'));

-- PostGIS ignores the point and recursively checks the clockwise polygon.
SELECT
    ST_IsPolygonCW(
        ST_GeomFromWKT(
            'GEOMETRYCOLLECTION (
                POINT (2 2),
                GEOMETRYCOLLECTION (
                    POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))
                )
            )'
        )
    );
```

Sedona returns `false` for these cases. The corresponding PostGIS results are `true`.

Because this change alters existing predicate results, it should include migration notes and be released only with a major-version compatibility boundary.

## Settings

Sedona version = 1.9.1-SNAPSHOT

Apache Spark version = 3.4

Apache Flink version = 1.19

API type = Common Java functions, Spark SQL, Flink SQL, Snowflake UDFs

Scala version = 2.12

JRE version = 11 and 17

Python version = N/A

Environment = Standalone

Related issues and pull requests: #3171, #3173

minhpham1810:

Closes #3174.

- Reworked `isPolygonCW`/`isPolygonCCW` in `common/Functions.java` to recursively inspect polygonal components (`Polygon`, `MultiPolygon`, and nested `GeometryCollection` at any depth) and vacuously return `true` for inputs with no polygonal components (points, line strings, empty geometry collections), matching PostGIS semantics. Both predicates share one recursive helper (`allPolygonalComponentsMatchOrientation`) to avoid duplicating the traversal logic.
- Changed the Flink `ST_IsPolygonCW`/`ST_IsPolygonCCW` wrappers from primitive `boolean` to nullable `Boolean` with an explicit null check, so a SQL `NULL` input now returns `NULL` instead of `false`.
- Updated documentation for `ST_IsPolygonCW`/`ST_IsPolygonCCW` across the SQL, Flink, and Snowflake docs, and added a breaking-change entry to the Sedona 2.0.0 release notes.
- Added test coverage for non-polygonal geometries, nested `GEOMETRYCOLLECTION` recursion, and `NULL` propagation across the common, Spark SQL, Flink, and Snowflake test suites.

No code changes were needed in the Snowflake or Spark UDF wrappers: Snowflake's Java UDF runtime already short-circuits `NULL` input before invoking the handler, and Spark's `InferredExpression` already short-circuits `NULL` geometry arguments before calling the underlying common function.

Since `master` is already `2.0.0-SNAPSHOT`, this breaking change lands directly at the major-version boundary called out in the issue, without needing a compatibility flag.

## How was this patch tested?

- `mvn -pl common -Dtest=FunctionsTest test` (328 tests, including new coverage for nested `GeometryCollection` recursion and mixed orientation)
- `mvn -pl flink -Dtest=FunctionTest#testIsPolygonCW,testIsPolygonCCW,testIsPolygonOrientationForEmptyGeometries,testIsPolygonOrientationForNonPolygonalAndCollectionGeometries,testIsPolygonOrientationNullPropagation test`
- `mvn -Dsnowflake=true -pl snowflake-tester -am -DskipTests package` (compiles cleanly; tests require live Snowflake credentials so were not executed)
- Spark: added coverage to `functionTestScala.scala`. Could not get a full `spark/common` reactor build in the local dev environment due to a pre-existing, unrelated `protobuf-java`/generated-sources mismatch in the OSM PBF datasource reader (`package proto4 does not exist`); no code outside the added Scala test was touched in the Spark module, and the fix relies on `InferredExpression`'s existing null-short-circuiting, which is exercised by many other predicates in this codebase.

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/Functions.java
+++ b/common/src/main/java/org/apache/sedona/common/Functions.java
@@ -26,6 +26,7 @@ import com.uber.h3core.exceptions.H3Exception;
 import com.uber.h3core.util.LatLng;
 import java.util.*;
 import java.util.List;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.sedona.common.S2Geography.Geography;
@@ -1635,29 +1636,19 @@ public class Functions {
   }
 
   /**
-   * This function accepts Polygon and MultiPolygon, if any other type is provided then it will
-   * return false. If the exterior ring is clockwise and the interior ring(s) are counter-clockwise
-   * then returns true, otherwise false. Empty Polygon and MultiPolygon inputs return true because
-   * they contain no rings with the opposite orientation.
+   * Recursively inspects the polygonal components of the input (Polygon, MultiPolygon, or any
+   * nested GeometryCollection), and returns true if every one of them has a clockwise exterior
+   * ring and counter-clockwise interior ring(s). Non-polygonal components (points, line strings)
+   * are ignored. Inputs with no polygonal components at all, including points, line strings, and
+   * empty geometry collections, vacuously return true, matching PostGIS semantics. Empty Polygon
+   * and MultiPolygon inputs also return true because they contain no rings with the opposite
+   * orientation.
    *
-   * @param geom Polygon or MultiPolygon
+   * @param geom any geometry
    * @return
    */
   public static boolean isPolygonCW(Geometry geom) {
-    if (geom instanceof MultiPolygon) {
-      MultiPolygon multiPolygon = (MultiPolygon) geom;
-
-      for (int i = 0; i < multiPolygon.getNumGeometries(); i++) {
-        if (!checkIfPolygonCW((Polygon) multiPolygon.getGeometryN(i))) {
-          return false;
-        }
-      }
-      return true;
-    } else if (geom instanceof Polygon) {
-      return checkIfPolygonCW((Polygon) geom);
-    }
-    // False for remaining geometry types
-    return false;
+    return allPolygonalComponentsMatchOrientation(geom, Functions::checkIfPolygonCW);
   }
 
   private static boolean checkIfPolygonCW(Polygon geom) {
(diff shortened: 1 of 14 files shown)


common/src/main/java/org/apache/sedona/common/Functions.java
```diff
@@ -1681,6 +1672,30 @@ private static boolean checkIfPolygonCW(Polygon geom) {
...
+      for (int i = 0; i < collection.getNumGeometries(); i++) {
+        if (!allPolygonalComponentsMatchOrientation(
+            collection.getGeometryN(i), ringOrientationCheck)) {
+          return false;
+        }
+      }
+      return true;
+    }
+    // No polygonal component to violate the orientation for remaining geometry types
+    return true;
```
jiayuasu, line 1699: Could we keep the previous behavior for a direct Java null here? It looks like null misses both branches and falls through to true, so Functions.isPolygonCW(null) and isPolygonCCW(null) now both return true instead of the previous false. The SQL wrappers seem to handle null separately, but direct common callers can still reach this. Maybe add a null guard and a small common test?

jiayuasu (approved):

jiayuasu:
Thank you!

Issue #3004, opened 2026-05-27
Website 404 in console for `sitemap.xml`


jbampton:
Load the website and open the developer tools network tab and see the 404:

<img width="1147" height="249" alt="Image" src="https://github.com/user-attachments/assets/dee55cb4-86ed-468d-8fb6-d604e8184748" />

MohammedThowfiq:
I am working on this. The screenshot points to /latest/zh/sitemap.xml, which still returns 404 while /latest/sitemap.xml returns 200. I will check the localized MkDocs build output and add a regression check for the sitemap URLs used by the language links.

PR #3309, opened 2026-09-02
[GH-3004] Publish sitemaps at documentation language roots

MohammedThowfiq:

Yes, I have read the Contributor Rules and Contributor Development Guide.

Related to #3004. This addresses the `/latest/zh/sitemap.xml` 404 shown in the issue's screenshot.

Material requests a sitemap from the Chinese language root, but mkdocs-static-i18n writes the combined sitemap only at the site root. A post-build hook copies the XML and compressed sitemap to each enabled, non-default language root after the nested language builds finish.

The existing documentation build checks both copies with `cmp`. No separate test harness or additional dependencies are needed.

This fixes language-root requests; handling of contextual alternate links on nested pages is unchanged.

## How was this patch tested?

- The baseline documentation build reproduced the missing Chinese sitemap.
- The full fixed MkDocs build and both sitemap comparisons passed locally.
- Bandit 1.9.4 passed for the documentation hooks with the repository configuration; no suppressions were added.
- YAML lint and `git diff --check` passed.

The local MkDocs build has existing missing generated API documentation/link warnings. Java/Scala/R/Python API documentation builds were not run locally; the previous upstream documentation workflow passed.

No public API changes. The hook documents its purpose.


jiayuasu:
Thanks for the fix. This seems like a lot of code for one missing sitemap. Could we simplify it, especially the test? The current version also fails the Bandit check.

MohammedThowfiq:
Simplified the checks in 583dcb1: removed the 94-line test harness and its separate workflow step. The existing MkDocs build now verifies the XML and gzip sitemap copies with two cmp commands. Bandit 1.9.4 passes with the repository configuration, without suppressions. The full local MkDocs build and both comparisons also pass.

jiayuasu (approved):
Thank you!

PR #3312, opened 2026-09-02
[GH-3311] Make RS_SetBandNoDataValue honor NULL to remove a band's no-data value in Spark SQL

### LINKED ISSUE #3311: RS_SetBandNoDataValue: documented "NULL removes the no-data value" behavior is unreachable from Spark SQL
## Expected behavior

Per the [RS_SetBandNoDataValue docs](https://sedona.apache.org/latest/api/sql/Raster-Operators/RS_SetBandNoDataValue/):

> Passing a `null` value for `noDataValue` will remove the no data value and that will ensure all pixels are included in functions rather than excluded as no data.

So `RS_SetBandNoDataValue(raster, 1, NULL)` should return the input raster with band 1's no-data value removed.

The common Java implementation (`RasterBandEditors.setBandNoDataValue`) actually implements this: its `noDataValue == null` branch strips the no-data value from the sample dimension and returns the cloned raster.

This is also the semantics of PostGIS [`ST_SetBandNoDataValue`](https://postgis.net/docs/RT_ST_SetBandNoDataValue.html) ("To mark a band as having no nodata value, set the nodata value = NULL"), and the semantics SedonaDB adopted in apache/sedona-db#1158, so Sedona Spark is currently inconsistent with both.

## Actual behavior

From Spark SQL, `RS_SetBandNoDataValue(raster, 1, NULL)` (and the 2-argument form with a NULL no-data value) returns a NULL raster. The documented "clear the no-data value" code path is unreachable from SQL.

Cause: the `RS_SetBandNoDataValue` catalyst expression wraps the Java function with `inferrableFunction2/3/4` (`InferrableFunctionConverter.scala`), and those evaluators return `null` as soon as *any* argument is null — the wrapped Java function is never invoked, so its `noDataValue == null` branch never runs.

The existing test in `rasteralgebraTest.scala` ("Passed RS_SetBandNoDataValue with raster") cannot distinguish the two behaviors:

```scala
val actualNull =
  df.selectExpr("RS_BandNoDataValue(RS_SetBandNoDataValue(raster, 1, null))").first().get(0)
assertNull(actualNull)
```

`RS_BandNoDataValue` returns null both when the raster's no-data value was cleared **and** when the raster itself is null, so the assertion passes under either semantics.

## Proposed fix

The codebase already has a pattern for this: `InferrableFunction.allowRightNull` (`InferredExpression.scala`), used by `ST_MakePolygon`, `ST_CollectionExtract`, etc., which lets the trailing argument be null while still null-propagating on a null first argument. It currently only exists for arity 2.

- Add an arity-3 variant (allowing the third argument to be null, still returning null if the raster or band index is null).
- Use it for the 2- and 3-argument overloads of `RS_SetBandNoDataValue`.
- Leave the 4-argument `replace` overload null-propagating: replacing pixel values with a null no-data value is meaningless, and the docs can note that clearing requires the 2-/3-argument form.
- Strengthen the test so it distinguishes clearing from null-propagation (assert the returned raster itself is non-null, then that `RS_BandNoDataValue` on it is null).

I have a patch ready and will open a PR.

Related (not duplicates): #1269 and #1148 both stem from confusion around the `replace` variant and null no-data values, which this inconsistency feeds into.

## Steps to reproduce

```sql
SELECT RS_SetBandNoDataValue(
         RS_SetBandNoDataValue(
           RS_MakeEmptyRaster(1, 20, 20, 2, 22, 2, 3, 1, 1, 0),
           -999),
         NULL) IS NULL;
-- returns true; expected: a non-null raster with no no-data value on band 1
```

## Sedona version

master (2.0.0-SNAPSHOT); behavior dates back to the introduction of the function (v1.5.0)

## Apache Spark version

3.5 (applies to all supported versions — the affected code is in `spark/common`)

## API type

SQL

## Scala version

2.12

## JRE version

17

## Environment

any

## Existing issues

james-willis:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #3311

Scoped down to just making a `NULL` `noDataValue` behave as documented. The `GC_NODATA` property work that grew out of the review is split into a follow-up, since it turned out to be a separate problem with its own failure modes (see the end of this description).

**Make the documented NULL behavior reachable from SQL.** The docs and the common Java implementation both say a `null` `noDataValue` removes the band's no-data value, but from Spark SQL that path was unreachable: the catalyst expression wrapped `RasterBandEditors.setBandNoDataValue` with `inferrableFunction2/3/4`, which null-propagate as soon as any argument is null, so `RS_SetBandNoDataValue(raster, 1, NULL)` returned a NULL raster instead. This matches PostGIS `ST_SetBandNoDataValue` NULL-clears semantics and the behavior SedonaDB adopted in apache/sedona-db#1158.

Added `InferrableFunction.allowRightNull3`, an arity-3 counterpart of the existing `allowRightNull` pattern (allows the third argument to be null; still null-propagates when the raster or band index is null), and wired the 2- and 3-argument overloads through it. It is a separate name rather than an overload because overloading `allowRightNull` breaks eta-expansion of the overloaded functions passed to it at existing call sites. The 4-argument `replace` overload still null-propagates, since replacing pixels with a null no-data value is meaningless; the docs now say so.

**Clear the target band, not band 1.** The null path consulted band 1's no-data value to decide whether there was anything to remove, so clearing band N was a no-op whenever band 1 had none. Latent until now because the path was unreachable from SQL and the Java tests only ever cleared band 1.

**Stop `RS_AsGeoTiff` inventing a no-data value.** `GeoTiffWriter` writes a default `GDAL_NODATA` of 0 when a coverage has no no-data value, so a raster that never had one came back from a write/read round trip reporting `0`. This is independent of NULL handling, but it is what made a *cleared* no-data value reappear on the round trip, so it is fixed here: `WRITE_NODATA` is set to false in that case only, leaving an explicit `-Dgeotiff.writenodata=false` untouched.

## How was this patch tested?

The existing `RS_SetBandNoDataValue` test could not distinguish clearing from null-propagation — `RS_BandNoDataValue` returns null for a null raster too, so it passed under both semantics. It now sets a no-data value, clears it with NULL through both the 2- and 3-argument forms, and asserts the resulting raster is non-null while `RS_BandNoDataValue` on it is null; a NULL raster input still yields NULL.

New coverage for the other two fixes: clearing band 2 of a raster whose band 1 has no no-data value (both at the Java and SQL levels), a SQL round trip asserting a cleared value does not come back through `RS_AsGeoTiff`, and two `RasterOutputTest` cases pinning that a raster without a no-data value round-trips without gaining one while a raster with one keeps it.

Each new assertion fails without its fix. Full runs: `common` 1339 passed, `rasteralgebraTest` 168 passed, `rasterIOTest` 21 passed.

## Follow-up, not in this PR

`GC_NODATA` is a single coverage-wide property while no-data is declared per band, and the two are not kept in sync in either direction. This PR deliberately does not touch that:

- Setting a no-data value only writes the band's sample dimension. On a raster with an existing `GC_NODATA` the declared value is then ignored by anything reading the property — `RS_MapAlgebra` keeps using the old sentinel. On a raster without one, nothing creates it, so a freshly set no-data value has no effect on `RS_MapAlgebra` at all.
- Clearing has the mirror problem, and removing the property is not a safe fix on its own: because it is coverage-wide, dropping it while another band still declares that value changes *that* band's results (#3324).
- A GeoTIFF also cannot represent a heterogeneous per-band state — the container has one `GDAL_NODATA` slot, and GDAL itself spills a divergent per-band value into a `.aux.xml` sidecar rather than the TIFF, which `RS_AsGeoTiff` has no equivalent of since it returns bytes.

Keeping the property in step with the sample dimensions in both directions, and deciding what `RS_AsGeoTiff` should do with a heterogeneous state, belongs in a separate PR.

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/raster/RasterBandEditors.java
+++ b/common/src/main/java/org/apache/sedona/common/raster/RasterBandEditors.java
@@ -53,7 +53,9 @@ public class RasterBandEditors {
 
     // Remove no-Data if it is null
     if (noDataValue == null) {
-      if (RasterBandAccessors.getBandNoDataValue(raster) == null) {
+      // Consult the target band, not band 1: otherwise clearing band N is a no-op
+      // whenever band 1 happens to have no no-data value.
+      if (rasterNoData == null) {
         return raster;
       }
       GridSampleDimension[] sampleDimensions = raster.getSampleDimensions();
--- a/common/src/main/java/org/apache/sedona/common/raster/RasterOutputs.java
+++ b/common/src/main/java/org/apache/sedona/common/raster/RasterOutputs.java
@@ -47,6 +47,7 @@ import org.geotools.coverage.grid.GridCoverage2D;
 import org.geotools.coverage.grid.io.AbstractGridFormat;
 import org.geotools.gce.arcgrid.ArcGridWriteParams;
 import org.geotools.gce.arcgrid.ArcGridWriter;
+import org.geotools.gce.geotiff.GeoTiffFormat;
 import org.geotools.gce.geotiff.GeoTiffWriteParams;
 import org.geotools.gce.geotiff.GeoTiffWriter;
 import org.geotools.referencing.operation.transform.AffineTransform2D;
@@ -62,6 +63,20 @@ public class RasterOutputs {
       throw new RuntimeException(e);
     }
     ParameterValueGroup defaultParams = writer.getFormat().getWriteParameters();
+    // GeoTiffWriter writes a default GDAL_NODATA of 0 for coverages that have no no-data
+    // value, so a cleared (or never set) no-data value would come back as 0 when the
+    // written bytes are read again. Only opt out in that case: forcing the flag on when a
+    // no-data value is present would override an explicit -Dgeotiff.writenodata=false.
+    boolean hasNoDataValue = false;
+    for (int band = 1; band <= raster.getNumSampleDimensions(); band++) {
+      if (RasterBandAccessors.getBandNoDataValue(raster, band) != null) {
+        hasNoDataValue = true;
+        break;
+      }
+    }
+    if (!hasNoDataValue) {
+      defaultParams.parameter(GeoTiffFormat.WRITE_NODATA.getName().toString()).setValue(false);
+    }
     if (compressionType != null && compressionQuality >= 0 && compressionQuality <= 1) {
       GeoTiffWriteParams params = new GeoTiffWriteParams();
       params.setCompressionMode(ImageWriteParam.MODE_EXPLICIT);
--- a/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/raster/RasterBandEditors.scala
+++ b/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/raster/RasterBandEditors.scala
@@ -22,13 +22,15 @@ import org.apache.sedona.common.raster.RasterBandEditors
 import org.apache.spark.sql.catalyst.expressions.Expression
 import org.apache.spark.sql.sedona_sql.expressions.InferrableFunctionConverter._
 import org.apache.spark.sql.sedona_sql.expressions.InferrableRasterTypes._
-import org.apache.spark.sql.sedona_sql.expressions.InferredExpression
+import org.apache.spark.sql.sedona_sql.expressions.{InferrableFunction, InferredExpression}
 
 private[apache] case class RS_SetBandNoDataValue(inputExpressions: Seq[Expression])
     extends InferredExpression(
       inferrableFunction4(RasterBandEditors.setBandNoDataValue),
-      inferrableFunction3(RasterBandEditors.setBandNoDataValue),
-      inferrableFunction2(RasterBandEditors.setBandNoDataValue)) {
+      // A null noDataValue removes the band's no-data value, so let it through instead of
+      // null-propagating the whole expression.
+      InferrableFunction.allowRightNull3(RasterBandEditors.setBandNoDataValue),
+      InferrableFunction.allowRightNull(RasterBandEditors.setBandNoDataValue)) {
   protected def withNewChildrenInternal(newChildren: IndexedSeq[Expression]) = {
     copy(inputExpressions = newChildren)
   }
--- a/common/src/test/java/org/apache/sedona/common/raster/RasterBandEditorsTest.java
+++ b/common/src/test/java/org/apache/sedona/common/raster/RasterBandEditorsTest.java
@@ -69,6 +69,19 @@ public class RasterBandEditorsTest extends RasterTestBase {
   }
 
   @Test
+  public void testSetBandNoDataValueWithNullOnSecondBand() throws FactoryException {
+    // The clear must consult the target band's no-data value, not band 1's: band 1 has
+    // none here, which used to short-circuit the removal on band 2.
+    GridCoverage2D raster = RasterConstructors.makeEmptyRaster(2, 20, 20, 0, 0, 8, 8, 0.1, 0.1, 0);
+    GridCoverage2D grid = RasterBandEditors.setBandNoDataValue(raster, 2, 444d);
+    assertEquals(444, (double) RasterBandAccessors.getBandNoDataValue(grid, 2), 0.1d);
+
+    grid = RasterBandEditors.setBandNoDataValue(grid, 2, null);
+    assertNull(RasterBandAccessors.getBandNoDataValue(grid, 2));
+    assertNull(RasterBandAccessors.getBandNoDataValue(grid, 1));
+  }
+
+  @Test
   public void testGetSummaryStats() throws IOException {
     GridCoverage2D raster =
         rasterFromGeoTiff(resourceFolder + "raster/raster_with_no_data/test5.tiff");
--- a/common/src/test/java/org/apache/sedona/common/raster/RasterOutputTest.java
+++ b/common/src/test/java/org/apache/sedona/common/raster/RasterOutputTest.java
@@ -30,6 +30,27 @@ import org.junit.Test;
 public class RasterOutputTest extends RasterTestBase {
 
   @Test
+  public void testAsGeoTiffDoesNotInventNoDataValue() throws FactoryException, IOException {
+    // GeoTiffWriter writes a default GDAL_NODATA of 0 unless told otherwise, which made a
+    // raster that never had a no-data value read back claiming one.
+    GridCoverage2D raster = RasterConstructors.makeEmptyRaster(1, 20, 20, 0, 0, 1, -1, 0, 0, 4326);
+    assertNull(RasterBandAccessors.getBandNoDataValue(raster, 1));
+
+    GridCoverage2D roundTripped = RasterConstructors.fromGeoTiff(RasterOutputs.asGeoTiff(raster));
+    assertNull(RasterBandAccessors.getBandNoDataValue(roundTripped, 1));
+  }
+
+  @Test
+  public void testAsGeoTiffKeepsNoDataValue() throws FactoryException, IOException {
+    // The opt-out must only apply when there is no no-data value to write.
+    GridCoverage2D raster = RasterConstructors.makeEmptyRaster(1, 20, 20, 0, 0, 1, -1, 0, 0, 4326);
+    raster = RasterBandEditors.setBandNoDataValue(raster, 1, -999d);
+
+    GridCoverage2D roundTripped = RasterConstructors.fromGeoTiff(RasterOutputs.asGeoTiff(raster));
+    assertEquals(-999d, RasterBandAccessors.getBandNoDataValue(roundTripped, 1), 0.0001d);
+  }
+
+  @Test
   public void testAsBase64() throws IOException {
     GridCoverage2D raster =
         rasterFromGeoTiff(resourceFolder + "raster/raster_with_no_data/test5.tiff");
--- a/docs/api/sql/Raster-Operators/RS_SetBandNoDataValue.md
+++ b/docs/api/sql/Raster-Operators/RS_SetBandNoDataValue.md
@@ -30,6 +30,8 @@ Since `v1.5.1`, this function supports the ability to replace the current no-dat
 
     To use this for no-data replacement, the input raster must first set its no-data value, which can then be selectively replaced via this function.
 
+    `noDataValue` must be non-null when using the `replace` variant; to remove a no data value, use the 2 or 3 argument form.
+
 Format:
 
 ```
--- a/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/InferredExpression.scala
+++ b/spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/InferredExpression.scala
@@ -468,4 +468,45 @@ object InferrableFunction {
       })
   }
 
+  /**
+   * A variant of ternary inferred expression which allows the third argument to be null. This is
+   * not an overload of [[allowRightNull]] since overloading it breaks eta-expansion of the
+   * overloaded methods passed to it at existing call sites.
+   * @param f
+   *   Function to be wrapped as a catalyst expression.
+   * @param typeTag
+   *   Type tag of the function.
+   * @tparam R
+   *   Return type of the function.
+   * @tparam A1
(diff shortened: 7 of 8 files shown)


spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/raster/RasterBandEditors.scala
```diff
@@ -22,13 +22,15 @@ import org.apache.sedona.common.raster.RasterBandEditors
...
+import org.apache.spark.sql.sedona_sql.expressions.{InferrableFunction, InferredExpression}
 
 private[apache] case class RS_SetBandNoDataValue(inputExpressions: Seq[Expression])
     extends InferredExpression(
       inferrableFunction4(RasterBandEditors.setBandNoDataValue),
-      inferrableFunction3(RasterBandEditors.setBandNoDataValue),
-      inferrableFunction2(RasterBandEditors.setBandNoDataValue)) {
+      // A null noDataValue removes the band's no-data value, so let it through instead of
+      // null-propagating the whole expression.
+      InferrableFunction.allowRightNull3(RasterBandEditors.setBandNoDataValue),
```
jiayuasu, line 32: Could this use the `rasterNoData` value read for `bandIndex`? The null path in the common implementation currently calls `getBandNoDataValue(raster)`, which checks band 1. I think this leaves band 2 unchanged when band 1 has no NoData:

```scala
val df = sparkSession.sql(
  "SELECT RS_MakeEmptyRaster(2, 20, 20, 0, 0, 8, 8, 0.1, 0.1, 0) AS raster")
val result = df
  .selectExpr("RS_SetBandNoDataValue(raster, 2, 444) AS raster")
  .selectExpr("RS_BandNoDataValue(RS_SetBandNoDataValue(raster, 2, NULL), 2)")
  .first().getDouble(0)
// 444.0
```

I expected the last value to be null. An asymmetric multi-band test may be useful here.

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/raster/RasterBandEditors.scala
```diff
@@ -22,13 +22,15 @@ import org.apache.sedona.common.raster.RasterBandEditors
...
+      InferrableFunction.allowRightNull(RasterBandEditors.setBandNoDataValue)) {
```
jiayuasu, line 33: Could we also clear the image-level `GC_NODATA` metadata? On `raster_with_no_data/test5.tiff`, this round trip seems to bring `0` back:

```scala
val result = sparkSession.read.format("binaryFile")
  .load(resourceFolder + "raster/raster_with_no_data/test5.tiff")
  .selectExpr("RS_SetBandNoDataValue(RS_FromGeoTiff(content), NULL) AS raster")
  .selectExpr("RS_AsGeoTiff(raster) AS bytes")
  .selectExpr("RS_BandNoDataValue(RS_FromGeoTiff(bytes)) AS nodata")
  .first().getDouble(0)
// 0.0
```

Before serialization the accessor reports null, but `clone(..., keepMetadata = true)` appears to preserve `GC_NODATA=0`. I think removing or recomputing that property along with the sample dimension would avoid downstream readers treating zero as NoData.

common/src/main/java/org/apache/sedona/common/raster/RasterOutputs.java
```diff
@@ -62,6 +63,19 @@ public static byte[] asGeoTiff(
       throw new RuntimeException(e);
     }
     ParameterValueGroup defaultParams = writer.getFormat().getWriteParameters();
+    // Without this, GeoTiffWriter writes a default GDAL_NODATA of 0 for coverages that
+    // have no no-data value, so a cleared (or never set) no-data value comes back as 0
+    // when the written bytes are read again.
+    boolean hasNoDataValue = false;
+    for (int band = 1; band <= raster.getNumSampleDimensions(); band++) {
```
jiayuasu, line 71: I may be missing a GeoTIFF constraint here, but this still looks lossy for mixed-band metadata. I reproduced it with a two-band GeoTIFF where both bands start with `-999`:

- clear band 2: `(-999, null)` in memory, `(-999, -999)` after `RS_FromGeoTiff(RS_AsGeoTiff(...))`
- clear band 1: `(null, -999)` in memory, `(null, null)` after the same round trip

GeoTools writes one dataset-wide NoData value derived from band 1, so `hasNoDataValue` cannot encode either heterogeneous state. Would it make sense to reject this state or use a lossless representation, and add both cases to the round-trip test?

common/src/main/java/org/apache/sedona/common/raster/RasterBandEditors.java
```diff
@@ -53,14 +55,22 @@ public static GridCoverage2D setBandNoDataValue(
...
-          raster.getRenderedImage(), null, sampleDimensions, raster, null, true);
+      // The GC_NODATA sentinel also rides on the rendered image itself as an image
+      // property and survives serialization there, so cloning the image would let
+      // readers and writers resurrect the cleared value. Copy the pixels into a
+      // property-free image and drop the sentinel from the coverage properties too.
+      Map propertyMap = raster.getProperties();
+      if (propertyMap != null) {
+        propertyMap.remove(NoDataContainer.GC_NODATA);
+      }
+      WritableRaster pixels = raster.getRenderedImage().copyData(null);
```
jiayuasu: Could we keep this on a `RenderedImage` path? I reproduced a clear on a valid coverage with image origin `(5, 7)` and NoData `0`. `copyData(null)` keeps that origin, then the `BufferedImage` constructed in `RasterUtils.clone` throws `IllegalArgumentException: Raster ... has minX or minY not equal to zero: 5 7`. The previous clone path succeeds and preserves the origin. A translated-raster regression test would cover it.

common/src/main/java/org/apache/sedona/common/raster/RasterBandEditors.java
```diff
@@ -53,14 +55,22 @@ public static GridCoverage2D setBandNoDataValue(
...
         return raster;
       }
       GridSampleDimension[] sampleDimensions = raster.getSampleDimensions();
       sampleDimensions[bandIndex - 1] =
           RasterUtils.removeNoDataValue(sampleDimensions[bandIndex - 1]);
-      return RasterUtils.clone(
-          raster.getRenderedImage(), null, sampleDimensions, raster, null, true);
+      // The GC_NODATA sentinel also rides on the rendered image itself as an image
+      // property and survives serialization there, so cloning the image would let
+      // readers and writers resurrect the cleared value. Copy the pixels into a
```
jiayuasu: This may also be expensive for the streaming GeoTIFF path. `RasterConstructors.fromGeoTiff(ImageInputStream)` returns a lazily decoded image, but `copyData(null)` allocates a raster for the full bounds and walks every tile. A 10,000 × 10,000 four-band float raster is about 1.6 GB of copied samples. Could the property be filtered with a delegating `RenderedImage` instead, so clearing metadata does not force pixel materialization?

common/src/main/java/org/apache/sedona/common/raster/RasterOutputs.java
```diff
@@ -62,6 +63,19 @@ public static byte[] asGeoTiff(
...
+      if (RasterBandAccessors.getBandNoDataValue(raster, band) != null) {
+        hasNoDataValue = true;
+        break;
+      }
+    }
+    defaultParams
+        .parameter(GeoTiffFormat.WRITE_NODATA.getName().toString())
+        .setValue(hasNoDataValue);
```
jiayuasu: Do we need to preserve the GeoTools opt-out here? With `-Dgeotiff.writenodata=false`, the descriptor default is false, but this overwrites it with true whenever any band has NoData, so `RS_AsGeoTiff` emits `GDAL_NODATA` despite the explicit setting. Could we only force false for the no-NoData case and otherwise leave the configured default alone?

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/raster/RasterBandEditors.scala
james-willis: fixed. preexisting bug that was latent because this whole case was unreachable before.

common/src/main/java/org/apache/sedona/common/raster/RasterOutputs.java
james-willis: Isn't this a larger question than this PR about a mismatch between serialized and in memory representations of no data values? 

I don't really see how we would fix this in this PR.

common/src/main/java/org/apache/sedona/common/raster/RasterOutputs.java
james-willis: done

common/src/main/java/org/apache/sedona/common/raster/RasterOutputs.java
james-willis: AFAICT theres no standard for heterogenous no data values in Tiffs. We should just reject this state in `RS_AsGeoTiff`. But this is all out of scope for this PR; you can file a separate issue if you'd like

common/src/main/java/org/apache/sedona/common/raster/RasterBandEditors.java
james-willis: done

common/src/main/java/org/apache/sedona/common/raster/RasterBandEditors.java
james-willis: fixed

spark/common/src/main/scala/org/apache/spark/sql/sedona_sql/expressions/raster/RasterBandEditors.scala
james-willis: I cleaned this story up + added a regression test

common/src/main/java/org/apache/sedona/common/utils/PropertyMaskedRenderedImage.java
```diff
@@ -0,0 +1,175 @@
...
+
+  /**
+   * Wrap {@code source} so that {@code propertyName} reads back as {@link Image#UndefinedProperty}.
+   * Returns the source unchanged when it does not carry the property.
+   */
+  public static RenderedImage mask(RenderedImage source, String propertyName) {
+    if (source.getProperty(propertyName) == Image.UndefinedProperty) {
+      return source;
+    }
+    return new PropertyMaskedRenderedImage(source, propertyName);
```
jiayuasu, line 56: Could we preserve `WritableRenderedImage` when the source is writable? Clearing NoData currently makes an editable coverage read-only. I reproduced this through the Java API:

```java
import java.awt.image.WritableRenderedImage;
import javax.media.jai.PlanarImage;
import it.geosolutions.jaiext.range.NoDataContainer;
import org.apache.sedona.common.raster.*;

var raster = RasterConstructors.makeEmptyRaster(1, "d", 4, 3, 0, 0, 1);
raster = RasterBandEditors.setBandNoDataValue(raster, 1, -9999.0);
((PlanarImage) raster.getRenderedImage()).setProperty(
    NoDataContainer.GC_NODATA, new NoDataContainer(-9999.0));

System.out.println(raster.isDataEditable()); // true
var cleared = RasterBandEditors.setBandNoDataValue(raster, 1, null);
System.out.println(cleared.isDataEditable()); // false
((WritableRenderedImage) cleared.getRenderedImage()).getWritableTile(0, 0);
// ClassCastException
```

This affects Java/GeoTools callers using writable tiles; I haven't found a Spark SQL failure from it. A writable subclass selected only for writable sources, forwarding the eight `WritableRenderedImage` methods, looks like a small way to preserve that capability while keeping the lazy wrapper. A test that acquires, writes, and releases a tile after clearing would cover it.

james-willis: Why does this matter?

james-willis: this is moot now anway

james-willis:
I simplified this PR. The issues with bandwise no data values, GC_NODATA persisting after SetNoDataValueCalls, and writing geotiff No Data Values are orthogonal to this PR and thus out of scope.

They are best handled separately.

jiayuasu (approved):

PR #3317, opened 2026-09-04
[GH-3316] Seed the SpiderTests diagonal distribution check

### LINKED ISSUE #3316: SpiderTests diagonal distribution check is flaky without a fixed seed
## Expected behavior

`SpiderTests` should be deterministic and should not fail when the diagonal generator produces a statistically valid random sample.

## Actual behavior

The `generate data with diagonal distribution` test failed in the Spark 4.1.1 / Scala 2.13.17 / JRE 17 CI job for PR #3310:

```text
generate data with diagonal distribution *** FAILED ***
649 was not greater than or equal to 650 (SpiderTests.scala:197)
```

Failed job: https://github.com/apache/sedona/actions/runs/33615369126/job/100547139466

The test requests 1,000 records with `percentage=0.7`, but does not set the Spider `seed`. The data source therefore defaults to `System.currentTimeMillis()`, and `DiagonalGenerator` independently uses a Bernoulli trial for each record. The number of diagonal points is random, so a legitimate sample can fall outside the asserted inclusive range of 650 to 750. In this run it missed the lower bound by one.

This unrelated failure also triggered matrix fail-fast and canceled four other JVM compatibility jobs on PR #3310.

Relevant code:

- Test assertion: https://github.com/apache/sedona/blob/0fdb69d4391da3dd34ebe2d6d0bdd88f985611f9/spark/common/src/test/scala/org/apache/sedona/sql/SpiderTests.scala#L181-L197
- Time-based default seed: https://github.com/apache/sedona/blob/0fdb69d4391da3dd34ebe2d6d0bdd88f985611f9/spark/common/src/main/scala/org/apache/sedona/sql/datasources/spider/SpiderDataSource.scala#L53-L57
- Bernoulli sampling: https://github.com/apache/sedona/blob/0fdb69d4391da3dd34ebe2d6d0bdd88f985611f9/common/src/main/java/org/apache/sedona/common/spider/DiagonalGenerator.java#L67-L79

## Steps to reproduce

Run the focused suite repeatedly without adding a `seed` option:

```shell
mvn -pl spark/common -Dspark=4.1 -Dscala=2.13 -Dspark.version=4.1.1 -DwildcardSuites=org.apache.sedona.sql.SpiderTests test
```

The failure is probabilistic, so individual runs will usually pass.

## Suggested fix

Set an explicit seed in this test and assert the deterministic result for that seed. This preserves coverage of the `percentage` behavior while eliminating random CI outcomes. Widening the statistical bounds alone would still leave a nonzero failure probability.

## Environment

- Sedona version: `2.0.0-SNAPSHOT` (PR #3310 merge commit `4c7585b9cd7eea2d3ae0eead87864fc1a973ed1c`)
- Apache Spark version: `4.1.1`
- Scala version: `2.13.17`
- JRE version: `17`
- API type: Scala
- Environment: GitHub Actions, Ubuntu 22.04

## Existing issues

Eliaaazzz:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #3316

Makes the `generate data with diagonal distribution` test in `SpiderTests` deterministic, as suggested in the issue.

The test asked for 1000 records with `percentage=0.7` but never set the Spider `seed` option, so `SpiderDataSource` fell back to `System.currentTimeMillis()` and the on-diagonal count was a binomial random variable. A legitimate sample could land outside the asserted `[650, 750]` window (a CI run recently produced 649), and the spurious failure also fail-fasted the JVM compatibility matrix.

This PR sets `seed=42` and asserts the exact deterministic count for that seed: 676 of the 1000 points on the diagonal, which keeps the `percentage` behavior covered (676/1000 is consistent with `percentage=0.7`) while removing the randomness. A comment in the test explains why the seed is fixed.

The expected count is fully determined by the seed regardless of Spark/Scala version: `SpiderScanBuilder` gives partition `i` the seed `seed + i`, `SpiderPartitionReader` draws from `java.util.Random(seed + i)`, and `java.util.Random` is specified platform-independently. I cross-checked the value by replicating the partition-range and generator logic (`computePartitionRanges(4, 1000)` + `GeneratorFactory.create("diagonal", ...)`) in a standalone driver against `sedona-common`, which also yields 676 for seed 42.

## How was this patch tested?

Ran the focused suite with the default profile (Spark 3.5.0, Java 11):

```
mvn -pl spark/common test -DwildcardSuites=org.apache.sedona.sql.SpiderTests -Dtest=none
```

All `SpiderTests` cases pass, including the now-deterministic diagonal test. Repeated runs produce the same result by construction since no time-based seed remains.

### DIFF
--- a/spark/common/src/test/scala/org/apache/sedona/sql/SpiderTests.scala
+++ b/spark/common/src/test/scala/org/apache/sedona/sql/SpiderTests.scala
@@ -185,6 +185,7 @@ class SpiderTests extends TestBaseScala with BeforeAndAfterAll {
         .option("numPartitions", "4")
         .option("percentage", "0.7")
         .option("distribution", "diagonal")
+        .option("seed", "42")
         .load()
       assert(spiderDf.rdd.getNumPartitions == 4)
       val records = spiderDf.collect()
@@ -194,7 +195,11 @@ class SpiderTests extends TestBaseScala with BeforeAndAfterAll {
         val geom = row.getAs[Geometry]("geometry").getCentroid
         Math.abs(geom.getX - geom.getY) < 0.000001
       }
-      assert(numPoints >= 650 && numPoints <= 750)
+      // With seed 42, exactly 676 of the 1000 generated points lie on the diagonal,
+      // consistent with percentage=0.7. Without a fixed seed the count is binomial
+      // (drawn with the current time as the seed), so a statistically valid sample
+      // can fall outside any bounded range and fail the test spuriously.
+      assert(numPoints == 676)
     }
 
     it("generate parcel data") {


jiayuasu (approved):
Thank you!

Issue #2392, opened 2025-10-12
chore(geopandas): Add more edge cases to the `test_match_geopandas_series.py` suite


petern48:
`test_match_geopandas_series.py` is where we compare that our results match the results of geopandas. We reuse the same lists of geometries (e.g `self.linestrings`, `self.geoms`, etc) in nearly every test. I wrote these cases very early and failed to consider various edge cases. In this issue, I'd like to add some test cases to these lists (e.g `self.linestrings`). I wouldn't be surprised if we fail some of these.

Here are the edge cases I think we should add. I think each of these should be their own PR, since it's possible that they cause test failures and require us to modify the implementation of existing sedona geopandas functions.
- empty geometries (e.g `POINT EMPTY`)
- non-valid polygons (e.g `POLYGON((0 0, 0 1, 2 1, 2 2, 1 2, 1 0, 0 0))`, see [here](https://postgis.net/workshops/postgis-intro/validity.html) to learn about validity)
- GeometryCollections cases (apparently, there's only [one case](https://github.com/apache/sedona/blob/6dc39e753b9d2087d27d2a6072b95d3eda91c53f/python/tests/geopandas/test_match_geopandas_series.py#L92) at the moment...)
  - geom with only empty geometries
  - geom collection w/ only a single geom (e.g is_closed only applies to linestrings, should it return true for this GeomCol (< a single closed linestring >)?
- Z and M dimensions for everything
- CRS
- Other geometry types that Sedona doesn't support (e.g [LinearRing](https://shapely.readthedocs.io/en/2.1.1/reference/shapely.LinearRing.html#shapely.LinearRing), [box](https://shapely.readthedocs.io/en/2.1.1/manual.html#shapely.geometry.box))

Here's an example of the current cases (and hence where we would need to add more cases).
https://github.com/apache/sedona/blob/6dc39e753b9d2087d27d2a6072b95d3eda91c53f/python/tests/geopandas/test_match_geopandas_series.py#L59-L67

ryux1:
I can take the empty-geometry slice as one focused PR, as requested. I’ll extend the shared GeoSeries fixtures with representative empty geometries, compare behavior against GeoPandas, and include any narrowly required implementation fix if the new parity cases expose one.

ryux1:
[...]

PR #3326, opened 2026-09-06
[GH-2392] Test single-member GeometryCollection

ryux1:

The contributor rules and Python development/test instructions were reviewed while preparing this patch.

Yes. Partially addresses #2392.

Extend the shared GeoSeries parity fixtures with a missing single-member GeometryCollection containing a closed LineString. Because `self.geomcollection` is part of `self.geoms`, the case flows through the existing Sedona-versus-GeoPandas checks, including construction, serialization, predicates, measurements, transformations, and `is_closed`.

The initial revision also included a collection whose members were all typed empty geometries. The full matrix demonstrated a distinct `to_arrow` WKB dimensionality mismatch for that shape, now tracked separately in #3329; it was removed here rather than masking a real defect with an expected failure.

## How was this patch tested?

- Black 26.5.1 formatting check
- Python bytecode compilation
- `git diff --check`
- Reference GeoPandas 1.1+ / Shapely 2.1+ evaluation of the fixture, including `is_empty`, `is_valid`, `is_closed`, `has_z`, `geom_type`, `length`, `area`, and `count_geometries`
- Upstream CI on the initial revision: 2,092 tests passed and the only failure was the separately tracked all-empty-member serialization mismatch in #3329
- Focused distributed checks for the narrowed fixture: constructor plus all four `explode` variants passed (5 tests)
- Broader fail-fast run: 44 passed and 1 skipped before `test_coverage_validation` reached an internal JVM function absent from the released Sedona 1.9.0 shaded JAR used locally

No. This only expands test coverage and does not change a public API.

### DIFF
--- a/python/tests/geopandas/test_match_geopandas_series.py
+++ b/python/tests/geopandas/test_match_geopandas_series.py
@@ -137,10 +137,6 @@ class TestMatchGeopandasSeries(TestGeopandasBase):
 
         self.geomcollection = [
             GeometryCollection(),
-            # A collection can be empty even when it contains typed empty members.
-            GeometryCollection([Point(), LineString(), Polygon()]),
-            # Collection predicates must retain collection semantics with one member.
-            GeometryCollection([LineString([(0, 0), (1, 1), (0, 0)])]),
             GeometryCollection(
                 [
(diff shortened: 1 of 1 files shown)


jiayuasu (approved):
Thank you!

PR #3327, opened 2026-09-06
[GH-3324] Keep other bands' no-data value when clearing one band

### LINKED ISSUE #3324: Clearing NoData on one raster band changes another band's MapAlgebra results
## Expected behavior

Clearing NoData on band 2 should leave band 1's NoData handling unchanged. If band 1 still reports NoData `0`, MapAlgebra should continue treating its zero-valued pixels as NoData.

This is a follow-up to #3312, tracked separately from the SQL NULL-handling fix in #3311.

## Actual behavior

Starting with a two-band GeoTIFF whose bands both use NoData `0`, clearing only band 2 leaves the band metadata at `(0, null)`, but changes band 1's result for `out = rast[0] + 1;` from `NaN` to `1.0`.

That change happens in memory, before writing the cleared raster. The clear path removes the shared `GC_NODATA` property from the coverage and image, while band 1's sample dimension still declares NoData `0`. Jiffle reads the image property, so it starts treating band 1's zeros as ordinary data.

A subsequent GeoTIFF round-trip also changes `(0, null)` back to `(0, 0)`. The writer uses one dataset-wide NoData value, so mixed per-band states cannot be preserved by the current path.

## Steps to reproduce

Run in JShell with `sedona-common` and its dependencies on the classpath. The first GeoTIFF round-trip creates an input with the image-level NoData property used by real GeoTIFFs. `rast[0]` refers to band 1.

```java
import org.apache.sedona.common.raster.*;

var raster = RasterConstructors.makeEmptyRaster(
    2, 20, 20, 0, 0, 1, -1, 0, 0, 4326);
raster = RasterBandEditors.setBandNoDataValue(raster, 1, 0.0);
raster = RasterBandEditors.setBandNoDataValue(raster, 2, 0.0);
var loaded = RasterConstructors.fromGeoTiff(RasterOutputs.asGeoTiff(raster));

var before = MapAlgebra.mapAlgebra(loaded, "d", "out = rast[0] + 1;");
System.out.println(before.getRenderedImage().getData().getSampleDouble(0, 0, 0));
// NaN

var cleared = RasterBandEditors.setBandNoDataValue(loaded, 2, null);
System.out.println(RasterBandAccessors.getBandNoDataValue(cleared, 1));
// 0.0
System.out.println(RasterBandAccessors.getBandNoDataValue(cleared, 2));
// null

var after = MapAlgebra.mapAlgebra(cleared, "d", "out = rast[0] + 1;");
System.out.println(after.getRenderedImage().getData().getSampleDouble(0, 0, 0));
// 1.0 (expected NaN: band 1 was not cleared)

var roundTrip = RasterConstructors.fromGeoTiff(RasterOutputs.asGeoTiff(cleared));
System.out.println(RasterBandAccessors.getBandNoDataValue(roundTrip, 2));
// 0.0 (band 2's cleared NoData is restored)
```

The regression coverage should check both bands after clearing either one, including MapAlgebra before serialization. For GeoTIFF output, rejecting unsupported mixed NoData states would be preferable to silently changing them.

## Sedona version

`2.0.0-SNAPSHOT`, reproduced on #3312 at `2b06236af00c0789c9e0eff69d9562caf6e90209`.

## API type

Java (`sedona-common`; used by the SQL raster functions).

## JRE version

17, GeoTools 33.1.

## Environment

Local macOS. The reproducer does not require Spark.

## Existing issues

Eliaaazzz:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #3324

> **Stacked on #3312.** #3324 is a follow-up to that PR and only reproduces with it applied, so this branch starts from `2b06236` (its current head) and adds two commits on top. The first three commits in the diff belong to #3312. Please merge #3312 first and I will rebase this onto `master`, or fold this into #3312 if that is easier to review.

Two independent commits, so the second can be dropped without losing the first.

### 1. Keep other bands' no-data value when clearing one band (`309b16a`)

No-data values are tracked per band on the sample dimensions. `GC_NODATA` is a single sentinel for the whole coverage, carried both by the coverage properties and by the rendered image. The clear path dropped that sentinel unconditionally, so clearing one band also stopped GeoTools operations that read the sentinel from honouring the bands that still declared a no-data value. Jiffle reads it, which is how clearing band 2 changed band 1's `out = rast[0] + 1;` from `NaN` to `1.0`.

The fix drops the sentinel only once no band declares a no-data value at all. Reproduced before the change and verified after, on the exact steps in the issue:

- Loaded from GeoTIFF: band 1 `0.0`, band 2 `0.0`, `rast[0] + 1` gives `NaN`.
- After clearing band 2, before this PR: band 1 `0.0`, band 2 `null`, `rast[0] + 1` gives `1.0`. This is the bug.
- After clearing band 2, with this PR: band 1 `0.0`, band 2 `null`, `rast[0] + 1` gives `NaN`.
- After clearing band 1 as well: band 1 `null`, band 2 `null`, `rast[0] + 1` gives `1.0`.

### 2. Reject mixed per-band no-data values when writing GeoTiff (`9e5525c`)

This is the second half of the issue: a `(0, null)` raster came back from a GeoTIFF round trip as `(0, 0)`. GDAL_NODATA is one value for the whole file, so that state cannot be written faithfully. `RS_AsGeoTiff` now throws `IllegalArgumentException` on it, rather than resurrecting a no-data value on a band that had it cleared.

**This commit needs a maintainer decision, because its blast radius is wider than the clear path.** `RasterUtils.copyRasterAndAppendBand` sets the appended band's no-data value and leaves the existing bands' sample dimensions alone, so `RS_AddBand` and `RS_Union` routinely produce rasters whose bands disagree. Writing one of those to GeoTIFF silently rewrites the other bands' no-data today; with this commit it fails loudly. That follows the issue's preference for rejecting unsupported mixed states, but if it reads as too strong for `RS_AsGeoTiff`, dropping `9e5525c` leaves commit 1 intact and I will open a separate issue for the writer.

## How was this patch tested?

New tests in `common`:

- `RasterBandEditorsTest.testSetBandNoDataValueWithNullKeepsNoDataOnOtherBands` covers the issue's reproducer. It checks both bands and map algebra after clearing either one, before serialization, and that the input raster is not mutated.
- `RasterBandEditorsTest.testSetBandNoDataValueWithNullOnEveryBandClearsNoDataProperty` covers the case where every band is cleared: the sentinel is dropped from the coverage and the image, map algebra reads the zeros as ordinary data, and the state survives a GeoTIFF round trip.
- `RasterOutputTest.testAsGeoTiffRejectsBandsWithDifferentNoDataValues` covers "only one band set" and "both set but different". Rasters whose bands agree, and rasters with no no-data value at all, still round-trip.

`mvn -pl common test` was run before and after the change on Windows with JDK 17. The same 13 tests fail both times: `RasterBandEditorsTest.testClip`, 5 `RasterEditorsTest.testResample*`, and 7 `RasterOutputTest.testAsMatrix*`. They fail identically on unmodified `2b06236`, so they are pre-existing on this platform and unrelated to this change. Everything else passes, including the three new tests.

I could not run `spark/common` locally. It fails to compile on the generated OSM PBF protobuf sources (`package proto4 does not exist`) before reaching any raster code, so those suites are left to CI.

One note for whoever reviews the tests. `MapAlgebra` caches a compiled Jiffle runtime per script text in a `ThreadLocal`, so I checked whether a cached runtime could mask this regression by carrying a previous raster's no-data handling into a later call with the same script. It does not: `setSourceImage` re-reads the property. The tests therefore use a plain shared script.

- Yes, I have updated the documentation. `RS_SetBandNoDataValue` now states that no-data is tracked per band, and `RS_AsGeoTiff` documents the single-value-per-file constraint and the new rejection.

https://claude.ai/code/session_01P6hpqAnXfeaWRLkL1VQDMd

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/raster/RasterBandEditors.java
+++ b/common/src/main/java/org/apache/sedona/common/raster/RasterBandEditors.java
@@ -18,11 +18,13 @@
  */
 package org.apache.sedona.common.raster;
 
+import it.geosolutions.jaiext.range.NoDataContainer;
 import java.awt.geom.Point2D;
 import java.awt.image.Raster;
 import java.awt.image.RenderedImage;
 import java.awt.image.WritableRaster;
 import java.util.Collections;
+import java.util.Map;
 import javax.media.jai.RasterFactory;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.tuple.Pair;
@@ -53,14 +55,22 @@ public class RasterBandEditors {
 
     // Remove no-Data if it is null
     if (noDataValue == null) {
-      if (RasterBandAccessors.getBandNoDataValue(raster) == null) {
+      if (rasterNoData == null) {
         return raster;
       }
       GridSampleDimension[] sampleDimensions = raster.getSampleDimensions();
       sampleDimensions[bandIndex - 1] =
           RasterUtils.removeNoDataValue(sampleDimensions[bandIndex - 1]);
-      return RasterUtils.clone(
-          raster.getRenderedImage(), null, sampleDimensions, raster, null, true);
+      // The GC_NODATA sentinel also rides on the rendered image itself as an image
+      // property and survives serialization there, so cloning the image would let
+      // readers and writers resurrect the cleared value. Copy the pixels into a
+      // property-free image and drop the sentinel from the coverage properties too.
+      Map propertyMap = raster.getProperties();
+      if (propertyMap != null) {
+        propertyMap.remove(NoDataContainer.GC_NODATA);
+      }
+      WritableRaster pixels = raster.getRenderedImage().copyData(null);
+      return RasterUtils.clone(pixels, null, sampleDimensions, raster, null, true, propertyMap);
     }
 
     if (rasterNoData != null && rasterNoData.equals(noDataValue)) {
--- a/common/src/main/java/org/apache/sedona/common/raster/RasterBandEditors.java
+++ b/common/src/main/java/org/apache/sedona/common/raster/RasterBandEditors.java
@@ -24,10 +24,12 @@ import java.awt.image.Raster;
 import java.awt.image.RenderedImage;
 import java.awt.image.WritableRaster;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import javax.media.jai.RasterFactory;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.tuple.Pair;
+import org.apache.sedona.common.utils.PropertyMaskedRenderedImage;
 import org.apache.sedona.common.utils.RasterUtils;
 import org.geotools.api.parameter.ParameterValueGroup;
 import org.geotools.api.referencing.FactoryException;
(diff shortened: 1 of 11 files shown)


jiayuasu:
please do not take this kind of issues. this is not for new contributors.

PR #3328, opened 2026-09-06
[GH-3255] Clip line segments to the raster in pixel space

### LINKED ISSUE #3255: RS_AsRaster segment clipping drops crossings and loses grid-line sidedness
## Expected behavior

LineString segments, and polygon-boundary segments when `allTouched = true`, should be rasterized whenever their intersection with the raster window has positive length, regardless of whether either original endpoint is inside the window. Reversing a segment should produce the same band.

Segments that are disjoint from the window, or that only touch it at one corner point, should burn no cells. A segment clipped to a window edge should retain the exact side of nearby internal pixel grid lines.

## Actual behavior

`Rasterization.clipSegmentToRasterBounds` returns `null` whenever both endpoints are outside the raster extent. This drops valid segments that cross the raster from one side to another.

The current one-endpoint clipping path also stores an interpolated endpoint as ordinary `double` coordinates. If the unconstrained coordinate rounds onto an internal grid line, rasterization can select the wrong adjacent cell even though the segment remains strictly on one side of that line inside the raster.

## Steps to reproduce

On a 6 x 6 north-up unit raster, this horizontal line crosses all six columns but both endpoints are outside:

```sql
SELECT RS_BandAsArray(
  RS_AsRaster(
    ST_GeomFromWKT('LINESTRING (-1 2.5, 7 2.5)'),
    'B', true, 1.0, 0.0, false),
  1);
```

Expected: all six cells in zero-based row 3 are `1`.

Actual: every cell is `0`.

A secondary clipping case demonstrates loss of endpoint sidedness:

```sql
SELECT RS_BandAsArray(
  RS_AsRaster(
    ST_GeomFromWKT('LINESTRING (-1 3.0000000000000004, 1 3.0)'),
    'B', true, 1.0, 0.0, false),
  1);
```

Within the raster, the nondegenerate part of this line is strictly above `y = 3`, except at its final endpoint. It should therefore burn `(row 2, column 0)`. The clipped coordinate rounds to `y = 3`, and Sedona instead burns `(row 3, column 0)`. Reversing the input must preserve the correct result.

## Root cause

The clipper has three related numerical/topological limitations:

1. It treats "both endpoints outside" as equivalent to "no intersection."
2. It returns only rounded coordinates, losing whether an interpolated coordinate is exactly on, just below, or just above an internal grid line.
3. Once both-outside crossings are admitted, direct differences such as `x2 - x1` can overflow for valid finite endpoints such as `-1e308` and `1e308`.

## Acceptance criteria

- A nondegenerate segment with a positive-length window intersection is rasterized even when both source endpoints are outside.
- Disjoint and parallel-outside segments burn nothing.
- Corner-only tangencies burn nothing; boundary-coincident positive-length segments retain the established half-open side.
- Clipped endpoints preserve the side of internal grid lines, including exact, `Math.nextDown`, and `Math.nextUp` cases on both axes.
- Forward and reversed segments produce identical bands.
- An originally degenerate in-bounds LineString is distinguished from a nondegenerate segment collapsed to a tangent point by clipping.
- Clipping valid finite endpoints does not produce NaN-driven omissions or traversal streaks. Include horizontal and vertical `+/-1e308` crossings, or explicitly validate and document a smaller supported coordinate range.
- Tests use an independent exact-rational segment/window oracle. The oracle must not exclude intersections near grid lines.
- The replacement adds no arbitrary-precision or per-segment helper allocations on the ordinary finite-coordinate path; any exact fallback is limited to unresolved or overflow-prone cases.

## Related work

- #3118 replaced fixed-step sampling with exact cell traversal, but did not change segment clipping.
- #3120 and #3251 address endpoint-cell selection after coordinates are already in pixel space; #3251 deliberately leaves clipping unchanged.

## Settings

- Sedona version: PR #3251 head `72d9cc26541` / `2.0.0-SNAPSHOT` (the both-outside crossing omission also reproduces on current `master`)
- API type: SQL and Java
- JRE: 17
- Environment: local development checkout

I searched open and closed issues and did not find a duplicate.

Eliaaazzz:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #3255

`Rasterization.clipSegmentToRasterBounds` is replaced by a clip that runs in pixel space. Both reproducers from the issue now produce the expected band, and the four other numerical points in the acceptance criteria are covered.

### Both endpoints outside

The old clipper returned `null` as soon as neither endpoint was inside the extent, which dropped every segment that crosses the raster from one side to the other. On the issue's 6x6 unit raster, `LINESTRING (-1 2.5, 7 2.5)` left the band empty; it now burns all six cells of row 3. The parametric clip below that early return was already able to handle the case, it was simply never reached.

### Grid-line sidedness

Clipping moved to pixel space, where the window edges are the integers `0`, `width` and `height`. In world space the edges are grid lines that only round-trip approximately through `(coordinate - upperLeft) / scale`, so a clipped endpoint can land a rounding step off the line it belongs to.

The coordinate pinned to a window edge is now exact by construction, and only the interpolated one can round onto a grid line. When it does, its true side is resolved with `CGAlgorithmsDD.orientationIndex` against the lattice point and the coordinate steps one ulp off the line. Solving the clip for the interpolated coordinate leaves exactly the cross product that predicate returns, divided by the delta along the clipped axis, so the sign follows directly.

`LINESTRING (-1 3.0000000000000004, 1 3.0)` is strictly above `y = 3` inside the raster except at its final endpoint. Its clipped endpoint rounded onto `y = 3` and burned row 3; it now burns row 2. The mirrored case below the line still burns row 3, and the same rounding on the x axis is covered too.

### Direction independence

Segment endpoints are ordered before clipping. The two directions reach a clip edge at complementary parameters that are equal in exact arithmetic but round apart as doubles, and interpolating with them puts a coordinate on either side of an exact grid corner. My own oracle test caught this on `LINESTRING (-2 0.5, 8 5.5)`, which passes exactly through the pixel lattice point (5, 2).

### Overflow and collapsed parameters

A difference that would overflow to infinity for two large opposite-sign endpoints is taken on halved coordinates, which is exact for normal doubles, so the parameters are the ones the direct form would have produced. That path is entered only when the direct difference is infinite.

Whether the clipped piece has length is read from the clipped endpoints rather than from the parameters. The parameters answer it correctly for a single-point touch such as a corner, but they collapse together for a genuine crossing whose endpoints are far enough apart that the whole window falls inside one rounding step. `LINESTRING (-1E308 2.5, 1E308 2.5)` burns row 3 rather than nothing, in both directions and on both axes.

### Degenerate versus collapsed

A segment that was degenerate to begin with still marks the cell holding it. A nondegenerate segment that clipping collapses to a tangent point burns nothing: `LINESTRING (5 7, 7 5)` meets the raster only at the corner (6, 6).

### Allocation

The ordinary finite path is plain doubles with no per-segment allocation. The single scratch array for clipped endpoints is allocated once per geometry, not per segment. `CGAlgorithmsDD.orientationIndex` is reached only by an interpolated coordinate that landed exactly on a grid line, and it is already the predicate the traversal uses.

## How was this patch tested?

Six new tests in `RasterizationTest`:

- `testLineCrossingRasterWithBothEndpointsOutside` covers the issue's first reproducer plus vertical and diagonal crossings, each in both vertex orders.
- `testClippedEndpointKeepsGridLineSide` covers the second reproducer, its mirror below the grid line, and the same rounding on the x axis.
- `testLineTouchingRasterAtASinglePointBurnsNothing` covers the corner tangency.
- `testLineWithExtremeCoordinatesCrossesTheRaster` covers horizontal and vertical `+/-1e308` crossings.
- `testDegenerateLineStringBurnsItsOwnCell` covers the degenerate input.
- `testLineClipAgreesWithExactRationalOracle` cross-checks fifteen segments against `SegmentClipOracle`, a new test-only exact-rational clip of the same segment against the same window. It runs entirely in `BigInteger` rationals with no rounding anywhere, so a segment grazing a grid line is resolved rather than excluded, and it shares no arithmetic with the production clipper. For each segment it asserts that the band is empty exactly when the exact intersection has no length, that the cells holding the exact clipped endpoints are burned, and that reversing the vertices produces an identical band.

That oracle test is what caught the direction dependence and the collapsed-parameter omission described above, both of which were in my first draft of the clipper.

`mvn -pl common test` on Windows with JDK 17: 1350 tests, the same 13 failures as on unmodified `master` (`RasterBandEditorsTest.testClip`, 5 `RasterEditorsTest.testResample*`, 7 `RasterOutputTest.testAsMatrix*`). They fail identically before and after, so they are pre-existing on this platform and unrelated. Everything else passes, including the GH-3118 coastline fixtures with their pinned 1738/1842 counts and the GH-3120 and GH-3251 endpoint-cell tests.

I could not run `spark/common` locally: it fails to compile on the generated OSM PBF protobuf sources (`package proto4 does not exist`) before reaching any raster code. Leaving those suites to CI.

One note on scope. The clip window is now the raster's own pixel bounds rather than the snapped `geomExtent`. The two give the same burned cells, because a segment lies inside its geometry's snapped extent and `burnCell` already discards anything outside the raster, and the traversal step bound `width + height + 2` assumed the raster bounds anyway. It also means the window edges are exact integers, which is what makes the sidedness argument hold. `rasterizeLineString` no longer takes `geomExtent`.

https://claude.ai/code/session_01P6hpqAnXfeaWRLkL1VQDMd

### DIFF
--- a/common/src/main/java/org/apache/sedona/common/raster/Rasterization.java
+++ b/common/src/main/java/org/apache/sedona/common/raster/Rasterization.java
@@ -153,7 +153,7 @@ public class Rasterization {
       case "LineString":
       case "MultiLineString":
       case "LinearRing":
-        rasterizeLineString(geom, params, value, geomExtent);
+        rasterizeLineString(geom, params, value);
         break;
       default:
         rasterizePolygon(geom, params, geomExtent, value, allTouched);
@@ -216,32 +216,50 @@ public class Rasterization {
     }
   }
 
-  private static void rasterizeLineString(
-      Geometry geom, RasterizationParams params, double value, ReferencedEnvelope geomExtent) {
+  private static void rasterizeLineString(Geometry geom, RasterizationParams params, double value) {
+    int width = params.writableRaster.getWidth();
+    int height = params.writableRaster.getHeight();
+    double[] clipped = new double[4];
+
     for (int i = 0; i < geom.getNumGeometries(); i++) {
       LineString line = (LineString) geom.getGeometryN(i);
       Coordinate[] coords = line.getCoordinates();
 
       for (int j = 0; j < coords.length - 1; j++) {
-        // Extract start and end points for the segment
-        LineSegment clippedSegment =
-            clipSegmentToRasterBounds(coords[j], coords[j + 1], geomExtent);
+        // Clip in pixel space, where the window edges are the integers 0, width and height. In
+        // world space the edges are grid lines that only round-trip approximately, and an
+        // interpolated endpoint landing a rounding step off a grid line selects the wrong cell.
+        double x0 = (coords[j].x - params.upperLeftX) / params.scaleX;
+        double y0 = (coords[j].y - params.upperLeftY) / params.scaleY;
+        double x1 = (coords[j + 1].x - params.upperLeftX) / params.scaleX;
+        double y1 = (coords[j + 1].y - params.upperLeftY) / params.scaleY;
 
-        Coordinate start;
-        Coordinate end;
-        if (clippedSegment != null) {
-          start = new Coordinate(clippedSegment.p0.x, clippedSegment.p0.y);
-          end = new Coordinate(clippedSegment.p1.x, clippedSegment.p1.y);
-        } else {
-          continue; // Skip case where segment is completely outside geomExtent
+        if (x0 == x1 && y0 == y1) {
+          // A segment that was degenerate to begin with still marks the cell holding it. This is
+          // distinct from a nondegenerate segment that clipping collapses to a tangent point,
+          // which the clipper rejects below.
+          if (x0 >= 0 && x0 <= width && y0 >= 0 && y0 <= height) {
+            traverseSegment(params, x0, y0, x1, y1, value);
+          }
+          continue;
         }
 
-        double x0 = (start.x - params.upperLeftX) / params.scaleX;
-        double y0 = (start.y - params.upperLeftY) / params.scaleY;
-        double x1 = (end.x - params.upperLeftX) / params.scaleX;
-        double y1 = (end.y - params.upperLeftY) / params.scaleY;
+        // Order the endpoints before clipping, so the cells burned cannot depend on which vertex
+        // the line visits first. The two directions reach a clip edge at complementary parameters
+        // that are equal in exact arithmetic but round apart as doubles, and interpolating with
+        // them lands a coordinate on either side of an exact grid corner.
+        if (x1 < x0 || (x1 == x0 && y1 < y0)) {
+          double swap = x0;
+          x0 = x1;
+          x1 = swap;
+          swap = y0;
+          y0 = y1;
+          y1 = swap;
+        }
 
-        traverseSegment(params, x0, y0, x1, y1, value);
+        if (clipSegmentToRaster(x0, y0, x1, y1, width, height, clipped)) {
+          traverseSegment(params, clipped[0], clipped[1], clipped[2], clipped[3], value);
+        }
       }
     }
   }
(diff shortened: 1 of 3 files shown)


jiayuasu:
please do not take this kind of issues. this is not for new contributors.

Please work on `good first issues`

PR #3334, opened 2026-09-07
[GH-1024] Add Sedona R instructions to the AWS EMR setup page

### LINKED ISSUE #1024: Install instructions Sedona/R for AWS EMR?
Hi do you have instructions for installing Sedona for R use on AWS EMR?

minhpham1810:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #1024

Adds a `## Use Sedona in R` section to `docs/setup/emr.md` and its Chinese mirror `docs/setup/emr.zh.md`.

There is currently no R deployment guidance anywhere in the mkdocs site — both "Install Sedona R" nav entries point at the generated pkgdown site, which documents the R API but not how to get Sedona onto a cluster. The EMR page covers Scala and Python and stops there.

The new section covers three things:

- extending the page's existing bootstrap script to install R, `sparklyr` and `apache.sedona`;
- connecting with `spark_connect(master = "yarn", spark_home = "/usr/lib/spark")`, pointing `SEDONA_JAR_FILES` at the two jars the bootstrap script already downloads into `/jars`;
- a one-liner to verify the R side.

The Sedona-specific claims are grounded in `R/R/dependencies.R`: `apache.sedona` registers itself as a `sparklyr` extension whose initializer calls `SedonaContext.create(spark_session(sc), "r")`, so there is no manual registration step in R; `SEDONA_JAR_FILES` is split on `:` and, when set, replaces **both** Maven coordinates, which is why the `geotools-wrapper` jar has to be listed alongside the Sedona jar; and the R interface rejects Spark 4.x outright, so this is Spark 3.x only.

## Two open questions for reviewers

I don't have an EMR cluster to validate against, and these two would change the text:

1. **Which EMR release should this target?** The page still documents EMR 6.9.0 / Spark 3.3.0 with `sedona-spark-shaded-3.3_2.12`. EMR 7.x ships Spark 3.5. I wrote the R section against the page's existing baseline rather than silently changing it, but the whole page looks due a refresh — happy to do that here or in a follow-up, whichever you prefer.

2. **Is `SEDONA_JAR_FILES` actually necessary here?** The page's `spark-defaults` classification already sets `spark.yarn.dist.jars` to the same two `/jars` paths. If `spark_connect()` inherits that, the `Sys.setenv()` block is redundant and should be presented as a no-internet fallback rather than as the default path.

Three further assumptions on the AWS side are flagged in https://github.com/apache/sedona/issues/1024#issuecomment-5576489007 — whether the EMR AMI already ships R, whether core/task nodes need the R packages in YARN client mode, and how R is expected to be driven on EMR now that Studio has no R kernel and the RStudio integration is deprecated.

## How was this patch tested?

- `prek run --files docs/setup/emr.md docs/setup/emr.zh.md` — all hooks pass (markdownlint, codespell, editorconfig-checker, license, whitespace).
- `mkdocs build` — succeeds; the `{{ sedona.current_version }}` / `{{ sedona.current_geotools }}` macros resolve to `1.9.1` and `1.9.1-33.5` in both locales, the three new headings anchor, and both new `!!!note` admonitions render.
- External links checked for lychee: all resolve 200 after redirects.
- The R snippets themselves were not executed against a live EMR cluster — see the open questions above.

https://claude.ai/code/session_01LGHuMH6Jk2ForsfJPEvev6

### DIFF
--- a/docs/setup/emr.md
+++ b/docs/setup/emr.md
@@ -78,3 +78,50 @@ spark.sql("SELECT ST_Point(0, 0)").show()
 ```
 
 Note that: you don't need to run the `SedonaRegistrator.registerAll(spark)` or `SedonaContext.create(spark)` because `org.apache.sedona.sql.SedonaSqlExtensions` in the config will take care of that.
+
+## Use Sedona in R
+
+The [`apache.sedona`](https://cran.r-project.org/package=apache.sedona) R package is a [`sparklyr`](https://spark.rstudio.com) extension. Attaching it before `spark_connect()` is enough to register Sedona's serializers, UDTs and UDFs, so there is no R equivalent of `SedonaContext.create()` to call by hand.
+
+!!!note
+	The R interface supports Spark 3.x only. Make sure the EMR release you pick ships a Spark 3 version.
+
+### Extend the initialization script
+
+Add the following to the bootstrap script above, so that R and the two R packages are available on the node you run R from:
+
+```bash
+# Install R and the Sedona R interface
+sudo yum install -y R
+sudo R -e 'install.packages(c("sparklyr", "apache.sedona"), repos = "https://cloud.r-project.org")'
+```
+
+### Connect to the cluster from R
+
+EMR installs Spark under `/usr/lib/spark`. Point `SEDONA_JAR_FILES` at the jars the bootstrap script already downloaded into `/jars` so that `sparklyr` uses them instead of resolving the Sedona coordinates from Maven Central every time you connect:
+
+```r
+library(sparklyr)
+library(apache.sedona)
+
+Sys.setenv(
+  "SEDONA_JAR_FILES" = paste(
+    "/jars/sedona-spark-shaded-3.3_2.12-{{ sedona.current_version }}.jar",
+    "/jars/geotools-wrapper-{{ sedona.current_geotools }}.jar",
+    sep = ":"
+  )
+)
+
+sc <- spark_connect(master = "yarn", spark_home = "/usr/lib/spark")
+```
+
+!!!note
+	`SEDONA_JAR_FILES` holds a `:`-separated list and replaces *both* Maven coordinates that `apache.sedona` would otherwise request, which is why the GeoTools wrapper jar has to be listed next to the Sedona jar. If you leave `SEDONA_JAR_FILES` unset, every connection downloads `org.apache.sedona:sedona-spark-shaded-<spark version>_<scala version>:{{ sedona.current_version }}` and `org.datasyslab:geotools-wrapper:{{ sedona.current_geotools }}`. That requires outbound internet access from the driver and can take long enough to exceed the default `sparklyr.connect.timeout`.
+
+### Verify the R installation
+
+```r
+sdf_sql(sc, "SELECT ST_Point(0.0, 0.0) AS geom") %>% collect()
+```
+
+For more on what the R interface offers, see the [Sedona R documentation](https://sedona.apache.org/latest/api/rdocs/).
--- a/docs/setup/emr.md
+++ b/docs/setup/emr.md
@@ -84,21 +84,23 @@ Note that: you don't need to run the `SedonaRegistrator.registerAll(spark)` or `
 The [`apache.sedona`](https://cran.r-project.org/package=apache.sedona) R package is a [`sparklyr`](https://spark.rstudio.com) extension. Attaching it before `spark_connect()` is enough to register Sedona's serializers, UDTs and UDFs, so there is no R equivalent of `SedonaContext.create()` to call by hand.
 
 !!!note
-	The R interface supports Spark 3.x only. Make sure the EMR release you pick ships a Spark 3 version.
+	The `apache.sedona` 1.9.1 release on CRAN supports Spark 3.x only. Every EMR 7.x release ships Spark 3.5, so any of them works.
 
 ### Extend the initialization script
 
-Add the following to the bootstrap script above, so that R and the two R packages are available on the node you run R from:
+Add the following to the bootstrap script above. Like the rest of the script, it runs on every node in the cluster. Spark SQL queries from R only need the R packages on the node you run R from; worker nodes need them only if you run R code on executors with `spark_apply()`.
 
 ```bash
-# Install R and the Sedona R interface
-sudo yum install -y R
-sudo R -e 'install.packages(c("sparklyr", "apache.sedona"), repos = "https://cloud.r-project.org")'
+# Install R and the Sedona R interface. libcurl-devel is needed to build the R curl package that sparklyr depends on.
+sudo yum install -y R libcurl-devel
+sudo R -e 'install.packages(c("sparklyr", "apache.sedona"), repos = "https://cloud.r-project.org", Ncpus = parallel::detectCores()); stopifnot(all(c("sparklyr", "apache.sedona") %in% rownames(installed.packages())))'
 ```
+
+`install.packages()` only warns when a package fails to build, so the `stopifnot()` call is what makes the `R` command fail instead of reporting success. Expect the R installation to add several minutes to the bootstrap of each node.
 
 ### Connect to the cluster from R
 
-EMR installs Spark under `/usr/lib/spark`. Point `SEDONA_JAR_FILES` at the jars the bootstrap script already downloaded into `/jars` so that `sparklyr` uses them instead of resolving the Sedona coordinates from Maven Central every time you connect:
+EMR installs Spark under `/usr/lib/spark`. When you connect from R, the Spark driver runs in YARN client mode on the node where R runs. The `spark.yarn.dist.jars` setting above only ships jars to the executors, so it does not put Sedona on the driver's classpath. `apache.sedona` adds the jars to the driver itself, either by downloading them from Maven Central or from the local files listed in `SEDONA_JAR_FILES`. Point `SEDONA_JAR_FILES` at the jars the bootstrap script already downloaded into `/jars`:
 
 ```r
 library(sparklyr)
@@ -106,7 +108,7 @@ library(apache.sedona)
 
 Sys.setenv(
   "SEDONA_JAR_FILES" = paste(
-    "/jars/sedona-spark-shaded-3.3_2.12-{{ sedona.current_version }}.jar",
+    "/jars/sedona-spark-shaded-3.5_2.12-{{ sedona.current_version }}.jar",
     "/jars/geotools-wrapper-{{ sedona.current_geotools }}.jar",
     sep = ":"
   )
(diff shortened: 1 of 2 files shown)


docs/setup/emr.md
```diff
@@ -78,3 +78,50 @@ spark.sql("SELECT ST_Point(0, 0)").show()
...
+	The R interface supports Spark 3.x only. Make sure the EMR release you pick ships a Spark 3 version.
+
+### Extend the initialization script
+
+Add the following to the bootstrap script above, so that R and the two R packages are available on the node you run R from:
+
+```bash
+# Install R and the Sedona R interface
+sudo yum install -y R
+sudo R -e 'install.packages(c("sparklyr", "apache.sedona"), repos = "https://cloud.r-project.org")'
```
jiayuasu: This doesn't work on EMR 7.x as written. The R `curl` package fails to configure because the libcurl headers aren't on the image, and that takes httr, sparklyr and apache.sedona down with it:

```
Configuration failed because libcurl was not found. Try installing:
 * rpm: libcurl-devel (Fedora, CentOS, RHEL)
ERROR: dependency 'curl' is not available for package 'httr'
ERROR: dependency 'httr' is not available for package 'sparklyr'
ERROR: dependency 'sparklyr' is not available for package 'apache.sedona'
```

`R -e` still exits 0, so the bootstrap reports success and you only find out at `library(sparklyr)`.

```bash
sudo yum install -y R libcurl-devel
```

is enough. I re-ran with that and all 38 packages built on both nodes (about 1.5 min for yum and 6.5 min for `install.packages` per node). openssl and xml2 are fine already because R-core-devel brings their headers.

Since this runs on every node, it may also be worth making the install fail loudly and use all cores:

```r
install.packages(c("sparklyr", "apache.sedona"), repos = "https://cloud.r-project.org", Ncpus = parallel::detectCores())
stopifnot(all(c("sparklyr", "apache.sedona") %in% rownames(installed.packages())))
```

docs/setup/emr.md
```diff
@@ -78,3 +78,50 @@ spark.sql("SELECT ST_Point(0, 0)").show()
...
+
+EMR installs Spark under `/usr/lib/spark`. Point `SEDONA_JAR_FILES` at the jars the bootstrap script already downloaded into `/jars` so that `sparklyr` uses them instead of resolving the Sedona coordinates from Maven Central every time you connect:
+
+```r
+library(sparklyr)
+library(apache.sedona)
+
+Sys.setenv(
+  "SEDONA_JAR_FILES" = paste(
+    "/jars/sedona-spark-shaded-3.3_2.12-{{ sedona.current_version }}.jar",
```
jiayuasu: This is the Spark 3.3 jar, but since #3300 the bootstrap on master downloads `sedona-spark-shaded-3.5_2.12`. The hunk merges cleanly so nothing flags it, and `spark_connect()` dies with `FileNotFoundException: File file:/jars/sedona-spark-shaded-3.3_2.12-1.9.1.jar does not exist`. Rebase and change this to `3.5_2.12`.

docs/setup/emr.md
```diff
@@ -78,3 +78,50 @@ spark.sql("SELECT ST_Point(0, 0)").show()
...
+    "/jars/geotools-wrapper-{{ sedona.current_geotools }}.jar",
+    sep = ":"
+  )
+)
+
+sc <- spark_connect(master = "yarn", spark_home = "/usr/lib/spark")
+```
+
+!!!note
+	`SEDONA_JAR_FILES` holds a `:`-separated list and replaces *both* Maven coordinates that `apache.sedona` would otherwise request, which is why the GeoTools wrapper jar has to be listed next to the Sedona jar. If you leave `SEDONA_JAR_FILES` unset, every connection downloads `org.apache.sedona:sedona-spark-shaded-<spark version>_<scala version>:{{ sedona.current_version }}` and `org.datasyslab:geotools-wrapper:{{ sedona.current_geotools }}`. That requires outbound internet access from the driver and can take long enough to exceed the default `sparklyr.connect.timeout`.
```
jiayuasu: This isn't what the code does. In `dependencies.R` the geotools-wrapper coordinate goes into `packages` unconditionally; `SEDONA_JAR_FILES` only drops the Sedona coordinate. With the env var set exactly like the snippet above, spark-submit was launched with

```
--jars /jars/sedona-spark-shaded-3.5_2.12-1.9.1.jar,/jars/geotools-wrapper-1.9.1-33.5.jar
--packages org.datasyslab:geotools-wrapper:1.9.1-33.5
```

and `~/.ivy2/jars` went from empty to containing the geotools-wrapper jar. So the driver still needs Maven Central (or a warm Ivy cache), and listing the geotools jar is harmless but not for the reason given here. Please reword this, and the "instead of resolving ... from Maven Central" sentence above the code block.

"every connection downloads" is also too strong: Ivy caches, so it's the first connection per user. On EMR that first connect took 26.7 s with the default 60 s timeout (both jars, about 109 MB), the second 24.3 s. The timeout warning is still fair for slow egress, just say "first connection".

docs/setup/emr.md
```diff
@@ -78,3 +78,50 @@ spark.sql("SELECT ST_Point(0, 0)").show()
...
+
+```bash
+# Install R and the Sedona R interface
+sudo yum install -y R
+sudo R -e 'install.packages(c("sparklyr", "apache.sedona"), repos = "https://cloud.r-project.org")'
+```
+
+### Connect to the cluster from R
+
+EMR installs Spark under `/usr/lib/spark`. Point `SEDONA_JAR_FILES` at the jars the bootstrap script already downloaded into `/jars` so that `sparklyr` uses them instead of resolving the Sedona coordinates from Maven Central every time you connect:
```
jiayuasu: Worth saying here that `SEDONA_JAR_FILES` (or letting sparklyr pull `--packages`) is required, not an optimisation. `spark.yarn.dist.jars` only distributes the jars to executors; a yarn-client driver never gets them on its classpath. With plain sparklyr and nothing adding `--jars`, the driver logged `ClassNotFoundException: org.apache.sedona.viz.sql.SedonaVizExtensions` and `ST_AsText` was unresolved. The Jupyter section works only because EMR runs Livy in cluster deploy mode. Without a sentence on this, readers will drop the env var assuming spark-defaults covers it, which is what your open question 2 asks.

docs/setup/emr.md
```diff
@@ -78,3 +78,50 @@ spark.sql("SELECT ST_Point(0, 0)").show()
...
+## Use Sedona in R
+
+The [`apache.sedona`](https://cran.r-project.org/package=apache.sedona) R package is a [`sparklyr`](https://spark.rstudio.com) extension. Attaching it before `spark_connect()` is enough to register Sedona's serializers, UDTs and UDFs, so there is no R equivalent of `SedonaContext.create()` to call by hand.
+
+!!!note
+	The R interface supports Spark 3.x only. Make sure the EMR release you pick ships a Spark 3 version.
+
+### Extend the initialization script
+
+Add the following to the bootstrap script above, so that R and the two R packages are available on the node you run R from:
```
jiayuasu: A bootstrap action runs on every node, not just the one you run R from. That's fine in practice, and workers only need the R packages for `spark_apply()` (an executor-side aggregation ran without sparklyr on the core node), but the wording suggests otherwise.

docs/setup/emr.md
```diff
@@ -78,3 +78,50 @@ spark.sql("SELECT ST_Point(0, 0)").show()
 ```
 
 Note that: you don't need to run the `SedonaRegistrator.registerAll(spark)` or `SedonaContext.create(spark)` because `org.apache.sedona.sql.SedonaSqlExtensions` in the config will take care of that.
+
+## Use Sedona in R
+
+The [`apache.sedona`](https://cran.r-project.org/package=apache.sedona) R package is a [`sparklyr`](https://spark.rstudio.com) extension. Attaching it before `spark_connect()` is enough to register Sedona's serializers, UDTs and UDFs, so there is no R equivalent of `SedonaContext.create()` to call by hand.
+
+!!!note
+	The R interface supports Spark 3.x only. Make sure the EMR release you pick ships a Spark 3 version.
```
jiayuasu: Maybe "apache.sedona 1.9.1 supports Spark 3.x only". Spark 4 support went in with #3340 for the next R release, so this will be stale soon without the version.

docs/setup/emr.md
```diff
@@ -78,3 +78,50 @@ spark.sql("SELECT ST_Point(0, 0)").show()
 ```
 
 Note that: you don't need to run the `SedonaRegistrator.registerAll(spark)` or `SedonaContext.create(spark)` because `org.apache.sedona.sql.SedonaSqlExtensions` in the config will take care of that.
+
+## Use Sedona in R
+
+The [`apache.sedona`](https://cran.r-project.org/package=apache.sedona) R package is a [`sparklyr`](https://spark.rstudio.com) extension. Attaching it before `spark_connect()` is enough to register Sedona's serializers, UDTs and UDFs, so there is no R equivalent of `SedonaContext.create()` to call by hand.
```
jiayuasu, line 84: Confirmed, no change needed. One side effect worth knowing: attaching the package sets `spark.kryo.registrator` to `SedonaVizKryoRegistrator` through an R option, which sparklyr passes as `--conf`, overriding the `SedonaKryoRegistrator` from spark-defaults. Harmless, it's a superset.

docs/setup/emr.zh.md
```diff
@@ -78,3 +78,50 @@ spark.sql("SELECT ST_Point(0, 0)").show()
 ```
 
 注意：您不需要再调用 `SedonaRegistrator.registerAll(spark)` 或 `SedonaContext.create(spark)`，因为配置中的 `org.apache.sedona.sql.SedonaSqlExtensions` 已经为您完成了这些工作。
+
+## 在 R 中使用 Sedona
```
jiayuasu, line 82: Same edits needed here: libcurl-devel, `3.5_2.12`, and the two reworded notes. The translation itself reads fine.

jiayuasu (changes requested):
I ran this on EMR 7.9.0 (Spark 3.5.5) before reviewing: the page's bootstrap from master plus the R lines from this PR, verbatim, on a 1 primary + 1 core m5.xlarge cluster, then a second cluster with the libcurl fix. Two things need to change before this can go in, details in the inline comments:

1. `yum install -y R` alone doesn't get you sparklyr on EMR 7. The R `curl` package fails to build because the libcurl headers are missing, and `R -e` still exits 0 so the bootstrap looks fine. `sudo yum install -y R libcurl-devel` fixes it.
2. The `SEDONA_JAR_FILES` snippet still points at the `3.3_2.12` jar. Master moved this page to `3.5_2.12` in #3300, so this needs a rebase.

The note about `SEDONA_JAR_FILES` replacing both Maven coordinates is also wrong: it only replaces the Sedona one, sparklyr still passes `--packages org.datasyslab:geotools-wrapper:...` and pulls it from Maven Central.

With those fixed everything in the section works. `spark_connect(master = "yarn", spark_home = "/usr/lib/spark")` comes up in about 26 s, the verify one-liner prints `<POINT (0 0)>`, `ST_Transform` works, and no extra `library()` is needed because sparklyr exports `%>%` and `collect`. The Chinese mirror is a faithful translation and just needs the same edits.

On your open questions:

- EMR release: the page is already on 7.9.0 / Spark 3.5 after #3300, so rebasing is all that's needed. Any 7.x works.
- `SEDONA_JAR_FILES` is not redundant. `spark.yarn.dist.jars` only ships the jars to executors; in yarn-client mode the driver never sees them. I checked with plain sparklyr and no apache.sedona loaded: `ClassNotFoundException: org.apache.sedona.viz.sql.SedonaVizExtensions` on the driver and `ST_AsText` unresolved. The notebook path only works because EMR runs Livy in cluster deploy mode. Worth a sentence in the doc.
- The AMI doesn't ship R. `yum install R` gives R 4.5.3 from the AL2023 repos in about 85 s.
- Core/task nodes don't need the R packages for `sdf_sql()`. An executor-side aggregation ran fine with sparklyr missing on the core node. Only `spark_apply()` needs R on workers.
- I didn't try RStudio, only Rscript in an EMR step.

Unrelated to this PR but noticed on the way: the existing pip lines break `/usr/bin/aws` on EMR 7.9 (`No module named 'dateutil'`) and the last pip line aborts on `Cannot uninstall packaging 21.3`, leaving matplotlib without its deps. That should be a separate issue.

jiayuasu (approved):
Re-ran the updated section on EMR 7.9.0 (Spark 3.5.5): with libcurl-devel and the 3.5 jar path the bootstrap installs sparklyr and apache.sedona on both nodes, spark_connect() comes up, and the verify query returns POINT (0 0). The reworded notes match what sparklyr actually does.

PR #3379, opened 2026-09-16
[GH-1951] Support zstd-compressed NetCDF4 in RS_FromNetCDF

### LINKED ISSUE #1951: RS_FromNetCDF fails on NetCDF4/HDF5 with filter type=32015 (e.g. ZSTD)
I encountered a runtime failure when using RS_FromNetCDF(content, 'v') on a valid NetCDF4 file stored in S3.

The error is:

```Exception occurred while evaluating expression RS_FromNetCDF - inputs: [[B@xxxx, v], cause: Unknown filter type=32015```

This happens because the file uses an HDF5 compression filter (likely BLOSC, filter ID 32015), which is not supported by the version of netcdf-java (4.6.11) used in netcdfAll.

Steps to reproduce:

    Load .nc file containing BLOSC/ZSTD compression into Spark via binaryFile.

    Try to evaluate RS_FromNetCDF(content, 'variable').

A workaround is to re-encode the file using nccopy or xarray to remove unsupported filters, but this limitation is undocumented and leads to runtime crashes.

Expected:

    An informative exception or warning.

    Ideally: support for filter 32015 

Environment:

    Sedona: 1.7.1

    Spark: 3.5.0

    File: NetCDF4/HDF5 with ZSTD filter

minhpham1810:

- Yes, and the PR name follows the format `[GH-XXX] my subject`. Closes #1951

`RS_FromNetCDF` failed on NetCDF4/HDF5 variables compressed with Zstandard (HDF5 filter 32015) with `Unknown filter type=32015`.

- Upgrade netCDF-Java (`cdm-core`) from 5.4.2 to 5.10.0. In 5.4.2 the HDF5 filters are hardcoded. Since 5.5.0, netCDF-Java loads them through the `ucar.nc2.filter.FilterProvider` service. 5.10.0 is still Java 8 bytecode. On the transitive side, the only changes are version bumps of jcommander, joda-time, jdom2 and httpmime.
- netCDF-Java has no Zstandard filter, so this PR adds `ZstdFilter` and registers it in `META-INF/services/ucar.nc2.filter.FilterProvider`. It decodes with `zstd-jni`, added as a `provided` dependency because Spark 3.5, 4.0 and 4.1 already ship it.
- Instead of a custom error message, this relies on netCDF-Java 5.10.0, which already reports other unsupported filters with their id (for example `Unknown filter: no filter found with id 307`). The docs now list the supported filters and the `nccopy -F none` workaround.

## How was this patch tested?

- New `RasterConstructorsTest#testNetCdfZstdCompressed`, using a small Zstandard-compressed fixture (`test_zstd.nc4`) generated with netCDF4-python. It checks the raster metadata and every pixel value.
  - It fails on 5.4.2 with the exact error from the issue.
  - It also fails on 5.10.0 when the service file is removed, with `no filter found with id 32015`.
- Existing NetCDF tests pass on 5.10.0: `RasterConstructorsTest`, plus `netcdfMetadataTest` and `rasteralgebraTest` (Spark 3.5, Scala 2.13).


jiayuasu (approved):
Thank you!

PR #3384, opened 2026-09-21
[GH-3372] Allow NaN nodata in RS_AsRaster

### LINKED ISSUE #3372: RS_AsRaster rejects NaN as a noDataValue on floating point pixel types
## Expected behavior

Since #3366 a raster band may declare `NaN` as its nodata value, which is what GDAL writes for floating point rasters. `RS_AsRaster` should accept `NaN` as the `noDataValue` argument for a floating point pixel type, and should be able to inherit a `NaN` nodata value from the reference raster when the argument is omitted.

## Actual behavior

Both forms fail with an `IllegalArgumentException`. `RasterUtils.assertNoDataValueRepresentable` rejects every non-finite value regardless of pixel type, so:

- `RS_AsRaster(geom, ref, 'F', 1, double('NaN'))` is refused even though a float band stores NaN exactly;
- omitting `noDataValue` when the reference raster's band 1 has a `NaN` nodata value is refused too, because the inherited value goes through the same check.

The second form is the one likely to be hit in practice: read a GDAL float raster whose `GDAL_NODATA` is `nan`, then rasterize a geometry against it as a reference. That worked before #3366 only because the NaN nodata value was silently dropped on read.

The check is deliberate for integral pixel types, where NaN cannot be stored and would collapse to 0. The fix is to allow non-finite values when the target pixel type is `F` or `D`, and keep rejecting them otherwise.

```sql
SELECT RS_AsRaster(ST_GeomFromWKT('POLYGON ((1 1, 3 1, 3 3, 1 3, 1 1))'), 'F', 1, double('NaN'))
-- expected a float raster whose nodata value is NaN
-- got: IllegalArgumentException, noDataValue ... cannot be represented
```

## Settings

Sedona version = master (2.0.0-SNAPSHOT)
API type = Scala, Java, Python, SQL
Spark version = 3.5
Scala version = 2.12
JRE version = 11

mtcd001-tech:

Yes, I have read the [Contributor Rules](https://sedona.apache.org/latest/community/rule/) and [Contributor Development Guide](https://sedona.apache.org/latest/community/develop/).

Closes #3372.

Allow NaN nodata values for float and double output in `RS_AsRaster`, both when supplied explicitly and when inherited from the reference raster. Preserve rejection for integer pixel types, finite-value checks for burn values, and exact storage checks for finite nodata values.

Infinity remains rejected: GeoTools cannot construct an infinite nodata category bound. This matches the existing infinite-nodata handling in `RS_FromGeoTiff`. The API documentation now makes this distinction explicit.

## How was this patch tested?

On Windows with Java 17:

- Both new NaN regression tests fail against unchanged upstream production code and pass with the fix. They verify background pixels, nodata metadata, and the count excluding nodata for F/D output.
- Tests also cover inherited NaN rejection for integer bands, nonfinite integer nodata, infinite floating nodata, and nonfinite burn values.
- All 128 tests in `RasterConstructorsTest`, `RasterizationTest`, `RasterBandAccessorsTest`, `RasterUtilsTest`, and `SerdeTest` pass:

  ```sh
  mvn -pl common -am -Dtest=RasterConstructorsTest,RasterizationTest,RasterBandAccessorsTest,RasterUtilsTest,SerdeTest -Dsurefire.failIfNoSpecifiedTests=false test
  ```

- Including `RasterBandEditorsTest` gives 146 passing tests and one existing failure in `testClip`: a Windows line-ending difference in GeoTools text output. The same failure was reproduced against unchanged upstream production code.
- Spotless and `git diff --check` pass.

Spark and Python suites were not run locally.

Yes, the `RS_AsRaster` documentation describes support for NaN nodata on F/D and the infinity restriction.


jiayuasu (approved):
Thank you!

